PR #43, opened 2018-01-21
Add support in Lorre for forks between runs

vishakh:
Lorre runs periodically to pull in from the Tezos client any new blocks not already in the database. For this, it queries for every block between the maximum level present in the database to the level of the head. Due to a database constraint, any new block record must have an existing block record with a hash corresponding to the new block's predecessor hash. 

In a boundary case, a fork might have occurred between two consecutive Lorre runs, in which case the new block with the least level will not have its predecessor present in the database. This bug fix allows the option to keep querying new blocks until a database record is found for a predecessor hash. 

In order to address all forks and not just this boundary condition, we might in the future have a process running occasionally to match the latest chain returned by the Tezos client to the ones present in the database. Should this check fail, it would trigger a download of the intermediary fork into the database.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
@@ -18,14 +18,14 @@ object ApiOperations {
   /**
     * Fetches the level of the most recent block stored in the database.
     *
-    * @return Max level.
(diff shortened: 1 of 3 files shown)


vishakh:
@grimpil I have created a ticket for invalidating blocks discarded by forks: 

https://app.zenhub.com/workspace/o/cryptonomic/conseil/issues/48

PR #45, opened 2018-01-21
Added ability to run Lorre as a service

vishakh:
Lorre can now run in a staggered infinite loop with a default interval of one minute. In addition, a minor bug relating to the generated Slick schema has been corrected.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -1,5 +1,6 @@
 package tech.cryptonomic.conseil
 
+import com.typesafe.config.ConfigFactory
 import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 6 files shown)


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -14,12 +15,21 @@ import scala.util.{Failure, Success, Try}
   */
 object Lorre extends App with LazyLogging {
 
+  private val conf = ConfigFactory.load
+  val refreshIntervalInSeconds = conf.getInt("lorre.sleepIntervalInSeconds")
```
grimpil: The variable name should ideally match the renamed configuration parameter.

vishakh: Fixed.

PR #73, opened 2018-05-09
Conseil 70 filtering operations by kind

ba205:
Updated batch api operations to filter blocks, operation groups, and operations by kinds.

All (conditional) joins seem to be working as expected.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
@@ -44,24 +44,27 @@ object ApiOperations {
                      operationGroupKinds: Option[Set[String]] = Some(Set[String]())
                    )
 
+  case class isOperationGroupFilter(exists: Boolean)
+  case class isOperationFilter(exists: Boolean)
+
   // Predicates to determine existence of specific type of filter
 
   private def isBlockFilter(filter: Filter): Boolean =
     (filter.blockIDs.isDefined && filter.blockIDs.get.nonEmpty) ||
-    (filter.levels.isDefined && filter.levels.get.nonEmpty) ||
-    (filter.chainIDs.isDefined && filter.chainIDs.get.nonEmpty) ||
-    (filter.protocols.isDefined && filter.protocols.get.nonEmpty)
+      (filter.levels.isDefined && filter.levels.get.nonEmpty) ||
+      (filter.chainIDs.isDefined && filter.chainIDs.get.nonEmpty) ||
+      (filter.protocols.isDefined && filter.protocols.get.nonEmpty)
 
   private def isOperationGroupFilter(filter: Filter): Boolean =
     (filter.operationIDs.isDefined && filter.operationIDs.get.nonEmpty) ||
-    (filter.operationSources.isDefined && filter.operationSources.get.nonEmpty) ||
-    (filter.operationGroupKinds.isDefined && filter.operationGroupKinds.get.nonEmpty)
+      (filter.operationSources.isDefined && filter.operationSources.get.nonEmpty) ||
+      (filter.operationGroupKinds.isDefined && filter.operationGroupKinds.get.nonEmpty)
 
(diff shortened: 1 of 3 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -124,17 +150,48 @@ object ApiOperations {
...
+        filterChainIDs(filter, block) &&
+        filterProtocols(filter, block)})
+    val filteredOpGroups = Tables.OperationGroups.filter({opGroup =>
+        filterOperationIDs(filter, opGroup) &&
+        filterOperationSources(filter, opGroup) &&
+        filterOperationGroupKinds(filter, opGroup)})
+    val filteredOps = Tables.Operations.filter({op => filterOperationKinds(filter, op)})
+
+    val isOGFilter = isOperationGroupFilter(filter)
+    val isOFilter = isOperationFilter(filter)
```
vishakh: 'OGFilter' and 'OFilter' are easy to confuse. Please give them longer, more descriptive names.

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -124,17 +150,48 @@ object ApiOperations {
...
+
+    // Join tables in appropriate ways dependents on filters chosen by user
+    val action = (isOGFilter, isOFilter) match {
+
+      case (true, false)  => {
+        val table = filteredBlocks.join(filteredOpGroups).on(_.hash === _.blockId)
+        for {
+          (b, opGroups) <- table
```
vishakh: How come you are using a for-expression here instead of a simple join?

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -124,17 +150,48 @@ object ApiOperations {
...
+        } yield (b.chainId, b.protocol, b.level, b.proto, b.predecessor, b.validationPass, b.operationsHash, b.protocolData, b.hash, b.timestamp, b.fitness)
+      }
+
+      case (false, false) => {
+        for {
+          b <- filteredBlocks
```
vishakh: Why use a for-expression here?

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
vishakh: Furthermore, since these two variables are being used only once, can we directly use the functions in the tuple on which we pattern match?

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -246,16 +316,47 @@ object ApiOperations {
...
+            filterOperationIDs(filter, opGroup) &&
+            filterOperationSources(filter, opGroup) &&
+            filterOperationGroupKinds(filter, opGroup)})
+        val filteredOps = Tables.Operations.filter({op => filterOperationKinds(filter, op)})
+
+        val isOGFilter = isOperationGroupFilter(filter)
+        val isOFilter = isOperationFilter(filter)
+
+        // Join tables in appropriate ways dependents on filters chosen by user
+        val action = (isOGFilter, isOFilter) match {
```
vishakh: It seems this block of code redundantly appears in three functions. Can we just have a single function which takes in the Filter object and returns a tuple with three flags..and then pattern match on that tuple in each of the three calling functions?

ba205:
I factored out the OpGroup/Op Filter pattern matching, but it's unclear whether that makes the code more readable.

As for the use of for expressions, it is syntactic sugar for flatMapping on the join, but the code would be a little cluttered because you would have all of the logic in the for expression inside standalone parentheses, which might end up obfuscating the code. I could see this change working if there resultant would be a tuple of two or three, but considering how many elements are in the query, I don't think the change would be useful.

Issue #75, opened 2018-05-15
Conseil should accept OPTIONS requests for CORS pre-flight checks


vishakh:
Arronax is not able to run Conseil API queries when Conseil is running remotely.

PR #76, opened 2018-05-15
CONSEIL-75: Add OPTIONS route to Conseil to facilitate CORS preflight checks

vishakh:
As described in #75, Arronax does not work with Conseil when the latter is run remotely. The underlying issue is discussed at https://stackoverflow.com/questions/8153832/xmlhttprequest-changes-post-to-option#8154264. As a fix, a trivial OPTIONS route is being added to ensure OPTIONS requests during CORS preflight checks are being respected. The correct CORS response headers were already being returned.


ba205:
Considering there's one line of change, as long as it compiles and acts as expected, I'm totally fine with this.

PR #79, opened 2018-05-24
Increased default logging verbosity

vishakh:
Lorre and Conseil are regularly failing in production as they are not able to process return values from the Tezos client. This PR increases the Logback verbosity to make it easier to look at production logs and fix issues.

Please note this change will reverted as soon as stability is achieved.


ba205:
Looks good to me!

PR #99, opened 2018-06-21
Conseil 88 conseil should not block on any database operations

ba205:
Tentative Pull Request. Conseil is supposedly non-blocking as of this point, while Lorre maintains a blocking nature. Still need to test on Narwhalnet, and there might be a merge conflict somehow


ba205:
Git issues, leaving pr without merging

PR #104, opened 2018-06-27
Conseil 88 conseil should not block on any database operations

ba205:
Asynchronous Conseil code


vishakh:
Supplanted by newer work.

Issue #69, opened 2018-04-25
Lorre should parameterize the blockchain and network it runs for


vishakh:
At the moment, both Lorre and Conseil are capable of connecting to a single database to write and read data. As a consequence, they effectively only support one network at a time. This should be rectified such that Tezos `zeronet` and `betanet` can both be supported by Lorre and Conseil and multiple instances of the programs can run on the same codebase. 

As part of this ticket, we should also validate the `network` parameter when the client makes an API call.

PR #109, opened 2018-07-03
Parametrizing Network in Lorre, and Network Validation in Conseil

ba205:
This addresses #69

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -21,23 +21,54 @@ object Lorre extends App with LazyLogging {
   lazy val db = DatabaseUtil.db
   val tezosNodeOperator = new TezosNodeOperator(TezosNodeInterface)
 
-  try {
-    while(true) {
-      logger.info("Fetching blocks")
-      processTezosBlocks()
-      logger.info("Fetching accounts")
-      processTezosAccounts()
-      logger.info("Taking a nap")
-      Thread.sleep(sleepIntervalInSeconds * 1000)
-    }
-  } finally db.close()
+  val usage =
+    """
+      Usage: runMain tech.cryptonomic.conseil.Lorre --platform <platform> --network <network>
+    """.stripMargin
+
+  val argsList = args.toList
+  val (platform, network) = argsList match {
+    case "--platform" :: pl :: "--network"  :: nw :: Nil => (pl, nw)
+    case "--network"  :: nw :: "--platform" :: pl :: Nil => (pl, nw)
+    case _ =>
+      println(usage)
+      sys.exit(1)
+  }
+
+  val supportedPlatformsAndNetworks =
+    """
+      | The supported platforms are: tezos.
+      | The supported networks are: zeronet.
+    """
+    .stripMargin
+
+  val networkPlatformValidation = conf.hasPath(s"platforms.$platform.$network")
+
+  networkPlatformValidation match {
+    case false =>
+      println(supportedPlatformsAndNetworks)
+      sys.exit(1)
+    case true =>
+      try {
+        while(true) {
+          logger.info("Fetching blocks")
+          processTezosBlocks(network)
+          logger.info("Fetching accounts")
+          processTezosAccounts(network)
+          logger.info("Taking a nap")
+          Thread.sleep(sleepIntervalInSeconds * 1000)
+        }
+      } finally db.close()
+  }
+
+
 
   /**
     * Fetches all blocks not in the database from the Tezos network and adds them to the database.
     */
-  def processTezosBlocks(): Try[Unit] = {
+  def processTezosBlocks(network: String): Try[Unit] = {
     logger.info("Processing Tezos Blocks..")
-    tezosNodeOperator.getBlocksNotInDatabase("zeronet", followFork = true) match {
+    tezosNodeOperator.getBlocksNotInDatabase(network, followFork = true) match {
       case Success(blocks) =>
         Try {
           val dbFut = TezosDatabaseOperations.writeBlocksToDatabase(blocks, db)
@@ -56,9 +87,9 @@ object Lorre extends App with LazyLogging {
   /**
     * Fetches and stores all accounts from the latest block stored in the database.
     */
-  def processTezosAccounts(): Try[Unit] = {
+  def processTezosAccounts(network: String): Try[Unit] = {
     logger.info("Processing latest Tezos accounts data..")
-    tezosNodeOperator.getLatestAccounts("zeronet") match {
+    tezosNodeOperator.getLatestAccounts(network) match {
       case Success(accountsInfo) =>
         Try {
           val dbFut = TezosDatabaseOperations.writeAccountsToDatabase(accountsInfo, db)
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -9,6 +9,7 @@ import scala.concurrent.Await
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration.{Duration, _}
 import scala.util.{Failure, Success, Try}
+import tech.cryptonomic.conseil.util.ConfigUtil
 
 /**
   * Entry point for synchronizing data between the Tezos blockchain and the Conseil database.
@@ -36,12 +37,14 @@ object Lorre extends App with LazyLogging {
       sys.exit(1)
   }
 
+  val platformsAndNetworks = ConfigUtil.parsePlatformsAndNetworks()
+
   val supportedPlatformsAndNetworks =
     """
-      | The supported platforms are: tezos.
-      | The supported networks are: zeronet.
+      | The supported platforms and their associated networks are as follows:
+      | (Format: <platform>: <network_one>, <network_two>...)
     """
-    .stripMargin
+    .stripMargin + ConfigUtil.outputPlatformsAndNetworks.stripMargin
 
   val networkPlatformValidation = conf.hasPath(s"platforms.$platform.$network")
 
@@ -61,8 +64,6 @@ object Lorre extends App with LazyLogging {
         }
       } finally db.close()
   }
-
-
 
   /**
     * Fetches all blocks not in the database from the Tezos network and adds them to the database.
--- a/src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
@@ -2,11 +2,12 @@ package tech.cryptonomic.conseil.routes
 
 import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.{Directive, Route}
+import com.typesafe.config.ConfigFactory
 import com.typesafe.scalalogging.LazyLogging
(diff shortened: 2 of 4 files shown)


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -21,23 +21,54 @@ object Lorre extends App with LazyLogging {
...
+    case "--network"  :: nw :: "--platform" :: pl :: Nil => (pl, nw)
+    case _ =>
+      println(usage)
+      sys.exit(1)
+  }
+
+  val supportedPlatformsAndNetworks =
+    """
+      | The supported platforms are: tezos.
+      | The supported networks are: zeronet.
```
vishakh: This should not be hard coded. Please read it from the conf.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -21,23 +21,54 @@ object Lorre extends App with LazyLogging {
...
-      Thread.sleep(sleepIntervalInSeconds * 1000)
-    }
-  } finally db.close()
+  val usage =
+    """
+      Usage: runMain tech.cryptonomic.conseil.Lorre --platform <platform> --network <network>
+    """.stripMargin
+
+  val argsList = args.toList
+  val (platform, network) = argsList match {
```
vishakh: What was the conclusion here? Were we going to use a proper arg parsing library or just assume a pattern for now?

src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
```diff
@@ -68,91 +71,95 @@ object Tezos extends LazyLogging {
...
-          pathEnd {
-            ApiOperations.fetchOperationGroups(filter) match {
-              case Success(operationGroups) => complete(JsonUtil.toJson(operationGroups))
-              case Failure(e) => failWith(e)
-            }
-          } ~ path(Segment) { operationGroupId =>
-            ApiOperations.fetchOperationGroup(operationGroupId) match {
-              case Success(operationGroup) => complete(JsonUtil.toJson(operationGroup))
-              case Failure(e) => failWith(e)
+    validate(conf.hasPath(s"platform.tezos.$network"), s"Network not supported, current options include: 'zeronet'") {
```
vishakh: Let's not hard code the available networks. Let's get them from the conf.

src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
```diff
@@ -68,91 +71,95 @@ object Tezos extends LazyLogging {
...
+      get {
+        validate(conf.hasPath(s"platforms.tezos.$network"), s"$network not supported, current options include: 'zeronet'") {
```
vishakh: Is this redundant?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ba205: Okay, will do.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ba205: I thought we were just doing this for now, since we're the ones calling it from the command line mostly.

src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
ba205: Got it. Although, network isn't available in this file, it's abstracted out in Conseil.scala

ba205: *tezos

src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
ba205: Wow, I have no idea how that happened. Definitely redundant

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
vishakh: Can you open a ticket then for using an arg parsing library?

vishakh: I opened https://github.com/Cryptonomic/Conseil/issues/121.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -17,43 +18,67 @@ object Lorre extends App with LazyLogging {
...
+  val sleepIntervalInSeconds = conf.getInt("lorre.sleepIntervalInSeconds")
 
   lazy val db = DatabaseUtil.db
   val tezosNodeOperator = new TezosNodeOperator(TezosNodeInterface)
 
-  var iterationsOfLorre = 0
-  var shouldFeeBeCalculatedThisIteration = true
+  val usage =
+    """
+      Usage: runMain tech.cryptonomic.conseil.Lorre --platform <platform> --network <network>
```
ivanopagano: why not a simple string, without all the formatting fuss?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -17,43 +18,67 @@ object Lorre extends App with LazyLogging {
...
+  val supportedPlatformsAndNetworks =
+    """
+      | The supported platforms and their associated networks are as follows:
+      | (Format: <platform>: <network_one>, <network_two>...)
+    """
+    .stripMargin + ConfigUtil.outputPlatformsAndNetworks.stripMargin
+
+  val networkPlatformValidation = conf.hasPath(s"platforms.$platform.$network")
+
+  networkPlatformValidation match {
```
ivanopagano: it's not that relevant to me, but pattern matching on bools can be frowned upon and it's not optimized as well as a simple`if-else`

Usually you should prefer

```
if (networkPlatformValidation) // or better networkPlatformValid (!)
  try {
    while (....
  }
else {
  println(...)
  system.exit(1)
}
```

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -64,17 +89,17 @@ object Lorre extends App with LazyLogging {
...
+    tezosNodeOperator.getLatestAccounts(network) match {
       case Success(accountsInfo) =>
         Try {
           val dbFut = TezosDatabaseOperations.writeAccountsToDatabase(accountsInfo, db)
           dbFut onComplete {
             case Success(_) => logger.info(s"Wrote ${accountsInfo.accounts.size} accounts to the database.")
             case Failure(e) => logger.error(s"Could not write accounts to the database because $e")
           }
-          Await.result(dbFut, Duration.Inf)
+          Await.result(dbFut, Duration.apply(awaitTimeInSeconds, SECONDS))
```
ivanopagano: this can be simplified everywhere you happen to notice using

```scala
import scala.concurrent.duration._ //import
....
//profit
Await.result(dbFut, awaitTimeInSeconds.seconds)
```

src/main/scala/tech/cryptonomic/conseil/util/ConfigUtil.scala
```diff
@@ -0,0 +1,61 @@
...
+  private val conf = ConfigFactory.load
+
+  /**
+    * Output Map, keys being blockchain platforms, values being comma and space
+    * separated string of their networks
+    * @return platform, network Map
+    */
+  def parsePlatformsAndNetworks(): Map[String, String] = {
+    val obj = conf.getObject("platforms").asScala.toMap
+    val platformsAndNetworks = obj.map { pair =>
```
ivanopagano: You can pattern match anytime you use a lambda in the `{ case .... }` form

I'll spend a few words here to make it clear:
``` { case x => doSomething(x) }``` is syntax shortcut to define a `PartialFunction[A, B]` which is a function object that is only defined on those input values where the pattern matches (it's a partial function as opposed to a total function, meaning that it's not defined on the whole domain declared, i.e. the `A` type)

You can usually pass such PF literals when a function is expected, with the caveat that you must consider what happens when the input value is not handled.

The advantage you get is that you can directly "decompose" the input values, especially tuples or case classes, to make the code less cluttered and more to the point.

In this case it would look like

```scala
  val platformsAndNetworks = obj.map {
    case (platforms, networkMap) =>
      val networks = networkMap.unwrapped().asInstanceOf...
  }
```

src/main/scala/tech/cryptonomic/conseil/util/ConfigUtil.scala
```diff
@@ -0,0 +1,61 @@
...
+object ConfigUtil {
+
+  private val conf = ConfigFactory.load
+
+  /**
+    * Output Map, keys being blockchain platforms, values being comma and space
+    * separated string of their networks
+    * @return platform, network Map
+    */
+  def parsePlatformsAndNetworks(): Map[String, String] = {
```
ivanopagano: Will the result change over different calls?

Otherwise it could be possible to make it a [lazy] `val` and avoid recomputing it everytime.

That will also simplify the following method on like `39`
```scala
  def parseNetworksForAGivenPlatform(platform: String): String =
    parsePlatformsAndNetworks(platform)
```

src/main/scala/tech/cryptonomic/conseil/util/ConfigUtil.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+    * Given a specific platform, output comma and space separated string of its
+    * networks
+    * @param platform Blockchain platform
+    * @return
+    */
+  def parseNetworksForAGivenPlatform(platform: String): String = {
+    val platformsAndNetworks = parsePlatformsAndNetworks()
+    platformsAndNetworks(platform)
```
ivanopagano: Consider this: 
using this form instead of `platformAndNetworks.get(platform)` will throw an `NoSuchElementException` if the input is not in the keyset.

the `get(...)` form returns an `Option` instead

src/main/scala/tech/cryptonomic/conseil/util/ConfigUtil.scala
```diff
@@ -0,0 +1,61 @@
...
+    * Return newline and tab separated string represented in the following
+    * format:
+    * <platform>: <network_one>, <network_two>...\n
+    * @return
+    */
+  def outputPlatformsAndNetworks(): String = {
+    val platformsAndNetworks = parsePlatformsAndNetworks()
+    //collect platform and it's list of networks into output string, separated
+    //by newlines and tabs
+    platformsAndNetworks.foldLeft(""){ (acc, platformsAndNetworksPair) =>
```
ivanopagano: As shown before, using a pattern match here can simplify the code a lot

```scala
  platformAndNetworks.foldLeft("") {
    case (acc, (platform, networks)) => s"$acc$platform: $networks\n\t"
  }
```

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ba205: That's so counterintuitive! I thought pattern matching was to way to go always. Interesting

ivanopagano: In fact this is a scala thing... we were even thinking of asking for a compiler rewrite rule to convert it under the hood...

I've seen the pattern used in F# just the other day... I wonder what haskell people think

vishakh:
Approach under consideration.

PR #110, opened 2018-07-05
Added avgFee route to Conseil

ba205:
operations/avgFee acts as expected, returns mean - stddev, mean, and mean + stddev as json

also, user can't ask for more than 10000 entries anymore

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
@@ -117,6 +117,11 @@ object Tezos extends LazyLogging {
               case Success(operations) => complete(JsonUtil.toJson(operations))
               case Failure(e) => failWith(e)
             }
+          } ~ path("avgFees") {
+            ApiOperations.averageFee(filter) match {
+              case Success(fees) => complete(JsonUtil.toJson(fees))
+              case Failure(e) => failWith(e)
+            }
           }
         }
       }
--- a/src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
@@ -70,60 +70,63 @@ object Tezos extends LazyLogging {
   val route: Route = pathPrefix(Segment) { network =>
     get {
       gatherConseilFilter{ filter =>
-        pathPrefix("blocks") {
-          pathEnd {
-            ApiOperations.fetchBlocks(filter) match {
-              case Success(blocks) => complete(JsonUtil.toJson(blocks))
-              case Failure(e) => failWith(e)
+        validate(!filter.limit.isDefined || (filter.limit.isDefined && (filter.limit.get <= 10000)), s"Cannot ask for more than 10000 entries") {
+          pathPrefix("blocks") {
+            pathEnd {
+              ApiOperations.fetchBlocks(filter) match {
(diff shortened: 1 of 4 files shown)


src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
```diff
@@ -70,55 +70,63 @@ object Tezos extends LazyLogging {
...
-          pathEnd {
-            ApiOperations.fetchAccounts(filter) match {
-              case Success(accounts) => complete(JsonUtil.toJson(accounts))
-              case Failure(e) => failWith(e)
-            }
-          } ~ path(Segment) { accountId =>
-            ApiOperations.fetchAccount(accountId) match {
-              case Success(account) => complete(JsonUtil.toJson(account))
-              case Failure(e) => failWith(e)
+        validate(!filter.limit.isDefined || (filter.limit.isDefined && (filter.limit.get <= 10000)), s"Cannot ask for more than 10000 entries") {
```
vishakh: IIRC, we currently allow the user not to provide a `limit` predicate and default to some number when a limit is not provided. Let's not change that behaviour. The point here is to merely prevent the user from asking for too many rows.

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -185,6 +186,51 @@ object ApiOperations {
                   (String, String, String, Boolean, Boolean, Option[String], Int, Option[String], BigDecimal),
                   Seq]) extends Action
 
+  /**
+    * Trait which enumerates the types of operations that can be performed between parties.
+    */
+  sealed trait OperationKind
+
+  case class SeedNonceRevelation() extends OperationKind
```
vishakh: Can we replace multiple case classes with an enum?

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -185,6 +186,51 @@ object ApiOperations {
...
+      case Origination() => "origination"
+      case Reveal() => "reveal"
+      case DoubleEndorsementEvidence() => "double_endorsement_evidence"
+      case Endorsement() => "endorsement"
+    }
+  }
+
+
+  /**
+    * Case class representing a range of values for a possible fee of a transaction.
```
vishakh: "..possible fee of an operation'"

src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
ba205: yes, exactly. !filter.limit.isDefined means that it's okay that the user does not define a limit in their query. if that fails, then it checks if it is defined and it's less than or equal to 10000

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
ba205: Oh, this ended up never being used, because I didn't refactor how we filter at all

src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
vishakh: LOL, I missed that you are testing for a positive, not a negative!

PR #112, opened 2018-07-06
Made Await times configurable

ba205:
Added field 'dbAwaitTimeInSeconds' to application.conf, and used this to said Duration's await times

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -22,7 +22,7 @@ lorre: {
   sleepIntervalInSeconds: 60
 }
 
-#Used for configuring await times
+#Used for configuring await times of blocking operations
 dbAwaitTimeInSeconds: 60
 
 platforms: {
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -22,6 +22,8 @@ lorre: {
   sleepIntervalInSeconds: 60
 }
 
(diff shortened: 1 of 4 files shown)


src/main/resources/application.conf
```diff
@@ -22,6 +22,8 @@ lorre: {
   sleepIntervalInSeconds: 60
 }
 
+dbAwaitTimeInSeconds: 60
```
vishakh: Please add a comment explaining what this does.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -45,7 +45,7 @@ object Lorre extends App with LazyLogging {
             case Success(_) => logger.info(s"Wrote ${blocks.size} blocks to the database.")
             case Failure(e) => logger.error(s"Could not write blocks to the database because $e")
           }
-          Await.result(dbFut, Duration.Inf)
+          Await.result(dbFut, Duration.apply(conf.getInt("dbAwaitTimeInSeconds"), SECONDS))
```
vishakh: Can we evaluate `conf.getInt("dbAwaitTimeInSeconds")` once up top so it fails fast if there us anything incorrect, then use a variable in all these function calls?

Issue #113, opened 2018-07-08
Disable all operations routes


vishakh:
As we have moved all operations logic to the client side through ConseilJS, we want to discourage the use of operations using private keys on the Conseil server. This can later be revived once we start shipping Conseil as a client-side JAR file. For now, let us remove the following routes from `routes/Tezos.scala`:

- [ ] generate_identity
- [ ] originate_account
- [ ] set_delegate
- [ ] send_transaction

PR #117, opened 2018-07-09
Removed write operation routes from Tezos.scala

ba205:
The operations routes have been removed. Also, there was a mistake in the config file reads (master said dbAwaitTimeinSecondsInSeconds instead of dbAwaitTimeInSeconds), so that was fixed too


vishakh:
Addresses #113

Issue #118, opened 2018-07-11
Add 'pkh' to the operations table


vishakh:
- [ ] Add `pkh` to `operations` table
- [ ] Ensure Lorre populates it 
- [ ] Ensure it shows up in Conseil for kind `activate_account`

PR #120, opened 2018-07-13
Conseil 118 add pkh to the operations table

ba205:
Added pkh to operation type and schema, parsed successfully for account activations


vishakh:
Addresses #118

Issue #61, opened 2018-03-03
Lorre should periodically purge old account data


vishakh:
As accounts are block contextual, account data can overwhelm the database. There should be a process to purge old account data from the database in a parametrized manner.

PR #124, opened 2018-07-18
Wrote function to purge accounts which aren't associated with the max…

ba205:
…imum block level, made iterations of purging configurable
#61

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -19,12 +19,14 @@ object Lorre extends App with LazyLogging {
   private val awaitTimeInSeconds = conf.getInt("dbAwaitTimeInSeconds")
   private val sleepIntervalInSeconds = conf.getInt("lorre.sleepIntervalInSeconds")
   private val feeUpdateInterval = conf.getInt("lorre.feeUpdateInterval")
+  private val purgeAccountsInterval = conf.getInt("lorre.purgeAccountsInterval")
 
   lazy val db = DatabaseUtil.db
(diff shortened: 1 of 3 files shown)


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -33,9 +35,13 @@ object Lorre extends App with LazyLogging {
       if (shouldFeeBeCalculatedThisIteration) {
         FeeOperations.processTezosAverageFees()
       }
+      if (shouldAccountsBePurgedThisIteration) {
```
vishakh: Can we not just change the condition to ```iterationsOfLorre % purgeAccountsInterval == 0```?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -199,4 +200,15 @@ object TezosDatabaseOperations {
     }
   }
 
+  def purgeOldAccounts(): Try[Int] = {
```
vishakh: Please document this function.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ba205: Should "shouldFeeBeCalculatedThisIteration" follow the same treatment, for consistency?

vishakh: Yes, please.

ba205: On it.

PR #136, opened 2018-08-03
Conseil-122 return correct mediatypes

### LINKED ISSUE #122: Conseil should return correct media types
At the moment, Conseil returns JSON data in all cases. However, Akka HTTP is returning `Content-Type →text/plain; charset=UTF-8`. We should change this to reflect the correct format.

ivanopagano:
fixes #122 
Handle the server content-type correctly by setting it to `application/json`

 * add some minor corrections and cleanup in the touched code

WIP: would like to see if there's an easy way to automate verification of the route

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -31,6 +31,9 @@ lorre: {
 #Used for configuring await times of blocking operations
 dbAwaitTimeInSeconds: 120
 
+#Used for configuring time-out for reading a web response content
+GET-ResponseEntityTimeoutInSeconds: 90
+POST-ResponseEntityTimeoutInSeconds: 1
 
 platforms: {
   tezos: {
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -26,6 +26,9 @@ lorre: {
   purgeAccountsInterval: 10
   #Used to select how many fees should be averaged together
   numberOfFeesAveraged: 1000
+  #read the tezos net to use from env or default
+  tezos-network: "zeronet"
+  tezos-network: ${?LORRE_TEZOS_NETWORK}
 }
 
 #Used for configuring await times of blocking operations
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -26,9 +26,6 @@ lorre: {
   purgeAccountsInterval: 10
   #Used to select how many fees should be averaged together
   numberOfFeesAveraged: 1000
-  #read the tezos net to use from env or default
-  tezos-network: "zeronet"
-  tezos-network: ${?LORRE_TEZOS_NETWORK}
 }
 
 #Used for configuring await times of blocking operations
--- a/src/main/scala/tech/cryptonomic/conseil/Conseil.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Conseil.scala
@@ -52,10 +52,12 @@ object Conseil extends App with LazyLogging with EnableCORSDirectives {
   }
 
   val bindingFuture = Http().bindAndHandle(route, conseil_hostname, conseil_port)
-  logger.info(s"Bonjour..")
-  while(true){}
-  bindingFuture
-    .flatMap(_.unbind())
-    .onComplete(_ => system.terminate())
+  logger.info("Bonjour...")
 
+  sys.addShutdownHook {
+    bindingFuture
+    .flatMap(_.unbind().andThen{ case _ => logger.info("Server stopped...")} )
+    .flatMap( _ => system.terminate())
+    .onComplete(_ => logger.info("We're done here, nothing else to see"))
+  }
 }
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -16,9 +16,14 @@ import scala.annotation.tailrec
   */
 object Lorre extends App with LazyLogging {
 
-  private final val DefaultNetwork = "zeronet"
-
-  private val network = if (args.length > 0) args(0) else DefaultNetwork
+  private val network =
+    if (args.length > 0) args(0)
+    else {
+      Console.err.println("""
+      | No tezos network was provided to connect to
+      | Please provide a valid network as an argument to the command line""".stripMargin)
+      sys.exit(1)
+    }
   
 
   private val conf = ConfigFactory.load
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -16,35 +16,37 @@ import scala.annotation.tailrec
   */
 object Lorre extends App with LazyLogging {
 
+  private final val DefaultNetwork = "zeronet"
+
+  private val network = if (args.length > 0) args(0) else DefaultNetwork
+  
+
   private val conf = ConfigFactory.load
   private val awaitTimeInSeconds = conf.getInt("dbAwaitTimeInSeconds")
   private val sleepIntervalInSeconds = conf.getInt("lorre.sleepIntervalInSeconds")
   private val feeUpdateInterval = conf.getInt("lorre.feeUpdateInterval")
   private val purgeAccountsInterval = conf.getInt("lorre.purgeAccountsInterval")
-  private val network = conf.getString("lorre.tezos-network")
 
   lazy val db = DatabaseUtil.db
   val tezosNodeOperator = new TezosNodeOperator(TezosNodeInterface)
 
-  var iterationsOfLorre = 0
-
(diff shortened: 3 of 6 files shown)


ivanopagano:
I could prepare the unit-test to verify the correct `content-type` on the route endpoints.

That would imply that the `Tezos.route` should take the `ApiOperation` as a parameter, to make testing possible without a real backend.

@vishakh, @ba205  do you deem it useful and worth the pain? Or should we merge as-is?

src/main/resources/application.conf
```diff
@@ -26,6 +26,9 @@ lorre: {
   purgeAccountsInterval: 10
   #Used to select how many fees should be averaged together
   numberOfFeesAveraged: 1000
+  #read the tezos net to use from env or default
+  tezos-network: "zeronet"
```
ba205: Use platforms.tezos to figure out which network to use in the code. "zeronet" is already there

src/main/resources/application.conf
```diff
@@ -26,6 +26,9 @@ lorre: {
...
+  tezos-network: ${?LORRE_TEZOS_NETWORK}
```
ba205: What is this line supposed to do?

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -52,10 +52,12 @@ object Conseil extends App with LazyLogging with EnableCORSDirectives {
...
 
   val bindingFuture = Http().bindAndHandle(route, conseil_hostname, conseil_port)
-  logger.info(s"Bonjour..")
-  while(true){}
-  bindingFuture
-    .flatMap(_.unbind())
-    .onComplete(_ => system.terminate())
+  logger.info("Bonjour...")
 
+  sys.addShutdownHook {
```
ba205: I'm guessing this is a safer way to run Conseil indefinitely?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -20,14 +21,15 @@ object Lorre extends App with LazyLogging {
...
+  private val network = conf.getString("lorre.tezos-network")
 
   lazy val db = DatabaseUtil.db
   val tezosNodeOperator = new TezosNodeOperator(TezosNodeInterface)
 
   var iterationsOfLorre = 0
 
-  try {
-    while(true) {
+  @tailrec
```
ba205: Wait, you need an annotation to make functions tail recursive? Would this speed up efficiency considerably?

src/main/resources/application.conf
ivanopagano: it first assigns the value `"zeronet"` to `tezos-network`
the it _tries_ to ovverride it by reading from the environment a variable named `LORRE_TEZOS_NETWORK`, and setting that to the config key _only the env-var is set_

see [here](https://github.com/lightbend/config#optional-system-or-env-variable-overrides)

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
ivanopagano: actually it's a safe way to terminate the additional threads on which the Http binding and the actor-system exist

Conseil would continue running even when the main thread exits (with no need to loop indefinitely), because the threads spawned to handle the server is `non-daemon`, which means that it will outlive the thread that spawned it

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: The annotation is only a compile-time check that will fail the compilation if the method is not actually tail-recursive.

Even if you don't add the annotation, `scalac` will rewrite the function optimizing the tail-call when possible. This leaves open the risk that you have incorrectly written the method, assuming it's tail-recursive when it's not, so the annotation is an hint that you expect it to be so, and the compiler confirms it.

In this case the code is pretty obvious, somehow, but it's also critical to check because a loop like this would soon exhaust the stack if not tail-rec

src/main/resources/application.conf
ivanopagano: great, will check tomorrow

src/main/resources/application.conf
vishakh: Could we read this as a command line argument instead of an env var?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeInterface.scala
```diff
@@ -62,7 +62,7 @@ object TezosNodeInterface extends TezosRPCInterface with LazyLogging {
           )
         )
       val response: HttpResponse = Await.result(responseFuture, Duration.apply(awaitTimeInSeconds, SECONDS))
-      val responseBodyFuture = response.entity.toStrict(90.second).map(_.data).map(_.utf8String)
+      val responseBodyFuture = response.entity.toStrict(90.second).map(_.data.utf8String)
```
vishakh: Can we make this configurable?

src/main/resources/application.conf
ivanopagano: sure, will do

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeInterface.scala
ivanopagano: the `90 seconds` timeout?

src/main/resources/application.conf
ivanopagano: you mean that I should look into the existing subtrees in the `platforms.tezos` configuration, and pick the first subtree's root?

what if there is more than one configuration? like 

```
platform.tezos = {
  alphanet = {
  ...
  }
  zeronet = {
  ...
  }
}
```

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
ba205: Awesome, thanks for the clarification!

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeInterface.scala
ba205: That's strange, awaitTimeInSeconds is already there, but that one line has 90 instead of that variable. Replacing 90 with awaitTimeInSeconds should be enough.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -9,12 +9,18 @@ import scala.concurrent.Await
...
 import scala.concurrent.duration.Duration
 import scala.util.{Failure, Success, Try}
+import scala.annotation.tailrec
 
 /**
   * Entry point for synchronizing data between the Tezos blockchain and the Conseil database.
   */
 object Lorre extends App with LazyLogging {
 
+  private final val DefaultNetwork = "zeronet"
```
vishakh: It's dangerous to have a fallback as there might be an existing process handling `zeronet`. It's best to fail here if a valid network is not provided.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeInterface.scala
vishakh: @ivanopagano Yes

@ba205 `awaitTimeInSeconds` is meant to be for DB operations.

ivanopagano: @vishakh I will change it with a property of the `application.conf`

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: fine for me

for the record: I just did refactor the code so that it wouldn't introduce any regression.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -9,12 +9,23 @@ import scala.concurrent.Await
...
 import scala.concurrent.duration.Duration
 import scala.util.{Failure, Success, Try}
+import scala.annotation.tailrec
 
 /**
   * Entry point for synchronizing data between the Tezos blockchain and the Conseil database.
   */
 object Lorre extends App with LazyLogging {
 
+  private val network =
```
ba205: Look up the network validation pull request. There's logic for this here you can reuse/improve if needed. We might scrap that pr and use whatever is useful/best practice here and elsewhere

ivanopagano: I did add my own comments on PR #109 for consideration.

I will not change the code here, and depending on which PR got merged first we will overwrite the code in this branch or the other with the one from #109

Issue #88, opened 2018-06-17
Conseil should not block on any database operations


vishakh:
At the moment, Conseil blocks indefinitely on database operations and this is a potential bottleneck. This should be changed to make all such operations non-blocking.

PR #137, opened 2018-08-07
Conseil-88 async db calls: step 1

ivanopagano:
First step to complete migration of database operations to asynchronous calls.

ref. issue #88 

The task will be completed in separate steps to make review handling simpler by avoiding a huge PR.

Current step includes:
 - average fees computation
 - accounts purging
 - additional cleanup

The approach taken is to
 - compose slick `DBIO` actions as much as possible without running them until the last responsible moment
 - never block to wait for database result but use the `Future` result to define callback operations that will be executed asynchronously from the calling thread.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/FeeOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/FeeOperations.scala
@@ -3,19 +3,27 @@ package tech.cryptonomic.conseil.tezos
 import com.typesafe.config.ConfigFactory
 import com.typesafe.scalalogging.LazyLogging
 import tech.cryptonomic.conseil.Lorre.db
-import scala.concurrent.ExecutionContext.Implicits.global
+import tech.cryptonomic.conseil.tezos.{TezosDatabaseOperations => Tdb}
 
-import scala.concurrent.Await
-import scala.concurrent.duration.{Duration, SECONDS}
+import scala.concurrent.{Future, ExecutionContext}
 import scala.util.{Failure, Success, Try}
+import slick.dbio.DBIOAction
 
 /**
   * Helper classes and functions used for average fee calculations.
   */
 object FeeOperations extends LazyLogging {
 
-  private val conf = ConfigFactory.load
-  private val awaitTimeInSeconds = conf.getInt("dbAwaitTimeInSeconds")
+  private val operationKinds = List(
+    "seed_nonce_revelation",
+    "delegation",
+    "transaction",
+    "activate_account",
+    "origination",
+    "reveal",
+    "double_endorsement_evidence",
+    "endorsement"
+  )
 
   /**
     * Representation of estimation of average fees for a given operation kind.
@@ -37,18 +45,20 @@ object FeeOperations extends LazyLogging {
     * Calculates average fees for each operation kind and stores them into a fees table.
     * @return
     */
-  def processTezosAverageFees(): Try[Unit] = {
+  def processTezosAverageFees()(implicit ex: ExecutionContext): Future[Option[Int]] = {
     logger.info("Processing latest Tezos fee data...")
-    val operationKinds = List("seed_nonce_revelation", "delegation", "transaction", "activate_account", "origination", "reveal", "double_endorsement_evidence", "endorsement")
-    val fees = operationKinds.map(TezosDatabaseOperations.calculateAverageFees)
-    Try {
-      val dbFut = TezosDatabaseOperations.writeFeesToDatabase(fees, db)
-      dbFut onComplete {
-        case Success(_) => logger.info(s"Wrote average fees to the database.")
-        case Failure(e) => logger.error(s"Could not write average fees to the database because $e")
+    val computeAndStore = for {
+      fees <- DBIOAction.sequence(operationKinds.map(Tdb.calculateAverageFeesIO))
+      dbWrites <- Tdb.writeFeesIO(fees.collect{ case Some(fee) => fee })
+    } yield dbWrites
+
+    db.run(computeAndStore)
+      .andThen{
+        case Success(Some(written)) => logger.info("Wrote {} average fees to the database.", written)
+        case Success(None) => logger.info("Wrote average fees to the database.")
+        case Failure(e) => logger.error("Could not write average fees to the database because", e)
       }
-      Await.result(dbFut, Duration.Inf)
-    }
+
   }
 
 }
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
@@ -198,7 +198,7 @@ object TezosDatabaseOperations extends LazyLogging {
     }
   }
 
-  /** convertions from domain objects to database row format */
+  /** conversions from domain objects to database row format */
   object RowConversion {
 
     private[TezosDatabaseOperations] def ofAverageFees(in: AverageFees) =
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
@@ -149,7 +149,7 @@ object TezosDatabaseOperations extends LazyLogging {
     */
   def writeFeesIO(fees: List[AverageFees]): DBIO[Option[Int]] =
     Tables.Fees ++= fees.map(RowConversion.ofAverageFees)
-  
+
   /**
     * Given the operation kind, return range of fees and timestamp for that operation.
     * @param kind  Operation kind
@@ -174,7 +174,7 @@ object TezosDatabaseOperations extends LazyLogging {
         .take(numberOfFeesAveraged)
         .result
 
-    opQuery.map { 
+    opQuery.map {
       timestampedFees =>
         timestampedFees.headOption.map {
           case (_, latest) =>
@@ -189,8 +189,9 @@ object TezosDatabaseOperations extends LazyLogging {
     */
   def purgeOldAccounts()(implicit ex: ExecutionContext): Future[Int] = {
     val purged = dbHandle.run {
-      val maxLevel = SupportQueries.accountsMaxBlockLevel
-      Tables.Accounts.filter(_.blockLevel =!= maxLevel).delete
+      accountsMaxBlockLevel.flatMap( maxLevel =>
+        Tables.Accounts.filter(_.blockLevel =!= maxLevel).delete
+      ).transactionally
     }
     purged.andThen {
       case Success(howMany) => logger.info("{} accounts where purged from old block levels.", howMany)
@@ -212,25 +213,17 @@ object TezosDatabaseOperations extends LazyLogging {
 
   }
 
-  object SupportQueries {
+  /* use as max block level when none exists */
+  private[tezos] val defaultBlockLevel: BigDecimal = -1
 
-    /* use as max block level when none exists */
-    private[TezosDatabaseOperations] val defaultBlockLevel: BigDecimal = -1
-
-     /*
-      * Computes the level of the most recent block in the accounts table
-      * or [[defaultBlockLevel]] if no block exists.
-      * The result is meant to be chained into other Query-like operations and not evaluated immediately.
-      *
-      * @return a [[Rep]] of the level
-      */
-    private[TezosDatabaseOperations] def accountsMaxBlockLevel: Rep[BigDecimal] = 
-      Tables.Accounts
-        .map(_.blockLevel)
-        .max
-        .ifNull(defaultBlockLevel)
-
-  }
-
+  /**
+    * Computes the level of the most recent block in the accounts table or [[defaultBlockLevel]] if none is found.
+    */
+  private[tezos] def accountsMaxBlockLevel: DBIO[BigDecimal] =
+    Tables.Accounts
+      .map(_.blockLevel)
+      .max
+      .getOrElse(defaultBlockLevel)
+      .result
 
 }
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
@@ -177,8 +177,8 @@ object TezosDatabaseOperations extends LazyLogging {
     opQuery.map { 
       timestampedFees =>
         timestampedFees.headOption.map {
-          case (_, earliest) =>
-            computeAverage(earliest, timestampedFees)
+          case (_, latest) =>
+            computeAverage(latest, timestampedFees)
         }
     }
   }
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
@@ -1,21 +1,21 @@
 package tech.cryptonomic.conseil.tezos
 
 import com.typesafe.config.ConfigFactory
+import com.typesafe.scalalogging.LazyLogging
 import slick.jdbc.PostgresProfile.api._
-import tech.cryptonomic.conseil.tezos.ApiOperations.{awaitTimeInSeconds, dbHandle}
+import tech.cryptonomic.conseil.tezos.ApiOperations.dbHandle
 import tech.cryptonomic.conseil.tezos.FeeOperations._
 import tech.cryptonomic.conseil.tezos.TezosTypes.{AccountsWithBlockHashAndLevel, Block}
 import tech.cryptonomic.conseil.util.MathUtil.{mean, stdev}
 
-import scala.concurrent.duration.{Duration, SECONDS}
-import scala.concurrent.{Await, Future}
+import scala.concurrent.{Future, ExecutionContext}
 import scala.math.{ceil, max}
-import scala.util.Try
+import scala.util.{Try, Success, Failure}
 
 /**
   * Functions for writing Tezos data to a database.
   */
-object TezosDatabaseOperations {
+object TezosDatabaseOperations extends LazyLogging {
 
   private val conf = ConfigFactory.load
   private val awaitTimeInSeconds = conf.getInt("dbAwaitTimeInSeconds")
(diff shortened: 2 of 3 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/FeeOperations.scala
```diff
@@ -37,18 +45,20 @@ object FeeOperations extends LazyLogging {
     * Calculates average fees for each operation kind and stores them into a fees table.
     * @return
     */
-  def processTezosAverageFees(): Try[Unit] = {
+  def processTezosAverageFees()(implicit ex: ExecutionContext): Future[Option[Int]] = {
```
ba205: Is explicitly writing that this function uses an ExecutionContext just good practice for documentation, or is there another benefit? (Totally fine either way)

src/main/scala/tech/cryptonomic/conseil/tezos/FeeOperations.scala
```diff
@@ -37,18 +45,20 @@ object FeeOperations extends LazyLogging {
...
-    val fees = operationKinds.map(TezosDatabaseOperations.calculateAverageFees)
-    Try {
-      val dbFut = TezosDatabaseOperations.writeFeesToDatabase(fees, db)
-      dbFut onComplete {
-        case Success(_) => logger.info(s"Wrote average fees to the database.")
-        case Failure(e) => logger.error(s"Could not write average fees to the database because $e")
+    val computeAndStore = for {
+      fees <- DBIOAction.sequence(operationKinds.map(Tdb.calculateAverageFeesIO))
+      dbWrites <- Tdb.writeFeesIO(fees.collect{ case Some(fee) => fee })
+    } yield dbWrites
```
ba205: I'm guessing the idea here is to leave as many computations in our asynchronous computation in the for expression as possible? Besides good practice and readability, does this have performance benefits? (It definitely looks a lot nicer! :) )

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -156,64 +143,94 @@ object TezosDatabaseOperations {
...
-          low = fee.low,
-          medium = fee.medium,
-          high = fee.high,
-          timestamp = fee.timestamp,
-          kind = fee.kind
-        )
-      }
-  }
-
+  def writeFeesIO(fees: List[AverageFees]): DBIO[Option[Int]] =
```
ba205: Why convert feesToDatabaseRows to this RowConversion function you've defined later?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -156,64 +143,94 @@ object TezosDatabaseOperations {
...
+      }
+      val m: Int = ceil(mean(values)).toInt
+      val s: Int = ceil(stdev(values)).toInt
+      AverageFees(max(m - s, 0), m, m + s, ts, kind)
     }
-  }
 
+    val opQuery =
+      Tables.Operations
+        .filter(_.kind === kind)
```
ba205: Oh wow, this might be a way simpler filter than what was originally there. Did you test this with postman to see if both versions gave the same result?

src/main/scala/tech/cryptonomic/conseil/tezos/FeeOperations.scala
```diff
@@ -3,19 +3,27 @@ package tech.cryptonomic.conseil.tezos
...
+import slick.dbio.DBIOAction
 
 /**
   * Helper classes and functions used for average fee calculations.
   */
 object FeeOperations extends LazyLogging {
 
-  private val conf = ConfigFactory.load
-  private val awaitTimeInSeconds = conf.getInt("dbAwaitTimeInSeconds")
+  private val operationKinds = List(
```
ba205: If your filter in Database Operations is semantically correct, this list might be redundant.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -156,64 +143,94 @@ object TezosDatabaseOperations {
...
+        .map(o => (o.fee, o.timestamp))
+        .distinct
+        .sortBy { case (_, ts) => ts.desc }
```
ba205: Really nice to actually give this sorting operation proper labels

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -156,64 +143,94 @@ object TezosDatabaseOperations {
...
+          case (_, latest) =>
+            computeAverage(latest, timestampedFees)
+        }
+    }
+  }
 
   /**
     * Delete all accounts in database not associated with block at maxLevel.
-    * @return No value, database update
+    * @return the number of rows removed
```
ba205: Very nice, I don't know why I wrote otherwise

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -156,64 +143,94 @@ object TezosDatabaseOperations {
...
+      Tables.Accounts.filter(_.blockLevel =!= maxLevel).delete
     }
+    purged.andThen {
+      case Success(howMany) => logger.info("{} accounts where purged from old block levels.", howMany)
+      case Failure(e) => logger.error("Could not purge old block-levels accounts", e)
+    }
+  }
+
+  /** convertions from domain objects to database row format */
+  object RowConversion {
```
ba205: Should we do this sort of thing for all row conversions? What's the rationale here?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -156,64 +143,94 @@ object TezosDatabaseOperations {
...
+        timestamp = in.timestamp,
+        kind = in.kind
+      )
+
   }
 
+  object SupportQueries {
+
+    /* use as max block level when none exists */
+    private[TezosDatabaseOperations] val defaultBlockLevel: BigDecimal = -1
```
ba205: A lot of code we use relies on the original version of this query. Why does putting this in a separate object help? Would rewriting everything that relies on maxBlockLevel to use this SupportQueries object be fruitful?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -156,64 +143,94 @@ object TezosDatabaseOperations {
...
+      val maxLevel = SupportQueries.accountsMaxBlockLevel
+      Tables.Accounts.filter(_.blockLevel =!= maxLevel).delete
     }
+    purged.andThen {
+      case Success(howMany) => logger.info("{} accounts where purged from old block levels.", howMany)
+      case Failure(e) => logger.error("Could not purge old block-levels accounts", e)
+    }
+  }
+
+  /** convertions from domain objects to database row format */
```
vishakh: "Conversions" :grin:

src/main/scala/tech/cryptonomic/conseil/tezos/FeeOperations.scala
ivanopagano: It's necessary, not perfunctory, as it's required to call `andThen` on the `Future` result.

I'm passing this as a parameter because I've removed the

`scala.concurrent.ExecutionContext.Implicits.global`

import from the top.

I consider this best-practice, both to control execution context as a (possibly implicit) parameters and to avoid using the `global` one.

src/main/scala/tech/cryptonomic/conseil/tezos/FeeOperations.scala
ivanopagano: I thought it's needed, it's used on line 51 to sequence the fee calculations for each operation kind.

I just moved the definition here so it stands out better both for reference and for future changes

src/main/scala/tech/cryptonomic/conseil/tezos/FeeOperations.scala
ivanopagano: in this case Slick should optimize the calls, trying to offload the computation to the database when possible or in any case by using callbacks to actually make the second call only after the first result is available.

So yes, you're guessing right, slick works at 3 levels: Queries, DBActions and Future results and all of those can be composed as monads (using `flatMap`, `map`, `for`).

Additionally, composing `Query` objects before calling `.result` will deliver the composed query to the database, that will convert it to sql for "single-shot" execution, and for the database engine to optimize it

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: I'm doing this because `RowConversion` is more simple in its responsibility, and can be reused as a composable piece of code in other computations.

Instead the `feesToDatabaseRows` was a very specialized operation that would have less chance for reuse, and can be expressed as a composition of other more basic operations anyway.

I've put it into a `RowConversion` to make the scope, and thus the responsibility of further similar converters added later on, clearer.

As a last point, you might've noticed that I split the responsibility to filter out possibly missing fees (`maybeFees`) from the actual conversion. That's because the fact that the input fees may not be there is not intrinsically related to writing fees to the database. I opted to shift this responsibility to the caller, which is in charge of putting things together by
 1. retrieving the fees
 2. removing those that are missing
 3. requesting to write the rest to db

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: will do soon

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: yes, I've found that exchanging calls to `_._1`, `_._2` and similar with a pattern matching that names all the pieces, almost always turns out to make the code much easier to follow

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: it happens to me too, all the time

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: oops, done

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: as explained in previous comments, my idea is that we can expect such conversions between "domain types"  and "row types" to increase in time.

Keeping such code in a clear scope and separated from the rest makes everything neater, as I see it.

Please let me know what you think

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: Please ignore this part, I later figured out that this will best fit into another class, and also that I got the semantic wrong, when using the `ifNull`, so I'll need to rewrite this specific piece of code.

Let's discuss  that after the change to see if it works better

src/main/scala/tech/cryptonomic/conseil/tezos/FeeOperations.scala
ba205: Got it, thanks for the clarification! andThen also does seems like a better way to handle things

src/main/scala/tech/cryptonomic/conseil/tezos/FeeOperations.scala
ba205: Awesome! We'll have to talk about this type of refactor for APIOperations.scala. Lots of room for optimization there :)

src/main/scala/tech/cryptonomic/conseil/tezos/FeeOperations.scala
ba205: You're right! Redacted

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ba205: Thanks for the detailed response! And your last paragraph makes sense! I'll keep this design decision in mind

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ba205: Got it.

src/main/scala/tech/cryptonomic/conseil/tezos/FeeOperations.scala
ivanopagano: To be totally honest I rarely see anyone else using it, everyone just sticks to onComplete usually.

Though the idea of `andThen` is to sequence additional side-effects operation after the future, without changing it for real. And there are many cases where this seems to apply fittingly

ivanopagano:
Will need to test for regressions before making this eligible to merge

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: I tried to verify the fees calculation on the staging zeronet environment, but it proved impractical. There are more than 85000 block levels and getting data from a remote connection takes much more than a local node, which would mean that to download and process the Conseil db from scratch seemed to be taking hours.

I did a different test, by using a locally started alphanet node, and running Lorre over a scratch database instance to load all data and make the fee computation, once on this PR's branch and once on the `master`.
This proved more reasonable and the results loaded on postman calling a local Conseil instance are exactly the same.

ivanopagano: These are the results for non-zero operation types, so you can check if they seems to be consistent with real results

```json
{
    "low": 0,
    "medium": 17110,
    "high": 111665,
    "timestamp": 1534163782000,
    "kind": "transaction"
}

{
    "low": 32435,
    "medium": 46000,
    "high": 59565,
    "timestamp": 1534121172000,
    "kind": "origination"
}

{
    "low": 14095,
    "medium": 36364,
    "high": 58633,
    "timestamp": 1534162102000,
    "kind": "delegation"
}
```

ivanopagano:
in the meantime, this is a basic comparison on performances of blocking vs. non-blocking for 12 fees calculation runs

### on branch `benchmarks/blocking-fees-computation`

starting at block count 36928 and going

```
-- Timers ----------------------------------------------------------------------
tech.cryptonomic.conseil.tezos.FeeOperations$.tezos-average-fees.time
             count = 12
         mean rate = 0.00 calls/millisecond
     1-minute rate = 0.00 calls/millisecond
     5-minute rate = 0.00 calls/millisecond
    15-minute rate = 0.00 calls/millisecond
               min = 121.40 milliseconds
               max = 241.63 milliseconds
              mean = 130.84 milliseconds
            stddev = 12.68 milliseconds
            median = 124.64 milliseconds
              75% <= 147.96 milliseconds
              95% <= 147.96 milliseconds
              98% <= 156.10 milliseconds
              99% <= 157.74 milliseconds
            99.9% <= 241.63 milliseconds
```

### on branch `benchmarks/async-fees-computation`

starting at block count 36913 and going

```
-- Timers ----------------------------------------------------------------------
tech.cryptonomic.conseil.tezos.FeeOperations$.tezos-average-fees.time
             count = 12
         mean rate = 0.00 calls/millisecond
     1-minute rate = 0.00 calls/millisecond
     5-minute rate = 0.00 calls/millisecond
    15-minute rate = 0.00 calls/millisecond
               min = 124.40 milliseconds
               max = 297.89 milliseconds
              mean = 134.91 milliseconds
            stddev = 13.87 milliseconds
            median = 132.98 milliseconds
              75% <= 134.65 milliseconds
              95% <= 155.75 milliseconds
              98% <= 166.12 milliseconds
              99% <= 166.12 milliseconds
            99.9% <= 297.89 milliseconds
```

### summary

Essentially no real perceptible performance improvement

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: results using data from `staging/zeronet`, limited to 1000 blocks, comparing once more
branch `CONSEIL-88-async-db-calls` vs `master`

results are identical once more
```
low  |medium |high  |timestamp           |kind                  |
-----|-------|------|--------------------|----------------------|
0    |0      |0     |2018-07-21 10:03:27 |seed_nonce_revelation |
0    |25000  |50000 |2018-07-21 10:02:47 |delegation            |
9763 |33334  |56905 |2018-07-21 10:01:27 |transaction           |
0    |0      |0     |2018-07-21 10:00:07 |activate_account      |
0    |0      |0     |2018-07-21 10:51:47 |origination           |
0    |0      |0     |2018-07-21 10:27:07 |reveal                |
0    |0      |0     |2018-07-21 10:51:47 |endorsement           |
```

ivanopagano:
This is ready-to-merge if everyone approves @vishakh @ba205

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
vishakh: @ivanopagano You can also point your conf file to the staging database if you are simply testing Conseil changes.

ivanopagano: I'd need the staging db conf in that case, but here we're writing fees to it so this case is not safe to test against staging

Issue #131, opened 2018-07-24
Refactor ProcessBlocks


ba205:
Ideally, processBlocks should be refactored into a more maintainable form, both for upcoming invalidated blocks logic, as well as for verifying correctness and testing purposes. This was also mentioned in Scalac's code review.

PR #138, opened 2018-08-07
Conseil 131 refactor processblocks

ba205:
#131 Refactored processBlocks to used batchedGet queries and to separate the fetching of block metadata from it's operations for performance reasons, revised getBlocks function to use new RPC command (head~n) to immediately get block at height n properly


src/main/scala/tech/cryptonomic/conseil/util/FPUtil.scala
```diff
@@ -0,0 +1,11 @@
+package tech.cryptonomic.conseil.util
+
+import scala.util.{Success, Try}
+
+object FPUtil {
+
+  def sequence[A](a: List[Try[A]]): Try[List[A]] = {
```
ivanopagano: implementation is correct

if you want to get a bit further and have both `traverse` and `sequence`, we can partly reuse the implementation from cats

```scala
def map2[A, B, Z](ta: Try[A], tb: Try[B])(f: (A, B) => Z): Try[Z] =
  (ta, tb) match {
    case (Success(a), Success(b)) => Try(f(a,b))
    case (f: Failure[_], _)       => f.asInstanceOf[Try[Z]]
    case (_, f: Failure[_])       => f.asInstanceOf[Try[Z]]
  }

def traverse[A, B](ts: List[A])(f: A => Try[B]): Try[List[B]] =
  ts.foldRight[Try[List[B]]](Success(List.empty)) {
    (a, tbs) => map2(f(a), tbs)(_ :: _)
  }

def sequence[A](ts: List[Try[A]]): Try[List[A]] = traverse(ts)(identity)
```

and be done with it

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -143,90 +141,79 @@ class TezosNodeOperator(node: TezosRPCInterface) extends LazyLogging {
...
+          getBlocks(network, maxLevel+1, headLevel, headHash, followFork)
       }
     }
 
   /**
     * Gets the latest blocks from the database using an offset.
     * @param network        Which Tezos network to go against
     * @param offset         How many previous blocks to get from the start
-    * @param startBlockHash The block from which to offset, using the hash provided or the head if none is provided.
+    * @param hash           The block from which to offset, using the hash provided or the head if none is provided.
```
ivanopagano: would it be better to update the param description with no mention of it being optional?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -143,90 +141,79 @@ class TezosNodeOperator(node: TezosRPCInterface) extends LazyLogging {
...
-        })
+  def getBlocks(network: String, minLevel: Int, maxLevel: Int, hash: String, followFork: Boolean): Try[List[Block]] = {
+    val blocksInRange = getBlock(network, hash, 0)
+      .flatMap(block => processBlocks(network, block.metadata.hash, minLevel, maxLevel))
+    // If followFork is true, create logic for that
+    val blocksFromFork = followFork match {
+      case false => Try{List[Block]()}
+      case true => Try{List[Block]()}
+    }
+    //Need to find good implementation of liftM2
```
ivanopagano: have a look at my comment and see if  `map2` fits

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -143,90 +141,79 @@ class TezosNodeOperator(node: TezosRPCInterface) extends LazyLogging {
...
+    (blocksInRange, blocksFromFork) match {
+      case (Success(rangeBlocks), Success(forkBlocks)) => Success(forkBlocks ++ rangeBlocks)
```
ivanopagano: just double-checkin the logic: the forked blocks comes "before" the requested range block?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -143,90 +141,79 @@ class TezosNodeOperator(node: TezosRPCInterface) extends LazyLogging {
...
-        else if(block.metadata.header.level > minLevel)
-          processBlocks(network, block.metadata.header.predecessor, minLevel, maxLevel, block :: blockSoFar, followFork)
-        else
-          List[Block]()
-      case Failure(e) => throw e
-    }
+                             maxLevel: Int
+                           ): Try[List[Block]] = {
+    val maxOffset: Int = maxLevel - minLevel
+    val offsets = (0 to maxOffset).toList.par
```
ivanopagano: trying to wrap my head around this offset, isn't this supposed to be `(minLevel to maxOffset).toList.par`?

src/main/scala/tech/cryptonomic/conseil/util/FPUtil.scala
ba205: Awesome, thanks!

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ba205: We're going to turn it back to being optional, but good catch!

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ba205: Maybe! I feel like the pattern matching isn't too bad here though

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ba205: Yes, the way it works is we traverse the chain backwards, both in the range and in the potential fork, and the blocks from the fork are always before the blocks in the range

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ba205: Nope! the way it works is getBlock has an offset from the head (latest) block in the tezos node, and maxLevel (currently) is always the latest block when this function is called. So the order of blocks being fetched from the Tezos RPC call is maxLevel, maxLevel - 1, maxLevel -2...maxLevel - (maxLevel - minLevel) = minLevel. 

Essentially, what this function does is it collects blocks from the tezos node using their RPC calls from range minLevel to maxLevel (head), which is a list of Try[Block], hence the need to sequence, in case any block fetching fails

ivanopagano: after additional analysis, we can possibly remove the issue of having blocking operations if we guarantee that the http calls behind the `getBlock` call are handled using asynchronous clients like it's done now with akka-http.
We can take advantage of this simply by converting the `runGetQuery` and `runPostQuery` in the "node" implementation to return an `Future` result, just like we'll do with the database calls

ba205: Okay, so if we change runGetQuery and runPostQuery to Future results (or any sufficient asynchronous type), we'd get over that concern in sequencing the parallel collection? Otherwise, the code would look safe?

vishakh: Don't forget to make the same change for fetching accounts!

ivanopagano: About accounts fetching, the current branch I'm working on [CONSEIL-88-async-db-calls-2](https://github.com/Cryptonomic/Conseil/tree/CONSEIL-88-async-db-calls-2) has a significant impact already on that front

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -108,21 +108,25 @@ class TezosNodeOperator(node: TezosRPCInterface) extends LazyLogging {
     * @param hash     Hash of the given block
     * @return         Block
     */
-  def getBlock(network: String, hash: String): Try[TezosTypes.Block] =
-    node.runGetQuery(network, s"blocks/$hash").flatMap { jsonEncodedBlock =>
+  def getBlock(network: String, hash: String, offset: Option[Int] = None): Try[TezosTypes.Block] = {
+    val offsetString = offset.map(_.toString).getOrElse("")
```
ivanopagano: maybe can be shortened?

I'll try something like this, if it's not too compact to the point of being cryptic

```scala
val offsetString = offset.fold("")(off => s"~$off")
```

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -108,21 +108,25 @@ class TezosNodeOperator(node: TezosRPCInterface) extends LazyLogging {
...
+    val offsetStringWithTilde = offsetString match {
+      case "" => ""
+      case offset => "~" + offset
+    }
+    node.runGetQuery(network, s"blocks/$hash${offsetStringWithTilde}").flatMap { jsonEncodedBlock =>
```
ivanopagano: no need for the curlies if the interpolated value is a simple name

`s"block/$hash$offsetWithTilde"` is ok

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ivanopagano: i.e. `option.fold(a)(f)`is exactly the same as `option.map(f).getOrElse(a)`

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -151,102 +158,68 @@ class TezosNodeOperator(node: TezosRPCInterface)(implicit ec: ExecutionContext)
...
+
     val jsonToOperationGroups: String => List[OperationGroup] =
       json => fromJson[List[List[OperationGroup]]](json).flatten
 
     for {
-      metadataChain <- fetchChainedMetadata(network, startingBlock.metadata, minLevel, maxLevel, followFork)
-      operationsUrls = metadataChain.map(d => s"blocks/${d.hash}/operations")
-      operations <- node.runBatchedGetQuery(network, operationsUrls, blockOperationsFetchConcurrency) map (all => all.map(jsonToOperationGroups))
-    } yield metadataChain.zip(operations).map(Block.tupled)
+      blockMetadata <- node.runBatchedGetQuery(network, blockMetadataUrls, blockOperationsFetchConcurrency) map (all => all.map(jsonToBlockMetadata))
```
vishakh: Rename `all` to `blockMetadata`.

vishakh: Rename so its clear this is not a singleton.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -151,102 +158,68 @@ class TezosNodeOperator(node: TezosRPCInterface)(implicit ec: ExecutionContext)
...
+      blockOperationUrls = blockMetadata.map(metadata => s"blocks/${metadata.hash}/operations")
+      blockOperations <- node.runBatchedGetQuery(network, blockOperationUrls, blockOperationsFetchConcurrency) map (all => all.map(jsonToOperationGroups))
```
vishakh: Rename `all` to something more indicative.

build.sbt
```diff
@@ -25,7 +27,8 @@ libraryDependencies  ++=  Seq(
   "org.scorexfoundation" %% "scrypto" % "2.0.0",
   "com.muquit.libsodiumjna" % "libsodium-jna" % "1.0.4" exclude("org.slf4j", "slf4j-log4j12"),
   "com.github.alanverbner" %% "bip39" % "0.1",
-  "ch.megard" %% "akka-http-cors" % "0.3.0"
+  "ch.megard" %% "akka-http-cors" % "0.3.0",
+  "org.typelevel" %% "cats-core" % "1.4.0"
```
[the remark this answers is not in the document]
ivanopagano: so eager to do it 😄

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -151,102 +158,68 @@ class TezosNodeOperator(node: TezosRPCInterface)(implicit ec: ExecutionContext)
...
-      }
-    else
-      Future.successful(List.empty)
-
+  def getBlocks(network: String, minLevel: Int, maxLevel: Int,
+                startBlockHash: Option[String], followFork: Boolean): Future[List[Block]] = {
+    val hash = startBlockHash.getOrElse("head")
+    val blocksInRange = processBlocks(network, hash, minLevel, maxLevel)
+    val blocksFromFork = followFork match {
+      case false => Future.successful(List.empty : List[Block])
```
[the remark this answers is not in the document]
ba205: Will put List.empty[Block]

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -151,102 +158,68 @@ class TezosNodeOperator(node: TezosRPCInterface)(implicit ec: ExecutionContext)
...
 
     for {
-      metadataChain <- fetchChainedMetadata(network, startingBlock.metadata, minLevel, maxLevel, followFork)
-      operationsUrls = metadataChain.map(d => s"blocks/${d.hash}/operations")
-      operations <- node.runBatchedGetQuery(network, operationsUrls, blockOperationsFetchConcurrency) map (all => all.map(jsonToOperationGroups))
-    } yield metadataChain.zip(operations).map(Block.tupled)
+      blockMetadata <- node.runBatchedGetQuery(network, blockMetadataUrls, blockOperationsFetchConcurrency) map (all => all.map(jsonToBlockMetadata))
+      blockOperationUrls = blockMetadata.map(metadata => s"blocks/${metadata.hash}/operations")
+      blockOperations <- node.runBatchedGetQuery(network, blockOperationUrls, blockOperationsFetchConcurrency) map (all => all.map(jsonToOperationGroups))
+    } yield blockMetadata.zip(blockOperations).map(Block.tupled)
```
ivanopagano: @ba205 since you asked I go for a longer answer here for: "Are we guaranteed that the `zip` operation correctly matches the operations with the corresponding metadata?"

In general we have no guarantees unless the `runBatchedGetQuery` adds those (guarantees) and documents them in the method API.

Right now it works correctly because the implementation of the batched call uses a stream that transforms each input url to the corresponding operation result, and doesn't re-order the stream elements (i.e. we use `mapAsync` instead of `mapAsyncUnordered`).

To be more robust to future changes and refactors we should add a visible marker in the implementation, like comments about the expected invariants of the call...
or returning some identifier with the operation get result that allows to match it with the proper metadata

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -151,102 +158,68 @@ class TezosNodeOperator(node: TezosRPCInterface)(implicit ec: ExecutionContext)
...
-        fetchChainedMetadata(network, pred, minLevel, maxLevel, followFork, current :: collected)
-      }
-    else
-      Future.successful(List.empty)
-
+  def getBlocks(network: String, minLevel: Int, maxLevel: Int,
+                startBlockHash: Option[String], followFork: Boolean): Future[List[Block]] = {
+    val hash = startBlockHash.getOrElse("head")
+    val blocksInRange = processBlocks(network, hash, minLevel, maxLevel)
+    val blocksFromFork = followFork match {
```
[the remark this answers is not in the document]
ba205: Noted. Will change.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ba205: Okay, added documentation to address this concern.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -151,102 +158,68 @@ class TezosNodeOperator(node: TezosRPCInterface)(implicit ec: ExecutionContext)
...
-    else if (aboveMin)
-      getMetadata(network, current.header.predecessor) flatMap { pred =>
-        fetchChainedMetadata(network, pred, minLevel, maxLevel, followFork, current :: collected)
-      }
-    else
-      Future.successful(List.empty)
-
+  def getBlocks(network: String, minLevel: Int, maxLevel: Int,
+                startBlockHash: Option[String], followFork: Boolean): Future[List[Block]] = {
+    val hash = startBlockHash.getOrElse("head")
```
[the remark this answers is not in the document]
ba205: Okay, refactored so instead of startBlockHash being an Option[String], it will be a string that's an optional parameter with "head" as the default value.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -151,102 +158,68 @@ class TezosNodeOperator(node: TezosRPCInterface)(implicit ec: ExecutionContext)
...
-        fetchChainedMetadata(network, pred, minLevel, maxLevel, followFork, collected)
-      }
-    else if (aboveMin)
-      getMetadata(network, current.header.predecessor) flatMap { pred =>
-        fetchChainedMetadata(network, pred, minLevel, maxLevel, followFork, current :: collected)
-      }
-    else
-      Future.successful(List.empty)
-
+  def getBlocks(network: String, minLevel: Int, maxLevel: Int,
```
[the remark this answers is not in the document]
ba205: I wouldn't mind using such a tool!

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -151,102 +158,68 @@ class TezosNodeOperator(node: TezosRPCInterface)(implicit ec: ExecutionContext)
...
-      asyncGetBlock(network, "head").flatMap { blockHead =>
-        val headLevel =  blockHead.metadata.header.level
+      asyncGetBlock(network, "head", None).flatMap { blockHead =>
+        val headLevel = blockHead.metadata.header.level
+        val headHash = blockHead.metadata.hash
         if(headLevel <= maxLevel)
           Future.successful(List.empty)
         else
-          fetchBoundedBlockChain(network, blockHead, maxLevel+1, headLevel, followFork)
+          getBlocks(network, headLevel - 1000, headLevel, Some(headHash), followFork)
```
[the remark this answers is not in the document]
ba205: Right!

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -125,22 +126,28 @@ class TezosNodeOperator(node: TezosRPCInterface)(implicit ec: ExecutionContext)
     * @param hash      Hash of the block
     * @return          the block data wrapped in a [[Future]]
     */
-  def asyncGetBlock(network: String, hash: String): Future[TezosTypes.Block] =
+  def asyncGetBlock(network: String, hash: String, offset: Option[Int]): Future[TezosTypes.Block] = {
+    val offsetString = offset.fold("")(off => s"~$off")
```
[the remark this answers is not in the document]
ba205: Rewrote it, but yeah, they are both the same

build.sbt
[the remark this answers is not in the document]
ba205: Very eager! :)

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -149,96 +150,66 @@ class TezosNodeOperator(val node: TezosRPCInterface)(implicit executionContext:
...
-    */
-  def fetchChainedMetadata(
-    network: String,
-    current: BlockMetadata,
-    minLevel: Int,
-    maxLevel: Int,
-    followFork: Boolean,
-    collected: List[BlockMetadata] = List.empty
-  ) : Future[List[BlockMetadata]] = {
-    val currentLevel = current.header.level
```
ivanopagano: how comes all the following logic disappeared completely?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -9,7 +9,7 @@ import tech.cryptonomic.conseil.util.CryptoUtil.KeyStore
 import tech.cryptonomic.conseil.util.JsonUtil.fromJson
 
 import scala.concurrent.{ExecutionContext, Future}
-import scala.util.{Failure, Try}
+import scala.util.Try
 
 /**
   * Operations run against Tezos nodes, mainly used for collecting chain data for later entry into a database.
```
ivanopagano: I guess `asyncGetAccountForBlock` is now a duplicate, since there's no sync version anymore

ba205: Right, got the yellow bar in IntelliJ

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -113,15 +112,17 @@ class TezosNodeOperator(val node: TezosRPCInterface)(implicit executionContext:
     * @param hash      Hash of the block
     * @return          the block data wrapped in a [[Future]]
     */
-  def getBlock(network: String, hash: String): Future[TezosTypes.Block] =
+  def getBlock(network: String, hash: String, offset: Option[Int] = None): Future[TezosTypes.Block] = {
+    val offsetString = offset.map(_.toString).getOrElse("")
     for {
-      block <- node.runAsyncGetQuery(network, s"blocks/$hash").map(fromJson[TezosTypes.BlockMetadata])
-      ops   <-
+      block <- node.runAsyncGetQuery(network, s"blocks/$hash$offsetString").map(fromJson[TezosTypes.BlockMetadata])
```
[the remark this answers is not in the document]
ba205: yes we did! Nice catch

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ba205: That would require a lot of explanation, but the gist is that I spent a lot of time simplifying this algorithm by separating the part that deals with forks and the part that doesn't deal with forks. Once my forking logic is done, the result will be equivalent, and hopefully a lot more obvious as to why the algorithm is correct.

ba205: Plus, this part is unaware of the recent fact we have learned that we can use offsets from the head using rpc calls.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ivanopagano: did you remove the duplicate?

Issue #143, opened 2018-08-14
Lorre should have retry logic


ba205:
When Lorre processes Blocks and Accounts from the Tezos node, sometimes there's is an error in transmitting the data from the node side, and Lorre just throws an exception. We want Lorre to throw eventually retry the processing at some point, rather than just throw the exception and crash completely.

ivanopagano:
have a look at https://github.com/softwaremill/retry

PR #144, opened 2018-08-16
Conseil 143 lorre should have retry logic

ba205:
#143

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -21,7 +21,7 @@ logger.scala.slick=INFO
 
 lorre: {
   #Used to determine how many times Lorre should retry processing blocks/accounts
-  numberOfRetries: 10
+  numberOfRetries: 5
   #Used to time how long Lorre should sleep for after an iteration
   sleepIntervalInSeconds: 60
   #Used to make sure Lorre records average fees every n iterations
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -19,6 +19,9 @@ conseildb = {
 logger.scala.slick=INFO
 
 lorre: {
+  #Used to determine how many times Lorre should retry processing blocks/accounts
+  numberOfRetries: 10
+  #Used to time how long Lorre should sleep for after an iteration
   sleepIntervalInSeconds: 60
   #Used to make sure Lorre records average fees every n iterations
   feeUpdateInterval: 20
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -3,7 +3,7 @@ package tech.cryptonomic.conseil
 import com.typesafe.config.ConfigFactory
 import com.typesafe.scalalogging.LazyLogging
 import tech.cryptonomic.conseil.tezos.{FeeOperations, TezosDatabaseOperations, TezosNodeInterface, TezosNodeOperator}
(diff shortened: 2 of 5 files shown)


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -35,10 +35,17 @@ object Lorre extends App with LazyLogging {
   lazy val db = DatabaseUtil.db
   val tezosNodeOperator = new TezosNodeOperator(TezosNodeInterface)
 
+  val numberOfRetries = conf.getInt("lorre.numberOfRetries")
+  @tailrec
+  def retry[A](attempts: Int)(fn: => Try[A]): Try[A] = fn match {
```
ivanopagano: This looks like a general purpose function which may be useful somewhere else too.
Would it make sense to put it in a separate file in the `util` package?

src/main/resources/application.conf
```diff
@@ -19,6 +19,9 @@ conseildb = {
 logger.scala.slick=INFO
 
 lorre: {
+  #Used to determine how many times Lorre should retry processing blocks/accounts
+  numberOfRetries: 10
```
ivanopagano: Might be a personal feeling, but this looks like a big number to me for a possibly-heavy operation.
What about 3 or 5?

vishakh: I would agree we should lower it to at least five.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
vishakh: Agreed with @ivanopagano .

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -35,10 +35,17 @@ object Lorre extends App with LazyLogging {
...
+    case Failure(_) if attempts > 1 => retry(attempts-1)(fn)
+    case noMoreAttempts => noMoreAttempts
+  }
+
   @tailrec
   def mainLoop(iteration: Int): Unit = {
-      processTezosBlocks()
-      processTezosAccounts()
+      retry(numberOfRetries)(processTezosBlocks())
```
vishakh: Why do we retry the entire operation of processing all blocks or all accounts? Why not retry at the level of retrying individual blocks or accounts? The latter would be a lot more efficient. You don't want to re-download a million accounts just because one failed.

ivanopagano: @vishakh what would happen if the process can't complete and only part of the blocks were downloaded?
I guess that the next iteration would retry the missing pieces?

What about failing and being unable to complete accounts? That would look even more undesirable, as we'll show a partially inconsistent situation, until the rest of the accounts would be downloaded... and that would not happen if I understand correctly that the next iteration would think that no new block was added and no accounts need downloading?

This leads to considerations about transactionality of the process operations that would need more thought, I think.

ba205: I agree with Ivano on this one, but will be willing to discuss more in person tomorrow.

src/main/resources/application.conf
ba205: Sure thing!

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ba205: Sounds good.

vishakh:
Out of scope.

PR #147, opened 2018-08-20
Conseil-145: streamed fetching

### LINKED ISSUE #145: Lorre http calls to fetch accounts may overflow the client connection pool
For a general overview of the problem refer to this [page in the akka docs](https://doc.akka.io/docs/akka-http/current/client-side/pool-overflow.html)

This behaviour was only experienced locally on my machine, as I tried to run Lorre on a limited number (1000) of blocks from the zeronet staging node.

The accounts to load are around 175 and the http calls to fetch details seems to be too slow to answer wrt the number of queued requests, resulting in the pool `max-open-requests` being reached. The operation is interrupted and is retried on the next loop cycle, with consistently failing results.

I've been able to overcome this only by switching from single requests to a streamed request pipeline for all accounts, which is back-pressured by akka-streams based on the connection pool configuration.

It would be advisable to also try to reuse a similar streaming logic to load the blocks from tezos when the levels missing are significant.

This and other related changes could significantly improve Lorre's first "data bootstrap".

ivanopagano:
Fixes #145 

Some issues arose when loading a lot of block records in Lorre, on a fresh database instance.

The accounts data requests sent with the akka-http client would be too many at the same time, and some form of rate-limiting was needed.

Loading significant numbers of blocks would also take a lot, as the operation was totally sequenced. The current PR take a step toward reducing the operation time, by trying to handle operations concurrently, when possible.

This also led to a lot of simplification on the `TezosNodeOperator` object.

The last part needs careful review to verify that nothing important was actually removed.


ivanopagano:
Just started to download the whole staging zeronet blocks during the night...
meanwhile, on a far away network node zeronet was rebuilt... 

I only found a handful of blocks, yet this is the timing

```
21:45:30.993 [lorre-system-akka.actor.default-dispatcher-14] INFO  tech.cryptonomic.conseil.Lorre$ - Wrote 431 blocks to the database in 311409.077741 seconds.
21:45:30.993 [lorre-system-akka.actor.default-dispatcher-14] INFO  tech.cryptonomic.conseil.Lorre$ - Processing latest Tezos accounts data..
21:45:32.372 [lorre-system-akka.actor.default-dispatcher-4] INFO  tech.cryptonomic.conseil.Lorre$ - Wrote 4 accounts to the database.
21:45:32.373 [run-main-0] INFO  t.c.conseil.tezos.FeeOperations$ - Processing latest Tezos fee data...
21:45:32.509 [scala-execution-context-global-124] INFO  t.c.conseil.tezos.FeeOperations$ - Wrote average fees to the database.
21:45:32.535 [run-main-0] INFO  tech.cryptonomic.conseil.Lorre$ - Taking a nap
```

ivanopagano:
Matching results from local and staging Conseil at block-level 2790 gave 

 - blocks response matches completely
 - operation groups seems to differ in all but the common fields (`protocol`, `chainId`)
 - operations differ widely
 - accounts differ widely
 - fees matches on `transaction`, no other kind is available

src/main/resources/application.conf
```diff
@@ -58,3 +58,33 @@ security: {
...
+# Customization on the underlying actor system
+
+akka {
+
+  # custom host pool for akka-http client connections used for streaming request/responses
+  # tune the configuration based on load-handling capability of tezos nodes
+  # refer to host-connection-pooling section in
+  # https://doc.akka.io/docs/akka-http/current/configuration.html
+  # for available properties and their meaning
+  tezos-streaming-client {
```
ba205: Are these configs defaults? How did you come at this specific configuration as the right way to go?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -24,30 +27,37 @@ object Lorre extends App with LazyLogging {
...
-      }
-      logger.info("Taking a nap")
-      Thread.sleep(sleepIntervalInSeconds * 1000)
-      mainLoop(iteration + 1)
+    Await.ready(
+      processTezosBlocks().flatMap(
+        _ =>
+          processTezosAccounts()
+      )
+    , atMost = Duration.Inf)
```
ba205: what purpose does this serve? Speedup?

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -643,6 +643,12 @@ object ApiOperations {
     Map("block" -> block, "operation_groups" -> operationGroups)
   }
 
+  def blockExists(hash: String)(implicit ec: ExecutionContext): Future[Boolean] =
```
ba205: Is this an APIOperation, so that the user can check if a specific hash exists in the chain, or a utility function to help with testing? Also, this would need logic to account for forks

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeInterface.scala
```diff
@@ -39,22 +60,25 @@ trait TezosRPCInterface {
 object TezosNodeInterface extends TezosRPCInterface with LazyLogging {
 
   private val conf = ConfigFactory.load
-  private val awaitTime = conf.getInt("dbAwaitTimeInSeconds").seconds
-  private val entityGetTimeout = conf.getInt("GET-ResponseEntityTimeoutInSeconds").seconds
-  private val entityPostTimeout = conf.getInt("POST-ResponseEntityTimeoutInSeconds").seconds
+  private[tezos] val awaitTime = conf.getInt("dbAwaitTimeInSeconds").seconds
```
ba205: I don't see a tezos "type" per say. What is this supposed to enforce? That we're in the configuration for tezos specifically?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -103,158 +107,170 @@ class TezosNodeOperator(node: TezosRPCInterface) extends LazyLogging {
...
+    node.runAsyncGetQuery(network, s"blocks/$blockHash/operations")
+      .map(ll => fromJson[List[List[OperationGroup]]](ll).flatten)
+
+  /**
+    * Fetches a single block from the chain, without waiting for the result
+    * @param network   Which Tezos network to go against
+    * @param hash      Hash of the block
+    * @return          the block data wrapped in a [[Future]]
+    */
+  def asyncGetBlock(network: String, hash: String): Future[TezosTypes.Block] =
```
ba205: We definitely should get this PR out first, this will change a lot about how I structure my code for this module (not that that is a bad thing!)

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -103,158 +107,170 @@ class TezosNodeOperator(node: TezosRPCInterface) extends LazyLogging {
...
-      case None =>
-        getBlockHead(network).flatMap(head => Try {
-          processBlocks(network, head.metadata.hash, minLevel, maxLevel, followFork = followFork)
-        })
-      case Some(hash) =>
-        getBlock(network, hash).flatMap(block => Try {
-          processBlocks(network, block.metadata.hash, minLevel, maxLevel, followFork = followFork)
-        })
-    }
+  def recursiveMetadataFetch(
```
ba205: okay, this part should definitely be scrapped. I've been working on a rewrite for this function that's significantly more intuitive (once it's debugged). This requires some domain knowledge, it won't be easily fixed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -103,158 +107,170 @@ class TezosNodeOperator(node: TezosRPCInterface) extends LazyLogging {
...
-          }
-        else if(block.metadata.header.level > maxLevel)
-          processBlocks(network, block.metadata.header.predecessor, minLevel, maxLevel, blockSoFar, followFork)
-        else if(block.metadata.header.level > minLevel)
-          processBlocks(network, block.metadata.header.predecessor, minLevel, maxLevel, block :: blockSoFar, followFork)
-        else
-          List[Block]()
-      case Failure(e) => throw e
-    }
+  private def recursiveBlockFetch(
```
ba205: what gave you the idea to rewrite these? Were you trying to go for speedups? If you want to try that, I would say look at the Refactor ProcessBlocksPr and try and give advice there. It's a separate, tricky issue

ba205: *ProcessBlocks PR

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ba205: actually, everything after this line should either be moved to a separate PR, or scrapped completely.

vishakh: The upshot of doing it this way is we compare strictly sync vs. async logic. Once we convince ourselves the async logic is correct and performant, we can add the improved block fetching functionality to the async code.

src/main/resources/application.conf
ivanopagano: we started with a parallel collection on a forkjoin pool of _parallelism level of 32_ to achieve high throughput, while default `max-connections` for the client is `4`. I tried to keep the number of connections high because the idea would be to reuse the same pool for all streaming http-client requests. I might be overshooting a bit here, possibly any number between 5-10 might work the same for now

The `max-open-requests` is `32` by default and this configuration is a precautionary control to avoid overflowing the server with simultaneous requests (considered bad-practice). On my local trials to download accounts for blocks at level 1000 I immediately hit the limit (something between 40-50 requests were sent together). So I kept it to a decent value above this limit, always considering that this pool would be reused for any similar operation (needing to concurrently send a high number of http requests to tezos, without flooding it) for any endpoint.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: As the operations return `Future` values now, the call to the `processXXX` methods will immediately return without waiting for the answer.

We need to sequence those operations instead (monadically in this case) because we can't start processing accounts before all the blocks are stored.
Additionally the same is true for fees and purge operations, they need to start only after the previous have finished.

In this case I didn't chain them monadically, but blocked on the results, because eventually the code will recursively call itself, and we can't do this within a different thread on a callback if we want it to be tail-recursive (preventing stack-overflows when running for many cycles)

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ivanopagano: @ba205 I perfectly understand your point, this part is the result from my need to load data from zeronet without taking so long, as I needed to compare data for the other work I's on.

As I said elsewhere, I don't expect this code to be merged.

Yet I took the chance to put the code here for review, because:
 1. I wanted to check with you both (which definitely have more domain knowledge!) if my assumptions on simplifying were generally correct (this will be of huge help to make me learn)
 2. I wanted to check with you how an async approach would look like and verify what needs to be fixed to make it correct and be able to later adopt a similar design on the real block-processing rewrite

I beg your patience here if you feel this is wasted time, but I think this work is useful to define the final shape that the code will eventually take

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -59,34 +60,37 @@ class TezosNodeOperator(node: TezosRPCInterface) extends LazyLogging {
...
   /**
     * Fetches all accounts for a given block.
     * @param network    Which Tezos network to go against
     * @param blockHash  Hash of given block.
     * @return           Accounts
     */
-  def getAllAccountsForBlock(network: String, blockHash: String): Try[Map[String, TezosTypes.Account]] = Try {
-    node.runGetQuery(network, s"blocks/$blockHash/context/contracts") match {
-      case Success(jsonEncodedAccounts) =>
+  def getAllAccountsForBlock(network: String, blockHash: String): Future[Map[String, TezosTypes.Account]] =
```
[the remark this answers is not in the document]
ivanopagano: It was actually too much, at least from what I saw when running on my machine (and the reason for this PR to exist).

The parallel calls sent too many requests and the client wouldn't accept them.

This code uses a solution referred to from [here](https://doc.akka.io/docs/akka-http/current/client-side/pool-overflow.html) which consist of using a streaming pipeline that will backpressure to limit the concurrent requests

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
ivanopagano: Good point, it's a utility so it would be fine to make it `private`.

I used this because of this code in the original `processBlocks`

```scala
  else if(block.metadata.header.level <= minLevel && followFork)
    ApiOperations.fetchBlock(block.metadata.header.predecessor) match {
      case Success(_)  => block :: blockSoFar
      case Failure(_)    => processBlocks(network, block.metadata.header.predecessor, minLevel, maxLevel, block :: blockSoFar, followFork)
          }
```
where it loads the predecessor block, seemingly for no other reason than to check that it's already in the db.

The corresponding excerpt in the rewritten code is
```scala
  else if (belowMinWithForking)
    ApiOperations.blockExists(current.header.predecessor) flatMap { predecessorStored =>
      if (predecessorStored)
        Future.successful(current :: collected) //pred on db, we're done here
      else
        getMetadata(network, current.header.predecessor) flatMap { pred =>
          recursiveMetadataFetch(network, pred, minLevel, maxLevel, followFork, current :: collected)
        }
      }
```

I would've simplified further to only check for `Block` existence and not for the associated `OperationGroups`, but it would've been more than a refactor, as the original code expects both to be available.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeInterface.scala
ivanopagano: It limits visibility scope of this `val` to the containing package, i.e.
` package tech.cryptonomic.conseil.tezos `

This lets me access the value in `TezosNodeOperator` where I need the `entityGetTimeout`

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeInterface.scala
```diff
@@ -100,4 +131,49 @@ object TezosNodeInterface extends TezosRPCInterface with LazyLogging {
...
+      Http().cachedHostConnectionPool[String](
+        host = host,
+        port = port,
+        settings = streamingRequestsConnectionPooling
+      )
+  }
+
+
+
+  override def runBatchedGetQuery(network: String, commands: List[String]): Future[List[String]] = {
```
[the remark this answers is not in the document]
ivanopagano: please refer to my reply on comment at `TezosNodeOperator.getAllAccountsForBlock`, down below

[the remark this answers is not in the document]
ivanopagano: As a thumb rule I'd use the batched call when a potentially significant number of http requests to tezos need to be made concurrently (about 20+), so that each requests will be queued for the bare minimum required, as defined in the `tezos-streaming-client` host pool in `application.conf`

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ivanopagano: Indeed, I was doing a different kind of rewrite, though.

As I explain in previous comment, this part of the code is here mostly to get a feedback, as I didn't expect it to be merged in the end.

I will strip this changes before removing the WIP label. The part that I really cared about, was the one to load accounts.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeInterface.scala
```diff
@@ -39,22 +60,25 @@ trait TezosRPCInterface {
...
+  private[tezos] val entityGetTimeout = conf.getInt("GET-ResponseEntityTimeoutInSeconds").seconds
+  private[tezos] val entityPostTimeout = conf.getInt("POST-ResponseEntityTimeoutInSeconds").seconds
 
   implicit val system: ActorSystem = ActorSystem("lorre-system")
   implicit val materializer: ActorMaterializer = ActorMaterializer()
   implicit val executionContext: ExecutionContextExecutor = system.dispatcher
 
-  @Override
-  def runGetQuery(network: String, command: String): Try[String] = {
+  private[this] def commandUrl(network: String, command: String): String = {
```
ivanopagano: use verbs for methods and functions

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeInterface.scala
```diff
@@ -100,4 +131,49 @@ object TezosNodeInterface extends TezosRPCInterface with LazyLogging {
...
+
+
+  override def runBatchedGetQuery(network: String, commands: List[String]): Future[List[String]] = {
+    val connections = hostPool(network)
+    val uris = Source(commands.map(commandUrl(network, _)))
+    val toRequest = (url: String) => (HttpRequest(uri = Uri(url)), url)
+
+    uris.map(toRequest)
+      .via(connections)
+      .mapAsync(5) {
```
ivanopagano: extract to config, possibly

PR #150, opened 2018-08-23
Conseil-149: Simplify ApiOperations syntax and reduce duplication

### LINKED ISSUE #149: ApiOperations Filters on joined tables could be simplified
There may be refactorings that could simplify the abstractions for filtering multiple table joins based on parameters in ApiOperations

ivanopagano:
* extract common parts of logic
 * use table rows types instead of tupled colums
 * use scala lib functions to simplify as needed

fixes #149


src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -85,79 +163,101 @@ object ApiOperations {
...
-  (table: Query[(Tables.Operations, Tables.Accounts),
-                (Tables.Operations#TableElementType, Tables.Accounts#TableElementType),
-                Seq]) extends JoinedTables
-
-  case class Operations
-  (table: Query[Tables.Operations, Tables.Operations#TableElementType, Seq]) extends JoinedTables
-
-  case class Accounts
-  (table: Query[Tables.Accounts, Tables.Accounts#TableElementType, Seq]) extends JoinedTables
+  case class BlocksOperationGroupsOperationsAccounts(
```
ba205: I'm crying at how much better this is. Now we change schemas without fear!

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -167,151 +267,52 @@ object ApiOperations {
...
 
-  case class BlocksAction
-    (action: Query[(Rep[Int], Rep[Int], Rep[String], Rep[Timestamp], Rep[Int],
-                    Rep[String], Rep[Option[String]], Rep[Option[String]], Rep[String],
-                    Rep[Option[String]], Rep[String], Rep[Option[String]]),
-                   (Int, Int, String, Timestamp, Int, String,
-                     Option[String], Option[String],
-                     String, Option[String], String, Option[String]),
-                   Seq]) extends Action
+  case class BlocksAction(action: Query[Tables.Blocks, Tables.BlocksRow, Seq]) extends Action
```
ba205: Still crying.

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -167,151 +267,52 @@ object ApiOperations {
...
-  private def filterOperationKindsForFees(filter: Filter, fee: Tables.Fees): Rep[Boolean] =
-    if (filter.operationKinds.isDefined && filter.operationKinds.get.nonEmpty)
-      fee.kind.inSet(filter.operationKinds.get) else true
+    (action: Query[Tables.OperationGroups, Tables.OperationGroupsRow, Seq]) extends Action
 
-  private def getFilterLimit(filter: Filter): Int = if (filter.limit.isDefined) filter.limit.get else 10
-
-  // End helper functions for constructing Slick queries
 
+  case class AccountsAction(action: Query[Tables.Accounts, Tables.AccountsRow, Seq]) extends Action
```
ba205: More tears of joy. Can you point out the relevant slick documentation for this, because I really wish I knew this when I first wrote this code.

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -167,151 +267,52 @@ object ApiOperations {
...
+          filterChainIDs(filter, block) &&
+          filterProtocols(filter, block)
+      })
+
+      FilteredTables(filteredAccounts, filteredBlocks, filteredOpGroups, filteredOps)
     }
+
+    for {
+      _ <- fetchLatestBlock() // is this needed?
+      maxLevelForAccounts <- fetchMaxBlockLevelForAccounts()
```
ba205: I'll have to double check the codebase, but if fetchMaxBlockLevelForAccounts still works without the latest block in scope from fetchLatestBlock, then we should be fine

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -333,41 +334,49 @@ object ApiOperations {
                               tables: FilteredTables,
                               filter: Filter): Option[JoinedTables] = {
 
-    val blocks = tables.filteredBlocks
-    val operationGroups = tables.filteredOperationGroups
-    val operations = tables.filteredOperations
-    val accounts = tables.filteredAccounts
+    val FilteredTables(accounts, blocks, operationGroups, operations)= tables
+
+    def unwrapQuadJoin[A,B,C,D](nest: (((A,B),C),D)): (A,B,C,D) = nest match {
```
ba205: So slick doesn't care about nested parentheses in type signatures when executing them? That would be such a relief.

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -403,199 +416,65 @@ object ApiOperations {
...
-                case Some(x) if x == "desc" => BlocksAction(blockAction.sortBy(_._12.desc))
-                case None => BlocksAction(blockAction.sortBy(_._12.desc))
-              }
-            case None =>
-              order.map(_.toLowerCase) match {
-                case Some(x) if x == "asc" => BlocksAction(blockAction.sortBy(_._1.asc))
-                case Some(x) if x == "desc" => BlocksAction(blockAction.sortBy(_._1.desc))
-                case None => BlocksAction(blockAction.sortBy(_._1.desc))
-              }
+        case blocks: BlocksAction =>
```
ba205: Should we eventually create ADTs for schema column names, or do you think string pattern matching is fine for this use case?

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -403,199 +416,65 @@ object ApiOperations {
     */
   private def fetchSortedAction(order: Option[String], action: Action, sortBy: Option[String]): Action = {
 
+    val sortingField: Option[String] = sortBy.map(_.toLowerCase)
+
+    def sortingOrder[A](col: ColumnOrdered[A]): ColumnOrdered[A] =
```
ba205: So slick does have some abstraction capability :) Nice find!

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
ba205: Have you done sanity tests on a few fields in postman for sorting?

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -738,49 +597,33 @@ object ApiOperations {
...
           case _ =>
-            throw new Exception("This exception should never be reached, but is included for completeness.")
+            throw new IllegalStateException("This exception should never be reached, but is included for completeness.")
         }
 
         val OperationGroupsAction(sortedAction) = fetchSortedAction(filter.order, OperationGroupsAction(action), filter.sortBy)
         val op = dbHandle.run(sortedAction.distinct.take(getFilterLimit(filter)).result)
-        val results = Await.result(op, Duration.apply(awaitTimeInSeconds, SECONDS))
-        results.map(x => Tables.OperationGroupsRow(x._1, x._2, x._3, x._4, x._5, x._6))
+        Await.result(op, awaitTimeInSeconds.seconds)
```
ba205: Is the mapping irrelevant now because  the return type of results is now a database row, thanks to the mapping in the pattern matching plus your changing of the Joined Tables type signatures?

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -846,13 +672,23 @@ object ApiOperations {
...
     */
   def fetchAverageFees(filter: Filter): Try[AverageFees] = Try {
-    val action = for {
-      fee <- Tables.Fees
-      if filterOperationKindsForFees(filter, fee)
-    } yield (fee.low, fee.medium, fee.high, fee.timestamp, fee.kind)
-    val op = dbHandle.run(action.distinct.sortBy(_._4.desc).take(1).result)
-    val results = Await.result(op, Duration.apply(awaitTimeInSeconds, SECONDS)).head
-    AverageFees(results._1, results._2, results._3, results._4, results._5)
+    val action =
```
ba205: Beautiful, this entire function just makes me happy

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -903,29 +736,21 @@ object ApiOperations {
...
           case _ =>
-            throw new Exception("You can only filter accounts by operation ID, operation source, account ID, account manager, account delegate, or inner and outer operation kind.")
+            throw new IllegalArgumentException("You can only filter accounts by operation ID, operation source, account ID, account manager, account delegate, or inner and outer operation kind.")
         }
 
         val AccountsAction(sortedAction) = fetchSortedAction(filter.order, AccountsAction(action), filter.sortBy)
         val op = dbHandle.run(sortedAction.distinct.take(getFilterLimit(filter)).result)
-        val results = Await.result(op, Duration.apply(awaitTimeInSeconds, SECONDS))
-        results.map(x => Tables.AccountsRow(x._1, x._2, x._3, x._4, x._5, x._6, x._7, x._8, x._9, x._10))
+        Await.result(op, awaitTimeInSeconds.seconds)
```
ba205: If this works, this is a significant improvement in maintainability and readability. Really awesome stuff :)

ba205: *this refers to the entire file.

ivanopagano:
I might need to check this as filtering seems to misbehave locally

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -43,23 +43,101 @@ object ApiOperations {
...
-                     protocols: Option[Set[String]] = Some(Set[String]()),
-                     operationGroupIDs: Option[Set[String]] = Some(Set[String]()),
-                     operationSources: Option[Set[String]] = Some(Set[String]()),
-                     operationDestinations: Option[Set[String]] = Some(Set[String]()),
-                     operationParticipants: Option[Set[String]] = Some(Set[String]()),
-                     operationKinds: Option[Set[String]] = Some(Set[String]()),
-                     accountIDs: Option[Set[String]] = Some(Set[String]()),
-                     accountManagers: Option[Set[String]] = Some(Set[String]()),
-                     accountDelegates: Option[Set[String]] = Some(Set[String]()),
+                     limit: Option[Int] = Some(defaultLimit),
```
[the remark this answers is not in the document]
ivanopagano: Removing duplicate code looks cleaner _to me_

Also the `defaultLimit` will be easier to pick up as a variable, and change it later, than having it buried within a block of code. Pulling out hardcoded values like this is usually considered best-practice

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -70,10 +148,10 @@ object ApiOperations {
     * @param filteredOperations       Filtered Operations table
    */
   case class FilteredTables(
-                             filteredAccounts: Query[Tables.Accounts, Tables.Accounts#TableElementType, Seq],
-                             filteredBlocks: Query[Tables.Blocks, Tables.Blocks#TableElementType, Seq],
-                             filteredOperationGroups: Query[Tables.OperationGroups, Tables.OperationGroups#TableElementType, Seq],
-                             filteredOperations: Query[Tables.Operations, Tables.Operations#TableElementType, Seq]
+                             filteredAccounts: Query[Tables.Accounts, Tables.AccountsRow, Seq],
```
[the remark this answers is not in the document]
ivanopagano: I actually followed the types, if you look into the definition of the `Tables.Accounts` class for instance, you can figure out that the internal `TableElementType` type definition is fixed to `AccountsRow`.

And I guess that this modeling is done exactly to simplify the kind of complicated tuples mapping you had to handle previously

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
ivanopagano: Great then!

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
ivanopagano: I think I just looked into the code definitions and the types coming out from the queries, and I just matched those I guess.

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
ivanopagano: Technically they can't influence each other, and if they do I would take it as a design shortcoming and try to change that to make it explicit.

The fetch is done on db for both operations, so I won't expect any side-effect from the first call to influence the second one.

The only semantic reason might be that we want to be sure that a _latest block exists on the DB_... but even if it doesn't, the max level for accounts would simply be the default, and I can't see anything wrong with that.

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
ivanopagano: Well I'm not sure how to read your comment.

Slick do care about the nesting when it handles data, because it determines the _shape_ of input and output values.

But _we_ don't care in the end, because the results of queries can always be mapped to a more useful format, just like when you use it to make projections on the columns using `select a, b, c from ...`

When joining the resulting rows are packed within nested tuples which are probably more amenable to be composed over if you need to bolt on other joins later, this is my guess at why the library authors did it like this.

But for our goals, it's easier to put all row types in a single flattened tuple. We do the mapping only at the end, because I don't think the `join` signatures would match so easily anymore.

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
ivanopagano: It does, but I'm afraid it mostly does for internal convenience of the library itself, and those details are hard to find on the docs or any other written source as useful concepts available to the user.

Personally I think this is the worst aspect of the library, it's rather complex under the hood (with all his compile-time abstractions) and there's no easy way to learn how to tweak it as needed, without learning how the code itself works...

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
ivanopagano: I'm doing now, and In fact I think I've found some problem that I'm looking into to solve.

ivanopagano: Right now I wouldn't go so far as to define a type for each column.

It will turn useful to use a "newtype wrapper" for the `String` to distinguish between sorting-field names and sorting-order for example... maybe you meant that?

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
ivanopagano: Yes, if I got your comment correctly.

The JoinedTables now maps to tuples of xxxRow results, and once you get the table-specific row type out of the tuple, that has already the matching fields needed

vishakh:
Closing in favor of #151.

Issue #127, opened 2018-07-23
Implement miscellaneous Scalac code review comments


vishakh:
Scalac conducted a Conseil code review in June. This ticket covers all items in the report not covered by other tickets.

ivanopagano:
I've checked the code review doc and verified the following points:

 - Database threads: the connection pool is handled by the underlyng HikariCP based on threads assigned to the AsyncHandler. Such pool is set to `10` which might be a reasonable value for a quad-core machine along the formula provided [here](https://github.com/brettwooldridge/HikariCP/wiki/About-Pool-Sizing#connections--core_count--2--effective_spindle_count), which should be applied to thread-count in this case
 - Akka threads: at the moment different PRs are introducing usage of the actor-system dispatcher as default and dedicated thread-pools for specific use cases, so this should be handled
 - Coding best practices are progressively introduced in PRs related to async-database access and block processing refactors
 - Akka Http: currently remapping the `Filter` case class to better match the requests parameters [IN PROGRESS]
 - Use of Strings: will provide classes (a.k.a. newtypes) to wrap strings that should avoid subtle bugs [TODO]
 - Splitting of `TezosNodeOperator` class [TO BE EVALUATED]

The advice related to extraction of SSL termination remains unsatisfied as it involves devops involvement.

ivanopagano:
The last point is deferred to another issue

ivanopagano:
The remaining last point is deferred to separate issue

PR #156, opened 2018-09-21
Conseil-127 misc refactors

ivanopagano:
Improve the code by taking into account miscellaneous advises from Scalac's review document

solves #127 

Considered for inclusion
 - Conversion of the Filter parameters defined as optional sets to unwrapped sets, where the semantic for a `None` and an empty `Set` are collapsed
 - Addition of a smart constructor that converts http params to the scala `Filter` instance
 - Usage of a fully typed object to describe ordering of conseil results instead of simple strings
 - Usage of fully typed wrappers to define hashes of blocks and operation groups, and account ids
 - Rejection of invalid `order` parameters

Non included points are documented in the issue page itself


src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -23,8 +23,23 @@ object ApiOperations {
   private val awaitTimeInSeconds = conf.getInt("dbAwaitTimeInSeconds")
   lazy val dbHandle: Database = DatabaseUtil.db
 
+  /** Define sorting order for api queries */
```
ba205: Really solid improvement, was thinking about this for a while

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -191,85 +242,85 @@ object ApiOperations {
...
-      (filter.protocols.isDefined && filter.protocols.get.nonEmpty)
+    filter.blockIDs.nonEmpty ||
+      filter.levels.nonEmpty ||
+      filter.chainIDs.nonEmpty ||
+      filter.protocols.nonEmpty
 
   private def isOperationGroupFilter(filter: Filter): Boolean =
-    (filter.operationGroupIDs.isDefined && filter.operationGroupIDs.get.nonEmpty) ||
-      (filter.operationSources.isDefined && filter.operationSources.get.nonEmpty)
+    filter.operationGroupIDs.nonEmpty ||
```
ba205: would it be perhaps helpful to make a single "isXFilter" function, because besides slight variations, the majority of the logic here is repeated? Or do you think that would be over engineering it?

src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
```diff
@@ -30,31 +30,13 @@ object Tezos extends LazyLogging {
...
-      operationDestinations = Some(op_destinations.toSet),
-      operationParticipants = Some(op_participants.toSet),
-      operationKinds = Some(operation_kind.toSet),
-      accountIDs = Some(account_ids.toSet),
-      accountManagers = Some(account_managers.toSet),
-      accountDelegates = Some(account_delegates.toSet),
-      sortBy = sort_by, order = order)
-    provide(filter)
-  }
+  ).as(Filter.readParams)
```
ba205: Way cleaner, love it

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
ivanopagano: indeed, but it's already done in one of the other pending PRs, so I'm reluctant to refactor here again, I'd solve the conflict when the PRs are merged

src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
ivanopagano: actually is just putting the dirt under the rug, but at least the client's cleaner

PR #159, opened 2018-09-27
Conseil 88: async db calls 2

### LINKED ISSUE #88: Conseil should not block on any database operations

ivanopagano:
closes #88 

This second step includes missing changes to make all database operations return async results, combined there and then.
Only the main app should wait for results now.


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -1,22 +1,22 @@
 
 package tech.cryptonomic.conseil
 
+import akka.Done
 import com.typesafe.config.ConfigFactory
 import com.typesafe.scalalogging.LazyLogging
-import tech.cryptonomic.conseil.tezos._
+import tech.cryptonomic.conseil.tezos.{FeeOperations, TezosNodeInterface, TezosNodeOperator, TezosDatabaseOperations => TezosDb, _}
```
vishakh: How come you use this alias here?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -51,61 +52,64 @@ object Lorre extends App with LazyLogging {
...
           noOp
     } yield ()
+
     Await.ready(processing, atMost = Duration.Inf)
     logger.info("Taking a nap")
     Thread.sleep(sleepIntervalInSeconds * 1000)
     mainLoop(iteration + 1)
   }
 
+  sys.addShutdownHook{
```
vishakh: Is this block in conflict with #154?

src/main/scala/tech/cryptonomic/conseil/tezos/Tables.scala
```diff
@@ -204,6 +204,9 @@ trait Tables {
   /** Collection-like TableQuery object for table OperationGroups */
   lazy val OperationGroups = new TableQuery(tag => new OperationGroups(tag))
 
+  val operationGroupsByHash =
```
vishakh: This is a generated file produced using Slick via `GenSchema.scala`. We should move any custom logic out of here or make these changes in the database schema so this functionality is also also auto-generated.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: I found it noisy at usage site

I'll revert if it's confusing

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: Yes sure, will need to properly merge.

I couldn't keep track anymore of where each change was made, so duplication crops out here and there.

src/main/scala/tech/cryptonomic/conseil/tezos/Tables.scala
ivanopagano: Sorry, my bad, didn't notice, I'll take care

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
vishakh: Generally, we try to avoid aliasing but we can see your point.

Issue #48, opened 2018-01-22
Lorre should invalidate discarded blocks in the database when it detects forks


vishakh:
Every time Lorre detects a fork in the Tezos chain, it should invalidate discarded blocks using an 'invalidated_blocks' table.

vishakh:
Closing for now. Will reopen in the future when bandwidth permits.

PR #161, opened 2018-10-12
Conseil 48 invalidate blocks during forks [Not merging yet]

ba205:
#48 

Updated getBlocks to handle forking functionality, which employs both collecting blocks in detected fork in a Future, as well as writing to an Invalidated Blocks table. SQL and Slick Schema updated to include InvalidatedBlocks table. Doc written to document Lorre's core algorithm.


doc/Lorre's Block Fetching Algorithm
```diff
@@ -0,0 +1,96 @@
...
+---------------------------------------------------------------------------------------------------------
+
+As long as assumptions are valid and the invariant is maintained, this algorithm will work as intended:
+
+[i] In a non-forked chain, there can only be one block at each level.
+
+[ii] Internal consistency with our database is all we can hope for, as there is no notion of a
+     globally correct chain.
+
+[iii] At any given point in time, there will only be at most one block at a given level
```
ivanopagano: ...at most one block at _any_ given level...

doc/Lorre's Block Fetching Algorithm
```diff
@@ -0,0 +1,96 @@
...
+
+------------------------------------------------------------------------------------------------------
+
+Lorre's algorithm for fetching blocks that occur as a result of a fork:
+
+The following logic is implemented via the followFork function, which is a private helper of the
+more general getBlocks function.
+
+In the normal case, we place blocks into a blocks table, with the fields being all of the associated
+metadata. In the forked case, we will have an invalidated blocks table to invalidate blocks that are
```
ivanopagano: Not clear to me if, in case of a detected fork, all blocks at any forked level should be present in the `Invalidated` table... or only the invalidated ones?

doc/Lorre's Block Fetching Algorithm
```diff
@@ -0,0 +1,96 @@
...
+globally correct chain. [ii]
+
+Suppose a fork happens, and an iteration of Lorre collects a batch of blocks with levels m, m+1...n.
+Blocks at levels m to n are added to the blocks table. we will then consider m's predecessors, as those
+are the blocks that may conflict with our internal view of what the current chain looks like.
+
+For the block at level m-1, we perform two database checks: Is this block currently in the Blocks Table,
+and is this block currently in the Invalidated Blocks Table?
+
+Case 1: Block at level m-1 is in the blocks table, but is not in the invalidated blocks table.
```
ivanopagano: just a trivial thought, does the different Cases (1,2,3,4) have any easy-to-remind name to talk about them? e.g.
 - case 1 unforked block
 - case 2 forked substitute
 - case 3 restored block
 - case 4 inconsistent

doc/Lorre's Block Fetching Algorithm
```diff
@@ -0,0 +1,96 @@
...
+We then continue the same checks for blocks m-2, m-3...until Case 1, our base case, has been reached. Then,
+we are done collecting forked blocks for writing, as well as keeping our invalidated blocks table up to speed.
+
+---------------------------------------------------------------------------------------------------------
+
+As long as assumptions are valid and the invariant is maintained, this algorithm will work as intended:
+
+[i] In a non-forked chain, there can only be one block at each level.
+
+[ii] Internal consistency with our database is all we can hope for, as there is no notion of a
```
ivanopagano: What do we mean exactly for "internal consistency" at this point, it's a very general term, it might be worthy here (at least in my opinion) to define a precise meaning for an internally consistent conseil db

doc/Lorre's Block Fetching Algorithm
```diff
@@ -0,0 +1,96 @@
...
+i) A regular, unforked chain
+
+ii) A chain with one fork, or two branches
+
+iii) A chain with one fork, and Lorre's iterations happen to go back and forth on either one of these chains.
+
+iv) A chain with two forks, or three branches, ending on the second fork/third branch
+
+v) A chain with two forks, or three branches, and Lorre's iterations go back and forth between all three of
+these chains, until it's decided that the second branch is the true internal chain for an extended period of time.
```
ivanopagano: how do we decide about a long-term internal chain? what would be the conditions?

doc/Lorre's Block Fetching Algorithm
```diff
@@ -0,0 +1,96 @@
...
+     globally correct chain.
+
+[iii] At any given point in time, there will only be at most one block at a given level
+      with is_invalidated set to false in the invalidated blocks table
+
+-----------------------------------------------------------------------------------------------------------
+
+Spec-ing Unit Tests for the Forking Algorithm:
+
+GetBlocks should correctly process the following chains:
```
ivanopagano: It might be useful to try and (at least) approach the unhandled cases too, and define a proper protocol to handle all such consistency scenarios

doc/Lorre's Block Fetching Algorithm
```diff
@@ -0,0 +1,96 @@
...
+
+Lorre's algorithm for fetching blocks that occur as a result of a fork:
+
+The following logic is implemented via the followFork function, which is a private helper of the
+more general getBlocks function.
+
+In the normal case, we place blocks into a blocks table, with the fields being all of the associated
+metadata. In the forked case, we will have an invalidated blocks table to invalidate blocks that are
+not internally consistent with the new chain being recognized by Lorre. Here, the necessary fields are
+block_id (the block's hash), level, and an is_invalidated boolean flag.
```
ivanopagano: what are the chances of conflicting blocks with the same block hash and level?

doc/conseil.sql
```diff
@@ -202,6 +213,14 @@ ALTER TABLE ONLY public.blocks
     ADD CONSTRAINT blocks_hash_key UNIQUE (hash);
 
 
+--
+-- Name: invalidated_blocks invalidated_blocks_pkey; Type: CONSTRAINT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY public.invalidated_blocks
```
ivanopagano: What about a foreign-key from invalidated_blocks.hash to blocks.hash?

It would eliminate the possibility of a persisting the case numbered 4 on the documentation.

Will it be too restrictive in a way that I'm not seeing now?

doc/Lorre's Block Fetching Algorithm
ba205: only blocks that have been invalidated at least once will be in the invalidated_blocks table. The idea is at level n, a block is the "true" block if either it is not in the invalidated blocks table, or it is in the invalidated blocks table with it's is_invalidated field set to false. The assumed invariant is that, at any given point in time, there is at most one block at a given level that has is_invalidated set to false in the invalidated blocks table

doc/Lorre's Block Fetching Algorithm
ba205: same block hash will never happen (the probability is astronomically low). If there are two blocks at the same level, a fork must have occurred, hence the need for this function

doc/Lorre's Block Fetching Algorithm
ba205: I'll look into this! I'll need to take care with proper, non misleading naming

doc/Lorre's Block Fetching Algorithm
ba205: An internally consistent mapping of the blockchain in our database must either have exactly one block per level, or if there are multiple blocks in a level, there is at most one block at that level in the invalidated_blocks table with their is_invalidated flag set to false, i.e., there is only one valid block per level.

doc/Lorre's Block Fetching Algorithm
ba205: Right! Thanks

doc/Lorre's Block Fetching Algorithm
ba205: Totally! Testing this thoroughly will be a nontrivial task

doc/Lorre's Block Fetching Algorithm
ba205: See above definition of an internally consistent database mapping of the blockchain

doc/conseil.sql
ba205: Not sure, will look into it, but not a bad idea.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -224,59 +235,92 @@ class TezosNodeOperator(val node: TezosRPCInterface)(implicit executionContext:
...
-
-    val predicate: Future[Block] => (Boolean, Boolean) =
-      futureBlock => {
-        val block = Await.result[Block](futureBlock, Duration.apply(awaitTimeInSeconds, SECONDS))
-        (blockExists(block), blockInInvalidatedBlocks(block))
+                  hash: BlockHash
+                ): Future[List[BlockWithAction]] = {
+
+    //Move some function to utils
+    def futureToIO[A](f: Future[A]): IO[A] = IO.fromFuture(IO(f))
```
ivanopagano: I think the `f` parameter should better be _by-name_ as in `f: => Future[A]`

This way the future won't start computing until the `IO` result gets evaluated (which is how you make it functionally pure)
Not critical but it might break assumptions down the road

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -224,59 +235,92 @@ class TezosNodeOperator(val node: TezosRPCInterface)(implicit executionContext:
...
       }
 
-    offsets
-      .map{ offset => {getBlock(network, hash, Some(offset))}}
-      .takeWhile{ block => predicate(block) != (true, false)}
-      .foldLeft(Future.successful(List.empty[Block]))(foldFunction)
+      for {
+        block <- getBlockIO(network, hash, Some(offset))
+        tuple <- predicate(block)
+        //scala compiler complained about tuples not having a withFilter instance with IO
```
ivanopagano: I had a chance to verify that this is related to binding with pattern-matching: the compiler rewrites that as `withFilter` calls to check that the pattern works

[further reference](https://github.com/oleg-py/better-monadic-for#desugaring-for-patterns-without-withfilters)

vishakh:
Deprecated

Issue #103, opened 2018-06-27
Conseil should have (ideally) no partial functions


ba205:
One of the comments on our code review was that there are several places in which our code will not terminate on all inputs, and we throw exceptions when we can use concrete types instead. Refactoring the codebase in this way should help the long-term health of the system

vishakh:
Addressed by #151

ivanopagano:
@vishakh I'm trying to better understand why you say that #151 addresses the issue?

Do you mean that it handles part of the problem or it solves it?

As I see it, there are still places where an internal failure on the input is returned as a `500 Internal Server Error` from Conseil. Is it the expected outcome from the frontend/client?

ba205:
Hi @ivanopagano, Vishakh mentioned that at my request. It solves a big chunk of the problem, as APIOperations is a big module, but there may be other partial functions lurking around in other modules.

ivanopagano:
Need to refactor the JsonUtil api to return a possibly failing result of parsing a json string.

It could return a `Try[T]` that have to be handled by the calling code

PR #167, opened 2018-10-19
Conseil-103: no partial functions

ivanopagano:
Removed possible api operation failures due to missing results, now handled at the http layer with a `404 Not Found` response code.

This includes fetching by hash or fetching fees with no matching filter.

Doing so we reduced the surface of functions that may throw exceptions for some inputs as per #103 

To check for other possible sources of inconsistencies or errors we introduce the usage of a plugin (scapegoat) that can check for potential code-style-level issues, and added several commonly advised compiler flags that warn about other sources of mistakes.

To allow usage of the sbt plugin, we **update the version of sbt to the latest available**. No visible impact was detected doing so.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,6 +5,9 @@ scalaVersion := "2.12.4"
 val akkaHttpVersion = "10.1.0"
 val akkaVersion = "2.5.11"
 
+scapegoatVersion in ThisBuild := "1.3.8"
+scapegoatIgnoredFiles := Seq(".*/tech/cryptonomic/conseil/tezos/Tables.scala")
+
 libraryDependencies  ++=  Seq(
   "ch.qos.logback" % "logback-classic" % "1.2.3",
   "com.typesafe.akka" %% "akka-http" % akkaHttpVersion,
@@ -33,3 +36,5 @@ excludeDependencies ++= Seq(
 )
 
 assemblyOutputPath in assembly := file("/tmp/conseil.jar")
+
+scalacOptions ++= ScalacOptions.common
--- a/src/main/scala/tech/cryptonomic/conseil/db/DatabaseQueryExecution.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/db/DatabaseQueryExecution.scala
@@ -21,100 +21,100 @@ object DatabaseQueryExecution {
 
   sealed trait JoinedTables
 
-  case class BlocksOperationGroupsOperationsAccounts(
+  final case class BlocksOperationGroupsOperationsAccounts(
     join: Query[
       (Tables.Blocks, Tables.OperationGroups, Tables.Operations, Tables.Accounts),
       (Tables.BlocksRow, Tables.OperationGroupsRow, Tables.OperationsRow, Tables.AccountsRow),
       Seq]
   ) extends JoinedTables
 
-  case class BlocksOperationGroupsOperations(
+  final case class BlocksOperationGroupsOperations(
     join: Query[
       (Tables.Blocks, Tables.OperationGroups, Tables.Operations),
       (Tables.BlocksRow, Tables.OperationGroupsRow, Tables.OperationsRow),
       Seq]
   ) extends JoinedTables
 
-  case class BlocksOperationGroupsAccounts(
+  final case class BlocksOperationGroupsAccounts(
     join: Query[
       (Tables.Blocks, Tables.OperationGroups, Tables.Accounts),
       (Tables.BlocksRow, Tables.OperationGroupsRow, Tables.AccountsRow),
       Seq]
   ) extends JoinedTables
 
-  case class BlocksOperationGroups(
+  final case class BlocksOperationGroups(
     join: Query[
       (Tables.Blocks, Tables.OperationGroups),
       (Tables.BlocksRow, Tables.OperationGroupsRow),
       Seq]
   ) extends JoinedTables
 
-  case class BlocksOperationsAccounts(
+  final case class BlocksOperationsAccounts(
     join: Query[
       (Tables.Blocks, Tables.Operations, Tables.Accounts),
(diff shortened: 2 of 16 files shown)


build.sbt
```diff
@@ -5,6 +5,9 @@ scalaVersion := "2.12.4"
 val akkaHttpVersion = "10.1.0"
 val akkaVersion = "2.5.11"
 
+scapegoatVersion in ThisBuild := "1.3.8"
```
ba205: Why are we putting the version for scapegoat in here, and not in libraryDependencies?

src/main/scala/tech/cryptonomic/conseil/db/DatabaseQueryExecution.scala
```diff
@@ -21,100 +21,100 @@ object DatabaseQueryExecution {
 
   sealed trait JoinedTables
 
-  case class BlocksOperationGroupsOperationsAccounts(
+  final case class BlocksOperationGroupsOperationsAccounts(
```
ba205: What's the reasoning behind making all of these case classes final? What does this impact?

src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
```diff
@@ -62,49 +61,74 @@ object Tezos {
...
-  implicit def jsonStringMarshaller[T]: ToEntityMarshaller[T] =
+  //add the correct content-type for [[JsonUtil]]-converted values
+  implicit val jsonMarshaller: ToEntityMarshaller[JsonString] =
     PredefinedToEntityMarshallers.StringMarshaller
-      .compose(JsonUtil.toJson[T])
-      .wrap(MediaTypes.`application/json`)(identity)
+      .compose((_: JsonString).json)
+      .wrap(MediaTypes.`application/json`)(identity _)
+
+  /* allow generic handling of optional results, embedded in async computations
```
ba205: Did you type all of this out yourself? If you type "/**<enter>" in IntelliJ, it automatically handles parameter docs. I'd suggest adding that extra start on the top part for proper JavaDoc syntax.

src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
```diff
@@ -62,49 +61,74 @@ object Tezos {
...
+   * in addition to converting any missing result to a NotFound http code, it allows to convert the existing content
+   * to something which is marshallable as a response
+   * @param operation is the computation that will provide, as soon as available, an optional result
+   * @param converter a final conversion function to turn the original T, when available to a marshallable result,
+   *        by default the function converts to a [[JsonString]]
+   * @param T the type of the possible result of the async computation
+   * @param R the final outcome, which must be compatible with an available [[ToResponseMarshaller]]
+   */
+  private[this] def handleNoneAsNotFound[T, R: ToResponseMarshaller](operation: => Future[Option[T]], converter: T => R = toJson[T] _): Future[ToResponseMarshallable] =
```
ba205: Small note, since you've been using module/object specific private modifiers fairly consistently, it'll probably be better if this is consistent across the entire codebase and any future privates should be module specific. Double check and make sure there are no standalones for this PR.

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -14,23 +13,24 @@ import scala.concurrent.{ExecutionContext, Future}
...
   object Sorting {
+
     /** Read an input string (`asc` or `desc`) to return a
       * (possible) [[tech.cryptonomic.conseil.tezos.ApiOperations.Sorting]] value
       */
     def fromString(s: String): Option[Sorting] = s.toLowerCase match {
       case "asc" => Some(AscendingSort)
       case "desc" => Some(DescendingSort)
-      case _ => throw new IllegalArgumentException("""Please provide a valid value for the ordering parameter ['asc', 'desc'] """)
+      case _ => None
```
ba205: I like it!

build.sbt
ivanopagano: it's a plugin for the build tool, not a dependency for code compilation

ivanopagano: sbt uses this version to know what code to use when you run the `scapegoat` task or any other added from the plugin itself

src/main/scala/tech/cryptonomic/conseil/db/DatabaseQueryExecution.scala
ivanopagano: You can't override them in any way using the `extends` mechanism, which is what you commonly want for defining "closed enumerations of types" as in ADTs.

More generally extending a case class usually breaks the methods automatically provided for you (e.g. `equals`, `hashcode`, `copy`), which are based on the original case class's fields for example

Look into this [SO answer](https://stackoverflow.com/questions/34561614/should-i-use-the-final-modifier-when-declaring-case-classes) for other details

src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
ivanopagano: Oh, yes, sorry. I'm doing it manually.

Actually I've seen that javadoc syntax is used in the codebase, but scaladoc syntax have different conventions in this regard: https://docs.scala-lang.org/style/scaladoc.html

That's why I might be making mistakes at times

src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
ivanopagano: makes sense

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
ivanopagano: it was a mistake in the first place

PR #173, opened 2018-10-25
Conseil-103: continued work on avoiding partial functions

### LINKED ISSUE #103: Conseil should have (ideally) no partial functions

ivanopagano:
There are more cases of partial results based on complexity of call scenarios.

This PR tries to address some of those to fix #103 

 * Use smart constructor to allow only valid json inside the `JsonString`
 * Use a typed configuration library to define parameters, and load them up-front
 * Check behaviour on unsupported/incorrect `operation_kinds` (no clear results here)

Additional subtle cases might arise from weird combinations of filtering parameters in Conseil's query endpoints and which sql joins are considered valid.

A separate issue will be opened to verify those.

**important notice**
To overcome a [scala bug](https://github.com/scala/scala/pull/6190) using one of the [commit changes](https://github.com/Cryptonomic/Conseil/pull/173/commits/960bf95913c76d2c8949ea73f1dc8ca67afe7b01), this request includes **an update of the scala compiler version to 2.12.8**, consider carefully if it may impact the system in so-far unforeseen ways


ivanopagano:
@vishakh @ba205 
If you dig into the new way configuration-handling is introduced into this PR, you'll see that there's a `network` parameter passed along to all Tezos Node interface, only to be ignored in favor of the single argument passed as input to the run script (sbt run, or java ... command) for `Lorre`.

Is it still useful to think of having a global network to run against from the launch and a different one to specify when calling the actual node operations?
Should the node interface remove that extra parameter now?

build.sbt
```diff
@@ -9,27 +9,28 @@ scapegoatVersion in ThisBuild := "1.3.8"
...
+  "ch.qos.logback"                   % "logback-classic"      % "1.2.3",
+  "com.typesafe.akka"               %% "akka-http"            % akkaHttpVersion,
+  "com.typesafe.akka"               %% "akka-stream"          % akkaVersion,
+  "com.typesafe.akka"               %% "akka-actor"           % akkaVersion,
+  "com.typesafe"                     % "config"               % "1.3.1",
+  "com.typesafe.scala-logging"      %% "scala-logging"        % "3.7.2",
+  "com.github.pureconfig"           %% "pureconfig"           % "0.9.2",
+  "org.scalaj"                       % "scalaj-http_2.12"     % "2.3.0",
+  "org.scalatest"                   %% "scalatest"            % "3.0.4" % "test",
+  "com.fasterxml.jackson.core"       % "jackson-databind"     % "2.9.0",
```
piotrkosecki: as you reformatted those dependencies, can you extract versions for the `jackson` and `slick` + remove `_2.12` from `scalaj`? Just small cosmetic changes.

ivanopagano: sure

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -19,13 +20,14 @@ import scala.util.{Failure, Success}
...
 
-  //keep this import here to make it evident where we spawn our async code
-  implicit val system: ActorSystem = ActorSystem("lorre-system")
-  implicit val dispatcher = system.dispatcher
-
-  //how long to wait for graceful shutdown of system components
-  private[this] val shutdownWait = 10.seconds
+  final case class CombinedConfiguration(
+    lorre: LorreConfiguration,
+    tezos: TezosConfiguration,
```
ba205: Would this have to be generalized the second we add another blockchain?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -35,17 +37,46 @@ object Lorre extends App with LazyLogging {
       sys.exit(1)
     }
 
-  private val conf = ConfigFactory.load
-  private val sleepIntervalInSeconds = conf.getInt("lorre.sleepIntervalInSeconds")
-  private val feeUpdateInterval = conf.getInt("lorre.feeUpdateInterval")
-  private val purgeAccountsInterval = conf.getInt("lorre.purgeAccountsInterval")
+  //reads all configuration upstart, will only complete if all values are found
+  val CombinedConfiguration(lorreConf, tezosConf, sodiumConf, batchingConf) = {
```
ba205: Maybe this can be moved to a ConfigUtil file? We want Lorre.scala to be minimal and do exactly what it needs to be run and be self explanatory with it's naming.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: It depends... as long as we have different classes/packages for separate networks, they will require separate custom configurations too.

Once we find a globally shared configuration fitting different chains, it will be appropriate to rename these configurations as well

ivanopagano: Since adding a new chain is not negligible work, I assume that updating the configuration objects will not be the pain point, but just one additional detail in the total cost of the operation

vishakh:
I am getting compile errors on my Ubuntu desktop. Any idea what the cause could be? The Travis build is succeeding! :thinking: 

image

src/main/scala/tech/cryptonomic/conseil/config/ConseilAppConfig.scala
```diff
@@ -0,0 +1,28 @@
+package tech.cryptonomic.conseil.config
+
+import tech.cryptonomic.conseil.config.Platforms._
+import pureconfig.generic.auto._
+
+
+/** wraps all configuration needed to run Conseil */
+trait ConseilAppConfig {
```
vishakh: Out of curiosity, why are these traits and not just regular behaviours of Conseil, Lorre, etc? Will multiple classes ever inherit these traits?

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -7,61 +7,61 @@ import akka.http.scaladsl.model.StatusCodes._
...
+        .onComplete(_ => logger.info("We're done here, nothing else to see"))
+      }
 
-  sys.addShutdownHook {
-    bindingFuture
-    .flatMap(_.unbind().andThen{ case _ => logger.info("Server stopped...")} )
-    .flatMap( _ => system.terminate())
-    .onComplete(_ => logger.info("We're done here, nothing else to see"))
+    case Left(errors) =>
+      //nothing to do
```
vishakh: Should the error be outputted here rather than the config loader?

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -7,61 +7,61 @@ import akka.http.scaladsl.model.StatusCodes._
...
+      }
 
-  implicit val system: ActorSystem = ActorSystem("conseil-system")
-  implicit val materializer: ActorMaterializer = ActorMaterializer()
-  implicit val executionContext: ExecutionContextExecutor = system.dispatcher
-  val sslConfig = AkkaSSLConfig()
+      implicit val system: ActorSystem = ActorSystem("conseil-system")
+      implicit val materializer: ActorMaterializer = ActorMaterializer()
+      implicit val executionContext: ExecutionContextExecutor = system.dispatcher
+      val sslConfig = AkkaSSLConfig() //Should get rid of this??
```
vishakh: Yes, please kill without mercy. :grin:

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -17,34 +17,32 @@ import scala.util.{Failure, Success}
...
   */
-object Lorre extends App with TezosErrors with LazyLogging {
+object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig {
 
-  //keep this import here to make it evident where we spawn our async code
+  //reads all configuration upstart, will only complete if all values are found
+  val config = loadApplicationConfiguration(args)
+
+  //stop if conf is not available
+  config.left.foreach { _ => sys.exit(1) }
```
vishakh: Once again, should the error output to the logs be written here instead of inside?

src/main/scala/tech/cryptonomic/conseil/config/LorreAppConfig.scala
```diff
@@ -0,0 +1,67 @@
...
+    //applies convention to uses CamelCase when reading config fields
+    implicit def hint[T] = ProductHint[T](ConfigFieldMapping(CamelCase, CamelCase))
+
+    val loadedConf = for {
+      network <- readArgs(commandLineArgs)
+      lorre <- loadConfig[LorreConfiguration](namespace = "lorre")
+      nodeRequests <- loadConfig[NetworkCallsConfiguration]("")
+      node <- loadConfig[TezosNodeConfiguration](namespace = s"platforms.tezos.$network.node")
+      streamingClient <- loadAkkaStreamingClientConfig(namespace = "akka.tezos-streaming-client")
+      sodium <- loadConfig[SodiumConfiguration](namespace = "sodium.libraryPath")
```
vishakh: Sodium config is needed for Conseil and not Lorre. libSodium is used to cryptographically sign blockchain transactions. In our case, clients used to be able to make a Conseil call for transaction signing. We took it out in order to avoid security issues with our launch. However, we plan to bring it back for those using the JAR (i.e. running locally) but not for those using the API (client / server comms).

src/main/scala/tech/cryptonomic/conseil/config/package.scala
```diff
@@ -0,0 +1,15 @@
+package tech.cryptonomic.conseil
+
+import com.typesafe.scalalogging.LazyLogging
+
+package object config extends LazyLogging {
```
vishakh: Is the package object config (all in lower case) some special type of object in Scala?

src/main/scala/tech/cryptonomic/conseil/tezos/FeeOperations.scala
```diff
@@ -44,10 +44,10 @@ object FeeOperations extends LazyLogging {
     * Calculates average fees for each operation kind and stores them into a fees table.
     * @return
     */
-  def processTezosAverageFees()(implicit ex: ExecutionContext): Future[Option[Int]] = {
+  def processTezosAverageFees(numberOfFeesAveraged: Int)(implicit ex: ExecutionContext): Future[Option[Int]] = {
```
vishakh: Please document this new parameter.

vishakh:
Loving the typed configurations! Bravo! :clap:

vishakh:
> @vishakh @ba205
> If you dig into the new way configuration-handling is introduced into this PR, you'll see that there's a `network` parameter passed along to all Tezos Node interface, only to be ignored in favor of the single argument passed as input to the run script (sbt run, or java ... command) for `Lorre`.
> 
> Is it still useful to think of having a global network to run against from the launch and a different one to specify when calling the actual node operations?
> Should the node interface remove that extra parameter now?

Lorre should take a single platform and network, e.g. `tezos` and `mainnet`, and use that consistently throughout. Conseil should be stateless, i.e. every call should specify from the client the platform and network to be used. The `network` parameter should be passed respectively to the various classes.

src/main/scala/tech/cryptonomic/conseil/config/ConseilConfig.scala
```diff
@@ -0,0 +1,52 @@
...
+    numberOfFeesAveraged: Int
+  )
+
+  final case class BatchFetchConfiguration(
+    accountConcurrencyLevel: Int,
+    blockOperationsConcurrencyLevel: Int
+  )
+
+  /** configurations related to tezos-node network node calls */
+  final case class TezosRequestsConfiguration(
```
ba205: As we add more configurations to different blockchains, are we just going to add them all to this file? Would it make sense to split up chain specific configs?

src/main/scala/tech/cryptonomic/conseil/config/ConseilAppConfig.scala
ivanopagano: The simple answer is "separation of concerns" and "flexibility".

I chose one between many possible configuration parsing libraries for scala (the choice in this regard is as rich almost as with json-parsing).
Keeping the trait separated allows to relatively restrain the amount of code directly dependent on the _pureconfig_ external library.
The goal is to keep the specifics of implementation grouped in as few possible sources as possible... the same reasoning applies to how `ConfigUtils` keep the code related to pureconfig separated from the rest of the codebase (by keeping it in a custom util file) and from other possible utilities (by enclosing that code in a specific sub-object)

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
ivanopagano: fine

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
ivanopagano: I chose to concentrate the logging configuration concern in the configuration trait, such that the main app needs only to take appropriate measures, specific to the app.

This also simplifies consistent handling of logging messages for config errors, without exposing details in the main objects.

If you have a strong opinion against such choice, I'm open to changes.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: The same reasoning exposed in the analogous comment applies here, please let me know if you see any strong argument against it

src/main/scala/tech/cryptonomic/conseil/config/LorreAppConfig.scala
ivanopagano: I've moved the dependency to a "send-enabled" subclass of the TezosNodeOperator.
This way:
- Lorre doesn't need the dependency
- Conseil will use Sodium only when we introduce the new operations
- the NodeOperator class is splitted in separate concerns (formally, read vs. write), which helps reduce bloating

src/main/scala/tech/cryptonomic/conseil/config/package.scala
ivanopagano: indeed, it's a special singleton instance whose members are visible to all the code under the same package.

e.g. every piece of code in the `tech.cryptonomic.conseil.config` package will have visibility of the `printConfigurationError` method.

You can import the members of a package object in scope by simply importing the package itself
i.e. `import tech.cryptonomic.conseil.config._`

src/main/scala/tech/cryptonomic/conseil/tezos/FeeOperations.scala
ivanopagano: sure, missed that

src/main/scala/tech/cryptonomic/conseil/config/ConseilConfig.scala
ivanopagano: I think it will be more sensible to split different networks when they add up...
Though I would advice to wait for the appropriate time, when the separation of the pieces will be clearer... at the moment it's not very clear-cut how and which of these configurations might be shared between platforms or not.

vishakh:
> > > @vishakh @ba205
> > > If you dig into the new way configuration-handling is introduced into this PR, you'll see that there's a `network` parameter passed along to all Tezos Node interface, only to be ignored in favor of the single argument passed as input to the run script (sbt run, or java ... command) for `Lorre`.
> > > Is it still useful to think of having a global network to run against from the launch and a different one to specify when calling the actual node operations?
> > > Should the node interface remove that extra parameter now?
> > 
> > 
> > Lorre should take a single platform and network, e.g. `tezos` and `mainnet`, and use that consistently throughout. Conseil should be stateless, i.e. every call should specify from the client the platform and network to be used. The `network` parameter should be passed respectively to the various classes.
> 
> Ok, I guess I figured that; Nevertheless I think that the most probable direction we'll take here will be to spawn different instances of TezosNodeInterface for each network... as the work done by @piotrkosecki is already hinting at.
> This will help us define before-hand how many networks are valid, and correctly configured to be contacted...
> Should we expect to possibly have different implementations in the future of the TezosRPCInterface, such that the `network` parameter could be actually used...?
> 
> I'm tempted to avoid confusion by removing it now, and eventually re-introduce it if we want to backtrack for some reason.
> Right now it only introduces an overlapping name in scope, making for a possible source of unexpected bugs...

I image we will have a `GenericNodeInterface` with multiple implementations in the future, e.g. for Tezos, Ethereum, etc. We might even have different packages for TezosMainnet vs. TezosZeronet, with the latter inheriting from the former. I think it's good to keep the `network` parameter in preparation for that.

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
vishakh: Sounds good.

PR #178, opened 2018-11-06
**PoC-No-Merge** Conseil-177: stream save blocks

### LINKED ISSUE #177: Lorre should not bulk-save all blocks in one shot
## Current State
Currently the process can be roughly summarized as

1. load all newly available blocks over http from tezos node
2. load all related operations
3. store everything in Conseil's db

## Issues
This has huge drawbacks when the number of blocks to read are more than a few hundreds.

The most critical scenario is that of rebuilding the whole chain for a new Conseil server, in this case the process seems to hang on several I/O operations (almost no CPU usage) for more than one hour, possibly (and this can only grow worse with time).

 - Until the whole process run is completed, no result is available on db.
 - If something breaks during the process, it might need to restart from scratch once again, even if most of the blocks were already loaded in memory
 - The memory consumption is certainly sub-optimal and might get unwieldy with time
 - CPU computational power is uselessly wasted during the whole process

## Proposals
Given those premises, a proposal would be to convert the process so that blocks are saved just as they come.

Given the current stream-based loading of blocks, the whole processing might make use of it, and deliver a staged-streaming-pipeline where blocks are processed as they are received on the stream, properly converted, filtered, enriched, stored. All of this using a back-pressure model by design.

## Additional Considerations
One point to raise is the integrity and consistency of the data. As the constraints between block-predecessors was removed from the db-schema itself, it might happen that we come up with "orphaned" blocks and similarly related inconsistencies.

My first suggestion would be to establish a consistency-checker process that starts immediately after the blocks are stored and marks possible integrity-violating record.

Details should be fleshed out during development, and a first choice is immediately apparent:
 1. mark stored blocks as "to-be-checked" and then mark them as valid as the checks complete
 2. mark stored blocks as "inconsistent" during the check process

In both cases we might generate a "report" after each checking round completes

ivanopagano:
**This is a proof-of-concept**
======================
## might never get into production

Converts Lorre's "bulk downloads" of nodes entities to a streaming one, including

 * blocks
 * accounts

Fixes #177 
This solution would avoid repeated work on failing process and reduce the memory footprint for Lorre runs.
There are additional considerations to take into account, though:
 1. accounts are still too slow to download, the sheer number of it, so a radically different solution is needed
 1. possibly failing streams introduce the risk of inconsistencies in the database, which needs some correction algorithm that validates/invalidates currently stored blocks as consistent and recovery processes that loads missing entities on each periodic sync with tezos
 1. based on the previous point, a local filtering process should be added (based on pre-cached keys?) to avoid duplicate work on the stream processing
 1. additional chances to increase parallelism might be enabled by the different approach and the sync cycle might be reduced to optimize inconsistency windows

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -3,10 +3,13 @@ package tech.cryptonomic.conseil
 
 import akka.actor.ActorSystem
 import akka.Done
+import akka.stream.scaladsl._
+import akka.stream._
 import com.typesafe.config.ConfigFactory
 import com.typesafe.scalalogging.LazyLogging
 import tech.cryptonomic.conseil.tezos.{FeeOperations, TezosNodeInterface, TezosNodeOperator, TezosDatabaseOperations => TezosDb}
 import tech.cryptonomic.conseil.util.DatabaseUtil
+import tech.cryptonomic.conseil.tezos.TezosTypes._
 
 import scala.concurrent.duration._
 import scala.annotation.tailrec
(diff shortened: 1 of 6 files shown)


vishakh:
Out of scope.

Issue #132, opened 2018-07-31
Project configuration for local execution should be less ad-hoc


ivanopagano:
The project could include additional setup that automates the needed steps to locally run the application, including but possibly not limited to

 - running and setting-up the underlying database
 - custom configuring the options to run the main classes in the project
 - custom configuring the options to run test coverage

The goal would be to reduce dependency from developer-specific tools that would need custom set-up when engaging the project.

PR #179, opened 2018-11-06
Conseil #132 - sbt tasks for running subprojects

piotrkosecki:
I've added 3 sbt tasks for easier running of the project. Also updated the readme showing how my improvement can be used. 

Improves #132

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Run locally using SBT:
 
 ` env SBT_OPTS="-Dconfig.file={path to custom config file}" sbt "runMain tech.cryptonomic.conseil.Conseil"`
 
-Or you can use the predefined tasks:
+Or you can use the predefined tasks which are custom-tailored with VM launch configuration that's best suited for the actual task:
 - for Conseil:
 `sbt -Dconfig.file="{path to custom config file}" runConseil`
 - for Lorre 
--- a/README.md
+++ b/README.md
@@ -37,6 +37,15 @@ Run locally using SBT:
 
 ` env SBT_OPTS="-Dconfig.file={path to custom config file}" sbt "runMain tech.cryptonomic.conseil.Conseil"`
 
+Or you can use the predefined tasks:
+- for Conseil:
+`sbt -Dconfig.file="{path to custom config file}" runConseil`
+- for Lorre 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -37,6 +37,15 @@ Run locally using SBT:
...
 
+Or you can use the predefined tasks:
+- for Conseil:
+`sbt -Dconfig.file="{path to custom config file}" runConseil`
+- for Lorre 
+`sbt -Dconfig.file="{path to custom config file}" "runLorre alphanet"`
+- for schema generation
+`sbt -Dconfig.file="{path to custom config file}" genSchema`
+
+
```
ivanopagano: I would explicitly state here that using the predefined tasks you get a custom-tailored vm launch configuration that's best suited for the actual task

build.sbt
```diff
@@ -38,4 +38,22 @@ excludeDependencies ++= Seq(
...
+
+fork in runConseil := true
+lazy val runConseil = taskKey[Unit]("A conseil run task.")
+javaOptions in runConseil ++= Seq("-Xms512M", "-Xmx4096M", "-Xss1M", "-XX:+CMSClassUnloadingEnabled")
+fullRunTask(runConseil, Runtime, "tech.cryptonomic.conseil.Conseil")
+
+fork in runLorre := true
+lazy val runLorre = inputKey[Unit]("A lorre run task.")
+javaOptions ++= Seq("-Xmx512M", "-Xss1M", "-XX:+CMSClassUnloadingEnabled")
+runLorre := Def.inputTaskDyn {
```
ivanopagano: Don't know in detail how this works, but from the [docs](https://www.scala-sbt.org/1.0/docs/Faq.html#How+can+I+create+a+custom+run+task%2C+in+addition+to+%3F) I would've thought that we only needed to use `fullRunInputTask` to get the cli argument.

Why the dynamic input part?

piotrkosecki: yeah, that would seem natural, but its sbt so just using `fullRunInputTask` does not want to read the config file 😄

piotrkosecki: dynamic part basically allows to evaluate args and put them into the task. In regular input task you could only hardcode things.

README.md
piotrkosecki: ok, quoted you in the docs 😄

PR #185, opened 2018-11-08
Conseil 180/181 expose list of available networks and entities in service route

piotrkosecki:
In this PR I'm adding 2 service endpoints:
1. `{host}:{port}/service/networks`
2. `{host}:{port}/service/networks/{network-name}/entities`

First one returns networks from config defined at `platforms` path(#180 ).
Second one returns entities and and counts their occurrences in the db (#181 ).

Issue #182, opened 2018-11-07
Expose list of available attributes via service call


anonymoussprocket:
Name it something along the lines of "ListAttributesForEntity", it will take a single parameter which would be a reference to an entity as described in #181.

The response should be an array of elements where the attributes are:
- Name: unique internal name that can be used to reference this object
- DisplayName: attribute name that may be visible to the user
- DataType: enum, hex, binary, date, datetime, string, int, largeint, decimal, boolean
- Cardinality: a count of district values, up to 1M.
- KeyType: enum: non-key, unique key
- Entity: entity reference

piotrkosecki:
Done in https://github.com/Cryptonomic/Conseil/pull/186

Issue #181, opened 2018-11-07
Expose list of available entities via service call


anonymoussprocket:
Name it something along the lines of "ListEntityForNetwork", it will take a single parameter which would be a reference to a network as described in #180

The response should be an array of elements where the attributes are:
- Name: unique internal name that can be used to reference this object
- DisplayName: entity name that may be visible to the user
- Count: a recent count of these elements
- Network: network reference

piotrkosecki:
Done in https://github.com/Cryptonomic/Conseil/pull/186

Issue #180, opened 2018-11-07
Expose list of available networks via service call


anonymoussprocket:
Name it something along the lines of "ListNetworks"

The response should be an array of elements where the attributes are:

- Name: unique internal name that can be used to reference this object
- DisplayName: network name that may be visible to the user
- Platform: enum, currently having a single value - Tezos
- Network: network-specific subtype, in the case of Tezos, one of mainnet, zeronet, alphanet

piotrkosecki:
Done in https://github.com/Cryptonomic/Conseil/pull/186

PR #186, opened 2018-11-13
Conseil 182 - expose list of available attributes via service call

piotrkosecki:
In this PR I'm adding 1 service endpoint in addition to already existing in the previous PR https://github.com/Cryptonomic/Conseil/pull/185:

{host}:{port}/service/networks
{host}:{port}/service/networks/{network-name}/entities
First one returns networks from config defined at platforms path(#180 ).
Second one returns entities and and counts their occurrences in the db (#181 ).

New one:
{host}:{port}/service/networks/{network-name}/entities/{entity-name}/attributes
and returns description of attributes from the DB - (#182 ).

Also made some refactor in things already done in the previous PR.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Conseil.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Conseil.scala
@@ -11,7 +11,7 @@ import com.typesafe.config.ConfigFactory
 import com.typesafe.scalalogging.LazyLogging
 import com.typesafe.sslconfig.akka.AkkaSSLConfig
 import tech.cryptonomic.conseil.directives.EnableCORSDirectives
-import tech.cryptonomic.conseil.routes.{Service, Tezos}
+import tech.cryptonomic.conseil.routes.{PlatformDiscovery, Tezos}
 import tech.cryptonomic.conseil.util.SecurityUtil
 
 import scala.concurrent.ExecutionContextExecutor
@@ -45,7 +45,7 @@ object Conseil extends App with LazyLogging with EnableCORSDirectives {
           }
         } ~ logRequest("Service route", Logging.DebugLevel) {
           pathPrefix("service") {
-            Service(conf)(system.dispatchers.lookup("akka.tezos-dispatcher")).route
+            PlatformDiscovery(conf)(system.dispatchers.lookup("akka.tezos-dispatcher")).route
           }
         }
       } ~ options {
--- a/src/main/scala/tech/cryptonomic/conseil/Conseil.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Conseil.scala
@@ -45,7 +45,7 @@ object Conseil extends App with LazyLogging with EnableCORSDirectives {
           }
         } ~ logRequest("Service route", Logging.DebugLevel) {
           pathPrefix("service") {
-            new Service(conf).route
+            Service(conf)(system.dispatchers.lookup("akka.tezos-dispatcher")).route
           }
         }
       } ~ options {
--- a/src/main/scala/tech/cryptonomic/conseil/Conseil.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Conseil.scala
@@ -11,7 +11,7 @@ import com.typesafe.config.ConfigFactory
 import com.typesafe.scalalogging.LazyLogging
 import com.typesafe.sslconfig.akka.AkkaSSLConfig
 import tech.cryptonomic.conseil.directives.EnableCORSDirectives
-import tech.cryptonomic.conseil.routes.Tezos
+import tech.cryptonomic.conseil.routes.{Service, Tezos}
 import tech.cryptonomic.conseil.util.SecurityUtil
 
 import scala.concurrent.ExecutionContextExecutor
@@ -42,6 +42,10 @@ object Conseil extends App with LazyLogging with EnableCORSDirectives {
         logRequest("Conseil", Logging.DebugLevel) {
           pathPrefix("tezos") {
             Tezos(system.dispatchers.lookup("akka.tezos-dispatcher")).route
+          }
+        } ~ logRequest("Service route", Logging.DebugLevel) {
+          pathPrefix("service") {
+            new Service(conf).route
           }
         }
       } ~ options {
--- a/src/main/scala/tech/cryptonomic/conseil/Conseil.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Conseil.scala
@@ -44,7 +44,7 @@ object Conseil extends App with LazyLogging with EnableCORSDirectives {
             Tezos(system.dispatchers.lookup("akka.tezos-dispatcher")).route
           }
         } ~ logRequest("Service route", Logging.DebugLevel) {
-          pathPrefix("service") {
+          pathPrefix("metadata") {
             PlatformDiscovery(conf)(system.dispatchers.lookup("akka.tezos-dispatcher")).route
           }
         }
--- a/src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
@@ -24,6 +24,8 @@ class PlatformDiscovery(config: Config)(implicit apiExecutionContext: ExecutionC
           pathPrefix("entities") {
             pathEnd {
               completeWithJson(PlatformDiscoveryOperations.getEntities(network))
+            }  ~ pathPrefix(Segment) { entity =>
+              completeWithJson(PlatformDiscoveryOperations.tableAttributes(entity))
             }
           }
         }
--- a/src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
@@ -6,7 +6,7 @@ import com.typesafe.config.Config
 import com.typesafe.scalalogging.LazyLogging
 import tech.cryptonomic.conseil.tezos.PlatformDiscoveryOperations
 import tech.cryptonomic.conseil.util.JsonUtil._
-import tech.cryptonomic.conseil.util.RoutesUtil
+import tech.cryptonomic.conseil.util.RouteHandling
 
 import scala.concurrent.ExecutionContext
 
@@ -20,7 +20,7 @@ object PlatformDiscovery {
   * @param config configuration object
   * @param apiExecutionContext is used to call the async operations exposed by the api service
   */
-class PlatformDiscovery(config: Config)(implicit apiExecutionContext: ExecutionContext) extends LazyLogging with RoutesUtil {
+class PlatformDiscovery(config: Config)(implicit apiExecutionContext: ExecutionContext) extends LazyLogging with RouteHandling {
   val route: Route =
     get {
       pathPrefix("networks") {
--- a/src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
@@ -4,26 +4,26 @@ import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.Route
 import com.typesafe.config.Config
 import com.typesafe.scalalogging.LazyLogging
-import tech.cryptonomic.conseil.tezos.ServiceOperations
+import tech.cryptonomic.conseil.tezos.PlatformDiscoveryOperations
 import tech.cryptonomic.conseil.util.JsonUtil._
 import tech.cryptonomic.conseil.util.RoutesUtil
 
 import scala.concurrent.ExecutionContext
 
-object Service {
-  def apply(config: Config)(implicit apiExecutionContext: ExecutionContext): Service = new Service(config)
+object PlatformDiscovery {
+  def apply(config: Config)(implicit apiExecutionContext: ExecutionContext): PlatformDiscovery = new PlatformDiscovery(config)
 }
 
-class Service(config: Config)(implicit apiExecutionContext: ExecutionContext) extends LazyLogging with RoutesUtil {
+class PlatformDiscovery(config: Config)(implicit apiExecutionContext: ExecutionContext) extends LazyLogging with RoutesUtil {
   val route: Route =
     get {
       pathPrefix("networks") {
         pathEnd {
-          complete(toJson(ServiceOperations.getNetworks(config)))
+          complete(toJson(PlatformDiscoveryOperations.getNetworks(config)))
         } ~ pathPrefix(Segment) { network =>
           pathPrefix("entities") {
             pathEnd {
-              completeWithJson(ServiceOperations.getEntities(network))
+              completeWithJson(PlatformDiscoveryOperations.getEntities(network))
             }
           }
         }
--- a/src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
@@ -24,8 +24,12 @@ class PlatformDiscovery(config: Config)(implicit apiExecutionContext: ExecutionC
           pathPrefix("entities") {
             pathEnd {
               completeWithJson(PlatformDiscoveryOperations.getEntities(network))
-            }  ~ pathPrefix(Segment) { entity =>
-              completeWithJson(PlatformDiscoveryOperations.tableAttributes(entity))
+            } ~ pathPrefix(Segment) { entity =>
+              pathPrefix("attributes") {
+                pathEnd {
+                  completeWithJson(PlatformDiscoveryOperations.tableAttributes(entity))
+                }
+              }
             }
           }
         }
--- a/src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
@@ -10,10 +10,16 @@ import tech.cryptonomic.conseil.util.RoutesUtil
 
 import scala.concurrent.ExecutionContext
 
+/** Companion object providing apply implementation */
 object PlatformDiscovery {
   def apply(config: Config)(implicit apiExecutionContext: ExecutionContext): PlatformDiscovery = new PlatformDiscovery(config)
 }
 
+/**
+  * Platform discovery routes.
+  * @param config configuration object
+  * @param apiExecutionContext is used to call the async operations exposed by the api service
+  */
 class PlatformDiscovery(config: Config)(implicit apiExecutionContext: ExecutionContext) extends LazyLogging with RoutesUtil {
   val route: Route =
     get {
@@ -27,7 +33,7 @@ class PlatformDiscovery(config: Config)(implicit apiExecutionContext: ExecutionC
             } ~ pathPrefix(Segment) { entity =>
               pathPrefix("attributes") {
                 pathEnd {
-                  completeWithJson(PlatformDiscoveryOperations.tableAttributes(entity))
+                  completeWithJson(PlatformDiscoveryOperations.getTableAttributes(entity))
                 }
               }
             }
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
@@ -6,7 +6,7 @@ import scala.concurrent.{ExecutionContext, Future}
 import scala.collection.JavaConverters._
 
 
-object ServiceOperations {
+object PlatformDiscoveryOperations {
 
   final case class Counts(blocks: Int, accounts: Int, operationGroups: Int, operations: Int, fees: Int)
 
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
@@ -18,7 +18,7 @@ object PlatformDiscoveryOperations {
   lazy val dbHandle: PostgresProfile.backend.Database = DatabaseUtil.db
 
   private val tables = List(Tables.Blocks, Tables.Accounts, Tables.OperationGroups, Tables.Operations, Tables.Fees)
-  private val tablesMap = tables.map(table => table.baseTableRow.tableName -> table).toMap
+  private val tablesMap = tables.map(table => table.baseTableRow.tableName -> table)
 
   def getNetworks(config: Config): List[Network] = {
     config.getObject("platforms").asScala.flatMap {
@@ -32,7 +32,7 @@ object PlatformDiscoveryOperations {
 
   def getEntities(network: String)(implicit ec: ExecutionContext): Future[List[Entity]] = {
     ApiOperations.countAll.map { counts =>
-      tablesMap.keys.toList.flatMap { tableName =>
+      tablesMap.map(_._1).flatMap { tableName =>
         counts.get(tableName).map { tableCount =>
           Entity(
             name = tableName,
@@ -52,7 +52,7 @@ object PlatformDiscoveryOperations {
   def tableAttributes(tableName: String)(implicit ec: ExecutionContext): Future[List[Attributes]] = {
     val res: DBIO[List[Attributes]] =
       DBIO.sequence(
-        tablesMap.toList.collect {
+        tablesMap.collect {
           case (name, table) if name == tableName =>
             val overallCount = TezosDb.countRows(table)
             table.baseTableRow.create_*.map { col =>
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
@@ -78,16 +78,16 @@ object PlatformDiscoveryOperations {
 
 
   private def mapType(tpe: String): DataType = {
+    val optionRegex = "Option\\[([A-Za-z0-9']+)\\]".r
     tpe match {
       case "java.sql.Timestamp'" => DataType.DateTime
       case "String'" => DataType.String
       case "Int'" => DataType.Int
       case "Long'" => DataType.LargeInt
       case "Float'" | "Double'" | "scala.math.BigDecimal'" => DataType.Decimal
-      case "Boolean" => DataType.Boolean
-      case x if x.startsWith("Option[") => mapType(x.drop(7).dropRight(1))
+      case "Boolean'" => DataType.Boolean
+      case optionRegex(t) => mapType(t)
       case _ => DataType.String
     }
   }
-
 }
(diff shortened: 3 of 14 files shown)


src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
```diff
@@ -0,0 +1,38 @@
...
+object PlatformDiscovery {
+  def apply(config: Config)(implicit apiExecutionContext: ExecutionContext): PlatformDiscovery = new PlatformDiscovery(config)
+}
+
+class PlatformDiscovery(config: Config)(implicit apiExecutionContext: ExecutionContext) extends LazyLogging with RoutesUtil {
+  val route: Route =
+    get {
+      pathPrefix("networks") {
+        pathEnd {
+          complete(toJson(PlatformDiscoveryOperations.getNetworks(config)))
```
ivanopagano: is this not `completeWithJson` too?

piotrkosecki: no it's not, `completeWithJson` works on asynchronous types -> `completeWithJson[T](futureValue: Future[T])`, and `PlatformDiscoveryOperations.getNetworks(config)` is synchronous

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -0,0 +1,130 @@
...
+                displayName = makeDisplayName(col.name),
+                dataType = mapType(col.tpe.toString),
+                cardinality = distinctCount,
+                keyType = if(distinctCount == overallCount) KeyType.UniqueKey else KeyType.NonKey,
+                entity = table.baseTableRow.tableName
+              )
+          }
+        }
+    }
+    futOpt(res).map(_.toList.flatten)
```
ivanopagano: I can suggest splitting the DBIO creation from execution with something of the form
```scala
val dbAction: Option[DBIO[...]] = tablesMap.collect {
  ...
  DBIO.sequence(...)
}
Future.traverse(dbActions.toList)(dbHandle.run)
  .map(_.toList.flatten) //<- if needed still?
```

This would also remove the need to define `futOpt`

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+
+  private def futOpt[A](x: Option[Future[A]])(implicit ec: ExecutionContext): Future[Option[A]] =
+    x match {
+      case Some(f) => f.map(Some(_))
+      case None => Future.successful(None)
+    }
+
+  class DataTypeRef extends TypeReference[DataType.type]
```
ivanopagano: Should this be a real class or a type alias (true question)?

ivanopagano: What about putting these data types in a separate source?

Similar to what was done with the TezosTypes object

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -0,0 +1,130 @@
...
+  class KeyTypeRef extends TypeReference[KeyType.type]
+
+  final case class Network(name: String, displayName: String, platform: String, network: String)
+
+  final case class Entity(name: String, displayName: String, count: Int, network: String)
+
+  final case class Attributes(
+    name: String,
+    displayName: String,
+    @JsonScalaEnumeration(classOf[DataTypeRef]) dataType: DataType,
```
ivanopagano: cool, how does this work?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -250,4 +250,12 @@ object TezosDatabaseOperations extends LazyLogging {
...
     Tables.Blocks.exists.result
 
+  /** count number of the elements in the table */
+  def count(table: TableQuery[_]): DBIO[Int] =
+    table.length.result
+
+  // Slick did not let me to this any other way
+  /** count number of the distinct elements in the table */
+  def countDistinct(table: String, column: String)(implicit ec: ExecutionContext): DBIO[Int] =
+    sql"""SELECT COUNT(DISTINCT #$column) FROM #$table""".as[Int].map(_.head)
```
ivanopagano: I think it does, but please make sure that the `sql` interpolator does safe-encoding of the passed in args

I don't want someone to ask for table "table; DELETE * from DB" 😁

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -250,4 +250,12 @@ object TezosDatabaseOperations extends LazyLogging {
   def doBlocksExist(): DBIO[Boolean] =
     Tables.Blocks.exists.result
 
+  /** count number of the elements in the table */
+  def count(table: TableQuery[_]): DBIO[Int] =
```
ivanopagano: just thinking, `countRows` might be more explicit?

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -0,0 +1,130 @@
...
+
+  private def mapType(tpe: String): DataType = {
+    tpe match {
+      case "java.sql.Timestamp'" => DataType.DateTime
+      case "String'" => DataType.String
+      case "Int'" => DataType.Int
+      case "Long'" => DataType.LargeInt
+      case "Float'" | "Double'" | "scala.math.BigDecimal'" => DataType.Decimal
+      case "Boolean" => DataType.Boolean
+      case x if x.startsWith("Option[") => mapType(x.drop(7).dropRight(1))
```
ivanopagano: ouch... how rude 😏

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -0,0 +1,130 @@
...
+            table.baseTableRow.create_*.map { col =>
+              for {
+                overallCnt <- overallCount
+                distinctCnt <-  TezosDb.countDistinct(table.baseTableRow.tableName, col.name)
+              } yield  (col, distinctCnt, overallCnt)
+            }
+          )
+        }.map { distinctCountList =>
+          distinctCountList.map {
+            case (col, distinctCount, overallCount) =>
```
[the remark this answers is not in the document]
piotrkosecki: ok, refactored

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  private def makeDisplayName(name: String): String = {
+    name.capitalize.replace("_", " ")
+  }
+
+  def tableAttributes(tableName: String)(implicit ec: ExecutionContext): Future[List[Attributes]] = {
+    val tablesMap = tables.map(table => table.baseTableRow.tableName -> table)
+
+    val res = tablesMap.find(_._1 == tableName).map {
```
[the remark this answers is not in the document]
piotrkosecki: ok, refactored

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -0,0 +1,130 @@
...
+      }
+    }
+  }
+
+  private def makeDisplayName(name: String): String = {
+    name.capitalize.replace("_", " ")
+  }
+
+  def tableAttributes(tableName: String)(implicit ec: ExecutionContext): Future[List[Attributes]] = {
+    val tablesMap = tables.map(table => table.baseTableRow.tableName -> table)
```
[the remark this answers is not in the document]
piotrkosecki: ok, don't mind moving it

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
piotrkosecki: made it a little different but with similar idea

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
piotrkosecki: https://github.com/FasterXML/jackson-module-scala/wiki/Enumerations
That is a thing I need to serialize enums with jackson.

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
piotrkosecki: My bad, I haven't noticed where were those object stored. Created a new one and moved all the data types there.

piotrkosecki: this is the way it was described in the manual so I could make enumerations work:
https://github.com/FasterXML/jackson-module-scala/wiki/Enumerations

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: unfortunately it does not(tested locally😄), but int his case this method is used with the only predefined earlier values by slick code gen `TezosDb.countDistinct(table:TableQuery[_].baseTableRow.tableName, col:FieldSymbol.name)`
The best I can do is to put the big warning comment in here, as slick didn't want me to handle this other way than creating custom sql.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: sure, why not 😄

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
piotrkosecki: this part is tricky, as in the ticket defined there was no definition of nullable or optional field I've decided that I need to extract it somehow. Probably there is some regex which could help me but I don't know it yet and still looking for it 😄

piotrkosecki: ok, found a regex and replaced this case 😄 also is worth to mention that I've only matched the cases we currently have in the DB(or are really easy to implement) so there is a possibility that this method will need to be extended in the future

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
ivanopagano: ok, so it's jackson stuff, thanks

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: This, or you may find what's the best way to sanitize sql-injected values (I suspect it might suffice to wrap in quotes?)

I thought too that _now_ it gets called only by safe code, but in the future who knows? We might  need to reuse that code from other callers and forget that it's unsafe (so at least the documented warning might turn out necessary)

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -0,0 +1,93 @@
...
+        strippedConf.atKey(platform).getObject(platform).asScala.map {
+          case (network, _) =>
+            Network(network, network.capitalize, platform, network)
+        }.toList
+    }.toList
+  }
+
+  def getEntities(network: String)(implicit ec: ExecutionContext): Future[List[Entity]] = {
+    ApiOperations.countAll.map { counts =>
+      tables.map(_.baseTableRow.tableName).flatMap { tableName =>
```
ivanopagano: This could now be written as `tablesMap.keys.flatMap` or equivalent `tablesMap.keySet.flatMap`

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
ivanopagano: awesome

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
piotrkosecki: Unfortunately it is not a real `Map` and I cannot create one -> I've got an error from compiler:
```
PlatformDiscoveryOperations.scala:21:86: Cannot prove that (String, slick.lifted.TableQuery[_1])( forSome { type _1 >: tech.cryptonomic.conseil.tezos.Tables.Fees with tech.cryptonomic.conseil.tezos.Tables.Operations with tech.cryptonomic.conseil.tezos.Tables.OperationGroups with tech.cryptonomic.conseil.tezos.Tables.Accounts with tech.cryptonomic.conseil.tezos.Tables.Blocks <: tech.cryptonomic.conseil.tezos.Tables.profile.Table[_ >: tech.cryptonomic.conseil.tezos.Tables.FeesRow with tech.cryptonomic.conseil.tezos.Tables.OperationsRow with tech.cryptonomic.conseil.tezos.Tables.OperationGroupsRow with tech.cryptonomic.conseil.tezos.Tables.AccountsRow with tech.cryptonomic.conseil.tezos.Tables.BlocksRow <: Product with Serializable] }) <:< (T, U).
[error]   private val tablesMap = tables.map(table => table.baseTableRow.tableName -> table).toMap
```
I changed it but not the way you suggested

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: I think this is the best I could do, as wrapping with quotes does not work and added comment with capital letters.

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
ivanopagano: Fine anyway!

Just FYI, I guess you simply could have defined `tablesMap` ending with `.toMap`? To have a real scala.Map

piotrkosecki: check out the error I've pasted above,`.toMap` doesn't work, I guess because I have a bit more extended `TableQuery` - apparently if I had typed it explicitly as `TableQuery` it would not infer any of methods I use here.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ba205: Let x be some defined table, like Table.Accounts, for example. Wouldn't x.distinct.length do the same thing? Maybe there's confusion because slick's countDistinct function is deprecated?

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -0,0 +1,93 @@
...
+      name = col.name,
+      displayName = makeDisplayName(col.name),
+      dataType = mapType(col.tpe.toString),
+      cardinality = distinctCount,
+      keyType = if (distinctCount == overallCount) KeyType.UniqueKey else KeyType.NonKey,
+      entity = tableName
+    )
+
+
+  private def mapType(tpe: String): DataType = {
```
ba205: Since this is DataType's constructor function, maybe it makes more sense to place this in PlatformDiscoveryTypes.scala? Thoughts?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: @ba205 there is function combination `distinctOn(<column>).length` in slick but I couldn't pass the column as an argument to this method so I had to do the plain sql - writing generic code in slick is tricky

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
piotrkosecki: I moved it.
I'm not proud of this function, I need to figure out how to get better type and map it to my enum, as matching on strings is not really good design. Probably some macro or shapeless would help, but I'm not sure yet.

piotrkosecki: So from what I've investigated this solution is not perfect but I couldn't come up with anything better at this time. The closes I've got was something like this:
```
  def mapType(tpe: Type): DataType = {
    val clazz = tpe.classTag.runtimeClass

    val Ts = classOf[Timestamp]
    val Str = classOf[String]
    val Integer = classOf[Int]
    val Lng = classOf[Long]
    val Bool = classOf[Boolean]
    val Opt = classOf[Option[_]]
    val Flt = classOf[Float]
    val Dble = classOf[Double]
    val Bdecimal = classOf[BigDecimal]

    clazz match {
      case Ts => DataType.DateTime
      case Str => DataType.String
      case Integer => DataType.Int
      case Lng => DataType.LargeInt
      case Flt | Dble | Bdecimal => DataType.Decimal
      case Bool => DataType.Boolean
      case Opt => ???  // <- here is the problem
      case _ => DataType.String
    }
  }
```
But I cannot match options with it because this `classOf(Option[String]) == classOf(Option[Int])` is `true`. Also I cannot extract type from `Option[T]`. So possibly I could do the second round of matching but now on strings, but it would repeat what I already have.

ba205: https://docs.scala-lang.org/overviews/reflection/typetags-manifests.html
ClassTag has the same problem, but maybe TypeTag doesn't?

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -0,0 +1,77 @@
...
+        }.toList
+      }
+    }
+  }
+
+  private def makeDisplayName(name: String): String = {
+    name.capitalize.replace("_", " ")
+  }
+
+  def tableAttributes(tableName: String)(implicit ec: ExecutionContext): Future[List[Attributes]] = {
```
vishakh: Please validate the function parameter and either throw an exception or return an error type.

vishakh:
General suggestion: Please document all new classes and public functions and types.

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
ivanopagano: Let's try to stick to avoiding exceptions and wrap into appropriate error types, even using cats ones, if needed.

We're trying to make all functions generally return a value, unless a critical failure happens

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -43,6 +43,10 @@ object Conseil extends App with LazyLogging with EnableCORSDirectives {
           pathPrefix("tezos") {
             Tezos(system.dispatchers.lookup("akka.tezos-dispatcher")).route
           }
+        } ~ logRequest("Service route", Logging.DebugLevel) {
+          pathPrefix("service") {
```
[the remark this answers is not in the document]
piotrkosecki: I renamed it to `metadata` as changing it back won't be a big deal.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -250,4 +250,16 @@ object TezosDatabaseOperations extends LazyLogging {
   def doBlocksExist(): DBIO[Boolean] =
     Tables.Blocks.exists.result
 
+  /** count number of the elements in the table */
+  def countRows(table: TableQuery[_]): DBIO[Int] =
+    table.length.result
+
+  // Slick did not let me to this any other way
```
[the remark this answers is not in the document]
piotrkosecki: changed to your version

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -250,4 +250,16 @@ object TezosDatabaseOperations extends LazyLogging {
...
+  /** count number of the distinct elements in the table
+    * WARNING!
+    * THIS METHOD IS VULNERABLE TO SQL INJECTION.
+    * USE IT WISELY.
+    */
+  def countDistinct(table: String, column: String)(implicit ec: ExecutionContext): DBIO[Int] =
```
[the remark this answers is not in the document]
piotrkosecki: ok, changed

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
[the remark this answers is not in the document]
piotrkosecki: I refactored this function but not everything can be smaller.

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -0,0 +1,77 @@
...
+
+
+object PlatformDiscoveryOperations {
+
+  lazy val dbHandle: PostgresProfile.backend.Database = DatabaseUtil.db
+
+  private val tables = List(Tables.Blocks, Tables.Accounts, Tables.OperationGroups, Tables.Operations, Tables.Fees)
+  private val tablesMap = tables.map(table => table.baseTableRow.tableName -> table)
+
+  def getNetworks(config: Config): List[Network] = {
```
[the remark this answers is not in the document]
piotrkosecki: I refactored this function but not everything can be smaller.

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -0,0 +1,77 @@
...
+    config.getObject("platforms").asScala.flatMap {
+      case (platform, strippedConf) =>
+        strippedConf.atKey(platform).getObject(platform).asScala.map {
+          case (network, _) =>
+            Network(network, network.capitalize, platform, network)
+        }.toList
+    }.toList
+  }
+
+  def getEntities(network: String)(implicit ec: ExecutionContext): Future[List[Entity]] = {
```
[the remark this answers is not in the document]
piotrkosecki: I refactored this function but not everything can be divided to smaller functions.

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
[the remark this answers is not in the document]
piotrkosecki: ok, changed

ivanopagano: What works best for you, I would say a failed `Future` as a first answer... but it might depend on where the parameter is passed-in and where the return value is handled

src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
```diff
@@ -0,0 +1,38 @@
...
+import tech.cryptonomic.conseil.util.JsonUtil._
+import tech.cryptonomic.conseil.util.RoutesUtil
+
+import scala.concurrent.ExecutionContext
+
+object PlatformDiscovery {
+  def apply(config: Config)(implicit apiExecutionContext: ExecutionContext): PlatformDiscovery = new PlatformDiscovery(config)
+}
+
+class PlatformDiscovery(config: Config)(implicit apiExecutionContext: ExecutionContext) extends LazyLogging with RoutesUtil {
```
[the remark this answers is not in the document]
piotrkosecki: Ok, updated

src/main/scala/tech/cryptonomic/conseil/util/RoutesUtil.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import akka.http.scaladsl.marshalling.{PredefinedToEntityMarshallers, ToEntityMarshaller, ToResponseMarshallable, ToResponseMarshaller}
+import akka.http.scaladsl.model.{MediaTypes, StatusCodes}
+import akka.http.scaladsl.server.Directives.complete
+import akka.http.scaladsl.server.StandardRoute
+import tech.cryptonomic.conseil.util.JsonUtil.{JsonString, toJson}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait RoutesUtil {
```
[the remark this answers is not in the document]
piotrkosecki: Do you have something in mind? I thought about `RoutesUtil` because there are literally utility functions and implicits used in routes

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -0,0 +1,77 @@
...
+import slick.jdbc.PostgresProfile.api._
+import tech.cryptonomic.conseil.tezos.PlatformDiscoveryTypes._
+import tech.cryptonomic.conseil.tezos.{TezosDatabaseOperations => TezosDb}
+import tech.cryptonomic.conseil.util.DatabaseUtil
+
+import scala.collection.JavaConverters._
+import scala.concurrent.{ExecutionContext, Future}
+
+
+object PlatformDiscoveryOperations {
```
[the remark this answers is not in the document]
piotrkosecki: Wrote the tests

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
piotrkosecki: I see usually there are returned just `Future[]` in other places so I thought `Future[]` will be fine here

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
piotrkosecki: Unfortunately I couldn't make it work, there is possibility of doing it slightly better with `slick.meta.MTable`, but it would require few additional calls to the DB, which I'm not ok with - it has already a lot of queries per request, which were unavoidable. Depending on how the rest of the tickets will change current concept I may consider caching some values.

src/main/scala/tech/cryptonomic/conseil/util/RoutesUtil.scala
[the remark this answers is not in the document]
vishakh: How about `RouteHandling` or `RouteProcessing`?

Issue #183, opened 2018-11-07
Query for distinct values given an attribute


anonymoussprocket:
Taking an attribute reference (described in #182) return a list of distinct values.

Error conditions:
- query for high-cardinality field
- query for date, or datetime field
- query for any cardinality numeric field

Query for boolean fields should return a canned list: ["True", "False"]

piotrkosecki:
Done in https://github.com/Cryptonomic/Conseil/pull/191

PR #191, opened 2018-11-23
Conseil #183/#184  query for distinct values given an attribute, and attribute filtering

piotrkosecki:
New set of features for metadata endpoint:
Endpoint for listing distinct values of given attributes - doesn't work for numeric and datetime fields as requested in the ticket - returns list of strings:
`{host}:{port}/metadata/networks/{network-name}/entities/{entity-name}/attributes/{attribute-name}`

Endpoint for listing distinct values of given attributes with filter:
`{host}:{port}/metadata/networks/{network-name}/entities/{entity-name}/attributes/{attribute-name}/filter/{filter-string}`

For security reasons `{filter-string}` can only contain letters and digits. We can extend it as we agree on some other safe solution.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
@@ -83,7 +83,7 @@ object PlatformDiscoveryOperations {
     Attributes(
       name = col.name,
       displayName = makeDisplayName(col.name),
-      dataType = mapType(col.tpe.toString),
+      dataType = mapType(col.tpe),
       cardinality = distinctCount,
       keyType = if (distinctCount == overallCount) KeyType.UniqueKey else KeyType.NonKey,
       entity = tableName
@@ -133,7 +133,7 @@ object PlatformDiscoveryOperations {
     (implicit ec: ExecutionContext): DBIO[List[String]] = {
     for {
       distinctCount <- TezosDb.countDistinct(tableName, column.name)
-      if canQueryType(mapType(column.tpe.toString)) && !isHighCardinality(distinctCount)
+      if canQueryType(mapType(column.tpe)) && !isHighCardinality(distinctCount)
       distinctSelect <- withFilter match {
         case Some(filter) =>
           TezosDatabaseOperations.selectDistinctLike(tableName, column.name, sanitizeForSql(filter))
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
@@ -1,7 +1,7 @@
 package tech.cryptonomic.conseil.tezos
 
 import com.typesafe.config.Config
-import slick.ast.FieldSymbol
+import slick.ast.{FieldSymbol, Type}
 import slick.jdbc.PostgresProfile
 import slick.jdbc.PostgresProfile.api._
 import tech.cryptonomic.conseil.tezos.PlatformDiscoveryTypes.DataType.DataType
@@ -66,6 +66,27 @@ object PlatformDiscoveryOperations {
     dbHandle.run(res)
   }
 
+  def listAttributeValues(entity: String, attribute: String)(implicit ec: ExecutionContext): Future[Option[List[String]]] = {
+    val res = tablesMap.collectFirst {
+      case (name, table) if name == entity =>
+        table.baseTableRow.create_*.collectFirst {
+          case col if col.name == attribute =>
+            for {
+              dc <- TezosDb.countDistinct(table.baseTableRow.tableName, col.name)
+              if canQueryType(mapType(col.tpe.toString)) && !isHighCardinality(dc)
+              sd <- TezosDatabaseOperations.selectDistinct(entity, attribute)
+            } yield sd
+        }
+    }.flatten
+    futOpt(res.map(dbHandle.run))
+  }
+
+  private def futOpt[A](x: Option[Future[A]])(implicit ec: ExecutionContext): Future[Option[A]] =
+    x match {
+      case Some(f) => f.map(Some(_))
+      case None => Future.successful(None)
+    }
+
   private def makeAttributes(col: FieldSymbol, distinctCount: Int, overallCount: Int, tableName: String): Attributes =
     Attributes(
       name = col.name,
@@ -89,5 +110,21 @@ object PlatformDiscoveryOperations {
       case optionRegex(t) => mapType(t)
       case _ => DataType.String
     }
+  }
+
+  private def canQueryType(dt: DataType): Boolean = {
+    // values described in the ticket #183
+    val cantQuery = List(DataType.Date, DataType.DateTime, DataType.Int, DataType.LargeInt, DataType.Decimal)
+    val res = !cantQuery.contains(dt)
+    println(res)
+
+    res
+  }
+
+  private def isHighCardinality(distinctCount: Int): Boolean = {
+    // arbitrary value which will be defined in the config
+    val maxCount = 1000
+    println(distinctCount + " " + maxCount)
+    distinctCount > maxCount
   }
 }
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
@@ -57,24 +57,26 @@ object PlatformDiscoveryOperations {
     * @return list of attributes as a Future
     */
   def getTableAttributes(tableName: String)(implicit ec: ExecutionContext): Future[List[Attributes]] = {
-    ApiOperations.runQuerySequence(makeAttributesList(tableName))
+    ApiOperations.runQuery(makeAttributesList(tableName))
   }
 
   /** Makes list of DB actions to be executed for extracting attributes
     *
     * @param  tableName name of the table from which we extract attributes
     * @return list of DBIO queries for attributes
-    **/
-  def makeAttributesList(tableName: String)(implicit ec: ExecutionContext): List[DBIO[Attributes]] = {
-    for {
-      (name, table) <- tablesMap
-      if name == tableName
-      col <- table.baseTableRow.create_*
-    } yield {
+    * */
+  def makeAttributesList(tableName: String)(implicit ec: ExecutionContext): DBIO[List[Attributes]] = {
+    DBIO.sequence {
       for {
-        overallCnt <- TezosDb.countRows(table)
-        distinctCnt <- TezosDb.countDistinct(table.baseTableRow.tableName, col.name)
-      } yield makeAttributes(col, distinctCnt, overallCnt, tableName)
+        (name, table) <- tablesMap
+        if name == tableName
+        col <- table.baseTableRow.create_*
+      } yield {
+        for {
+          overallCnt <- TezosDb.countRows(table)
+          distinctCnt <- TezosDb.countDistinct(table.baseTableRow.tableName, col.name)
+        } yield makeAttributes(col, distinctCnt, overallCnt, tableName)
+      }
     }
   }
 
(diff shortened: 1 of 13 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -88,4 +93,72 @@ object PlatformDiscoveryOperations {
...
+    ApiOperations.runQuerySequence(res).map(_.flatten)
+  }
+
+  /** Makes list of DBIO actions to get possible string values of the attributes
+    * @param  tableName name of the table from which we extract attributes
+    * @param  attribute name of the attribute
+    * @param  withFilter optional parameter which can filter attributes
+    * @return list of DBIO actions to get matching attributes
+    **/
+  def verifyAttributesAndGetQueries(tableName: String, attribute: String, withFilter: Option[String])
```
ivanopagano: Is it feasible here to return a `DBIO[List[String]]` by `DBIO.sequence`-ing the `for` result and then `.map(_.flatten)` the ensuing `DBIO[List[List[String]]`?

For me it would make more sense to present a simpler type to the caller, instead of forcing him to flatten himself... unless the single operations in the list are to be used individually somewhere else...

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -88,4 +93,72 @@ object PlatformDiscoveryOperations {
...
+
+  /** Checks the data types if cannot be queried by */
+  private def canQueryType(dt: DataType): Boolean = {
+    // values described in the ticket #183
+    val cantQuery = List(DataType.Date, DataType.DateTime, DataType.Int, DataType.LargeInt, DataType.Decimal)
+    !cantQuery.contains(dt)
+  }
+
+  /** Checks if cardinality of the column is to high so it should not be queried */
+  private def isHighCardinality(distinctCount: Int): Boolean = {
```
ivanopagano: What about reversing the check and name this as an acceptance function (e.g. `isAcceptableColumnCardinality`, `isLimitedCardinality`, `isLowCardinality`...)?

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -88,4 +93,72 @@ object PlatformDiscoveryOperations {
...
+        case None =>
+          TezosDatabaseOperations.selectDistinct(tableName, column.name)
+      }
+    } yield distinctSelect
+  }
+
+  /** Checks the data types if cannot be queried by */
+  private def canQueryType(dt: DataType): Boolean = {
+    // values described in the ticket #183
+    val cantQuery = List(DataType.Date, DataType.DateTime, DataType.Int, DataType.LargeInt, DataType.Decimal)
```
ivanopagano: `Set` is usually a better choice for inclusion-checks. It also allows to apply it as `!cantQuery(dt)`

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
```diff
@@ -88,4 +93,72 @@ object PlatformDiscoveryOperations {
...
+    !cantQuery.contains(dt)
+  }
+
+  /** Checks if cardinality of the column is to high so it should not be queried */
```
ivanopagano: typo: "_to_ high"

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryTypes.scala
```diff
@@ -45,20 +46,20 @@ object PlatformDiscoveryTypes {
...
   /**
-    * Function maps slick column type as a string to DataType
-    * @param tpe type as a string from slick TableQuery
+    * Function maps slick jdbc column type to DataType
+    * @param tpe type from slick TableQuery
     * @return one of DataType enums
     */
-  def mapType(tpe: String): DataType = {
-    val optionRegex = "Option\\[([A-Za-z0-9']+)\\]".r
+  def mapType(tpe: Type): DataType = {
```
ivanopagano: This is a great improvement

Will it create problems if we need to use any other db than postgres? Say, H2 for testing?

src/main/scala/tech/cryptonomic/conseil/util/RouteHandling.scala
```diff
@@ -34,6 +34,13 @@ trait RouteHandling {
       case None => StatusCodes.NotFound
     }
 
+  protected def handleNoneAsTooLarge[T, R: ToResponseMarshaller](operation: => Future[Option[T]], converter: T => R = toJson[T] _)
```
ivanopagano: I can't seem to find if this is used somewhere

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
```diff
@@ -163,4 +171,83 @@ class PlatformDiscoveryOperationsTest
     }
   }
 
+  "listAttributeValues" should {
+
+    "return list of values of kind attribute of Fees without filter" in {
+      val avgFee = AverageFees(1, 3, 5, Timestamp.valueOf(LocalDateTime.of(2018, 11, 22, 12, 30)), "example1")
+
+      Await.result(dbHandler.run(TezosDatabaseOperations.writeFees(List(avgFee))), Duration.Inf)
```
ivanopagano: Waiting an infinite amount of time doesn't seem useful, and may create issues on a CI environment.

Why not defining a reasonably low expectation instead, so that a problematic test will fail faster?

Side note: if you prefer there's also an extension method `future.isReadyWithin(Span)`that returns a Boolean, without using `Await.ready`

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
```diff
@@ -163,4 +171,83 @@ class PlatformDiscoveryOperationsTest
...
+      dbHandler.run(
+        DBIO.sequence(
+          PlatformDiscoveryOperations.verifyAttributesAndGetQueries("fees", "kind", None)
+        )
+      ).futureValue.flatten shouldBe List("example1")
+    }
+
+    "return throw an exeption when asked for medium attribute of Fees without filter - numeric attributes should not be displayed" in {
```
ivanopagano: typo: "return throw an exception when..." -> "throw an exception when..."

personally I'd say that it "returns a failed future when..."

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
```diff
@@ -163,4 +171,83 @@ class PlatformDiscoveryOperationsTest
...
+      Await.result(dbHandler.run(TezosDatabaseOperations.writeFees(List(avgFee))), Duration.Inf)
+
+      val thrown = intercept[Exception] {
+        dbHandler.run(
+          DBIO.sequence(
+            PlatformDiscoveryOperations.verifyAttributesAndGetQueries("fees", "medium", None)
+          )
+        ).futureValue
+      }
+      thrown.getMessage shouldBe "The future returned an exception of type: java.util.NoSuchElementException, with message: Action.withFilter failed."
```
ivanopagano: I guess there should be a way to extract a failed future's Exception type and check that without depending on a specific string message

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
```diff
@@ -163,4 +171,83 @@ class PlatformDiscoveryOperationsTest
...
+      dbHandler.run(
+        DBIO.sequence(
+          PlatformDiscoveryOperations.verifyAttributesAndGetQueries("fees", "kind", maliciousFilter)
+        )
+      ).futureValue.flatten shouldBe List.empty
+
+      dbHandler.run(Tables.Fees.length.result).futureValue shouldBe 1
+
+    }
+    "check if filtering works correctly" in {
```
ivanopagano: This might be better phrased, the subject is "listAttributeValues"

The test case could read like, e.g. "listAttributeValues" should "correctly apply the filter" in ...

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
```diff
@@ -163,4 +171,83 @@ class PlatformDiscoveryOperationsTest
...
+      val avgFees = List(
+        AverageFees(1, 3, 5, Timestamp.valueOf(LocalDateTime.of(2018, 11, 22, 12, 30)), "example1"),
+        AverageFees(2, 4, 6, Timestamp.valueOf(LocalDateTime.of(2018, 11, 22, 12, 31)), "example2")
+      )
+      Await.result(dbHandler.run(TezosDatabaseOperations.writeFees(avgFees)), Duration.Inf)
```
ivanopagano: Waiting infinite time, refer to previous comment

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
```diff
@@ -163,4 +171,83 @@ class PlatformDiscoveryOperationsTest
...
+
+      dbHandler.run(
+        DBIO.sequence(
+          PlatformDiscoveryOperations.verifyAttributesAndGetQueries("fees", "kind", None)
+        )
+      ).futureValue.flatten.size shouldBe 2
```
ivanopagano: Why not test the exact values?

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
```diff
@@ -163,4 +171,83 @@ class PlatformDiscoveryOperationsTest
...
+      dbHandler.run(
+        DBIO.sequence(
+          PlatformDiscoveryOperations.verifyAttributesAndGetQueries("fees", "kind", Some("1"))
+        )
+      ).futureValue.flatten shouldBe List("example1")
+
+    }
+
+  }
+
```
ivanopagano: What about adding the edge case where no value is available in the DB?

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryTypes.scala
piotrkosecki: if we introduce more complex types it may, but for the generated ones I don't think so

ivanopagano: Are you sure that the postgres-profile-specific import won't create issues with other slick profiles? I was referring to that essentially

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
piotrkosecki: ok, I'll change it but my reasoning was that I didn't want to repeat DBIO.sequence

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
piotrkosecki: I can only catch `org.scalatest.exceptions.TestFailedException` from what I see which hides my expected exception, so I can check for this exception but I would still need to check for the message - I can change to checking for `TestFailedException`- WDYT?

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
piotrkosecki: ok, changed

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
piotrkosecki: sure, changed

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
piotrkosecki: fixed

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
piotrkosecki: yes, you're right, changed

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
piotrkosecki: changed

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
piotrkosecki: changed

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
piotrkosecki: done

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
piotrkosecki: changed

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
piotrkosecki: ok, added

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
ivanopagano: You can always call `future.failed.futureValue` to get back the exception that failed the future.

Then you check the type with `should be a[NoSuchElementException]`

At least... I guess it should work

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperations.scala
ivanopagano: I thought about that too, but personally I don't think this should be a driving factor behind code design choices. I tend to focus more on the intent and ergonomics of the resulting interface.

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryTypes.scala
piotrkosecki: there are tests in place which are working fine - the whole schema is tested just because of this case. I think we should probably think abut using tests with the same database as we are using in production i.e. `postgresql-embedded` or with some docker containers.

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
piotrkosecki: with `future.failed.futureValue` it does not throw any exception:
`Expected exception java.util.NoSuchElementException to be thrown, but no exception was thrown`.

piotrkosecki: ok, I found `future.value.get` which returns exception it supposed to

ivanopagano: `val ex = future.failed.futureValue` is supposed to be the exception itself, not to throw the exception... 😀

src/main/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryTypes.scala
ivanopagano: As long as tests work it's all fine.

About using embedded pg, it might make sense if it's fast. Using docker is ok for me, but only in integration tests. In my experience it gets burdensome on a daily working cycle. Unit tests are supposed to be slick

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
ivanopagano: Oops, I meant a case when you have no rows in the database to start from

piotrkosecki: Yes, I've added it before insert

PR #208, opened 2018-12-17
Conseil 206: accounts fetch checkpoint

### LINKED ISSUE #206: Checkpointing Account data to fetch in Lorre
After the introduction of #194 to store only diffs of accounts, we lost consistency guarantees between newly fetched blocks and corresponding accounts.

That is because previously all accounts were always rewritten to the latest level at each iteration.
Now only those involved are stored, and any failure (in handling the data or in availability of the services we dependend upon, i.e. the tezos node) might lead to missing account updates that will not be eventually retried at next iteration.

I propose here to "checkpoint" the step before loading accounts, by storing the list of account-ids to be fetched with associated block data, so that at each iteration the list of accounts will be available on db.
A skipped iteration means that more accounts will be added to the checkpoint list.

The list should be optimized during load to only select the latest data (level) for each account, once.

Also the list should be purged every iteration from stale data (old levels) and especially after the fetching completes successfully (full reset)

ivanopagano:
closes #206 

The enhancements stores persistently the accounts ids to be fetched, returned as a side-result from the blocks processing.

Such persistent data is then read to update accounts data, thus avoiding the risk of data loss in the case of failure during the account processing, as the next cycle will simply add new data at the checkpoint, and then fetch what's needed again.

The chages also inverted the halting behaviour on accounts processing failure, now if you need to stop Lorre on failure the env-variable `LORRE_FAILURE_IGNORE` must be set to `"false"` or `"no"`

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
@@ -55,7 +55,7 @@ object TezosDatabaseOperations extends LazyLogging {
 
   /**
     * Writes association of account ids and block data to define accounts that needs update
-    * @param accountIds will have blocks, paired with correspoinding account ids to store
+    * @param accountIds will have blocks, paired with corresponding account ids to store
     * @return Database action possibly returning the rows written (if available form the underlying driver)
     */
   def writeAccountsCheckpoint(accountIds: List[(BlockHash, Int, List[AccountId])]): DBIO[Option[Int]] =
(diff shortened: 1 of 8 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -47,11 +47,58 @@ object TezosDatabaseOperations extends LazyLogging {
...
+    * @return the database action to run
+    */
+  def cleanAccountsCheckpoint(): DBIO[Unit] = Tables.AccountsCheckpoint.schema.truncate
+
+  /**
+    * Reads the account ids in the checkpoint table, considering
+    * only those that at the latest block level (highest value)
+    * @return a database action that loads the list of relevant rows
+    */
+  def getLatestAccountsFromCheckpoint(implicit ec: ExecutionContext): DBIO[Map[BlockReference, List[AccountId]]] =
```
vishakh: I'm wondering whether we should put in some tolerance, i.e. don't fetch any account that was checkpointed more than, let's say, 1,000 blocks ago. This way, we might avoid a scenario where some accounts are just not fetchable and the list grows slowly with each iteration until it slows the whole process down. By adding a tolerance, we give up on an account if it simply cannot be fetched after a reasonable amount of time. It can be left to the DevOps personnel to check the checkpoints table for 'lost' accounts and perform some kind of manual intervention.

vishakh: I just reread the code so I can give you a more concrete example:
- Some account ID was picked up due to a regular expression match. However, the account ID is actually invalid and somehow trying to fetch it causes the Lorre process to crash.
- Lorre restarts and, due to checkpointing, it picks up the old invalid account ID as well, tries again to retrieve it and crashes again.
- This keeps happening and cripples the Lorre process.

For this edge case, my initial proposal was to give up on checkpointed accounts that were more than 'x' blocks old. 

However, after thinking about it, we should not bother about this edge case until our approach for fetching Tezos data settles down in the coming weeks. The contents of this PR are more than good enough for now.

PR #210, opened 2018-12-18
*WiP* Conseil 69 lorre and conseil network parametrising approach

piotrkosecki:
So in this PR there is my concept of parametrising Lorre and Conseil, I would want to start discussion about this approach. This is work in progress so I know that I still have some stuff to change/refactor, but I want to get your opinion first if you agree with this approach.

I think that the best approach is that each network will have separate DB(it doesn't have to be different postgres instance, just DB).

Main changes that I had to do were:
- changing config(moving `conseildb` object deeper), so I could put different instances definitions into it
- extracting database as a dependency - because the code is fairly generic, the only thing that was globally instantiated was the DB, so I had to change that. 
- because of the previous point I had to split some objects into class + companion object

Please take a look and comment if you agree with this approach.


src/main/resources/application.conf
```diff
@@ -18,15 +18,19 @@ lorre: {
...
 
 # Used for configuring Typesafe Slick. Replace database name, user and password in an env-specific config file.
-conseildb = {
-  dataSourceClass = "org.postgresql.ds.PGSimpleDataSource"
-  properties = {
-    databaseName = "conseil"
-    user = "foo"
-    password = "bar"
-    reWriteBatchedInserts = true
+db {
```
vishakh: Could we change the nesting levels so it's like this?

```scala
databases = {
	tezos {
		zeronet = {
			conseildb = {
					dataSourceClass = "org.postgresql.ds.PGSimpleDataSource"
				    properties = {
				    databaseName = "conseil"
				    user = "foo"
				    password = "bar"
				    reWriteBatchedInserts = true
			    }
			  numThreads = 10
			}
		}
	}
}
```

What do you think, @ba205 and @anonymoussprocket ?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -40,8 +39,8 @@ object Lorre extends App with TezosErrors with LazyLogging {
   private val sleepIntervalInSeconds = conf.getInt("lorre.sleepIntervalInSeconds")
   private val feeUpdateInterval = conf.getInt("lorre.feeUpdateInterval")
 
-  lazy val db = DatabaseUtil.db
-  val tezosNodeOperator = new TezosNodeOperator(new TezosNodeInterface())
+  lazy val db = ApiNetworkOperations.getDatabaseMap(network)
```
vishakh: Let's check for both `platform` and `network`. You will have to start asking for `platform` as well from the command line. This might be a good time to use an argument parsing library.

src/main/scala/tech/cryptonomic/conseil/tezos/NetworkConfigOperations.scala
```diff
@@ -0,0 +1,20 @@
+package tech.cryptonomic.conseil.tezos
```
vishakh: This functionality is generic, i.e. applies for Tezos as well as any other chains we support, so it should not be in the `tezos` package.

src/main/scala/tech/cryptonomic/conseil/tezos/NetworkConfigOperations.scala
```diff
@@ -0,0 +1,20 @@
...
+import scala.collection.JavaConverters._
+
+object NetworkConfigOperations {
+  /**
+    * Extracts networks from config file
+    *
+    * @param  config configuration object
+    * @return list of networks from configuration
+    */
+  def getNetworks(config: Config): List[Network] = {
```
vishakh: The function fetches a list of platforms, not networks, so it should be called `getPlatforms()`.

src/main/scala/tech/cryptonomic/conseil/tezos/ApiNetworkOperations.scala
```diff
@@ -0,0 +1,42 @@
...
+
+
+object ApiNetworkOperations {
+  private lazy val networkOperationsMap: Map[String, Database] = getDatabaseMap
+
+  def apply(): ApiNetworkOperations = new ApiNetworkOperations(networkOperationsMap)
+
+  def getDatabaseMap: Map[String, Database] = {
+    val config: Config = ConfigFactory.load()
+    NetworkConfigOperations.getNetworks(config).map { network =>
```
vishakh: Don't forget to check both the network and platform everywhere!

src/main/scala/tech/cryptonomic/conseil/tezos/ApiNetworkOperations.scala
```diff
@@ -0,0 +1,42 @@
...
+import scala.concurrent.ExecutionContext
+import scala.util.Try
+
+
+object ApiNetworkOperations {
+  private lazy val networkOperationsMap: Map[String, Database] = getDatabaseMap
+
+  def apply(): ApiNetworkOperations = new ApiNetworkOperations(networkOperationsMap)
+
+  def getDatabaseMap: Map[String, Database] = {
```
vishakh: Instead of reading in all Slick database configurations, why not replace all this code by simply reading a config key called `databases.{platforms}.{network}.consildb' or whatever?

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -115,6 +112,13 @@ object ApiOperations {
 
   }
 
+  def apply(dbHandle: Database): ApiOperations = new ApiOperations(dbHandle)
+
+}
+
+class ApiOperations(dbHandle: Database) {
```
vishakh: Instead of passing around a database handle all through the code, could we keep the old logic of fetching the handle when needed from `DatabaseUtil`? The code should be as stateless as possible. Every incoming query into Conseil provides the platform and network the query is for. Those can be provided to DatabaseUtil to get a handle each time. In order to prevent multiple IO operations 'DatabaseUtil` could cache the handles in a singleton instance.

Each public method of `ApiOperations` would have to take new `platform` and `network` parameters just as `TezosNodeOperator` does.

src/main/scala/tech/cryptonomic/conseil/db/DatabaseQueryExecution.scala
```diff
@@ -2,13 +2,13 @@ package tech.cryptonomic.conseil.db
...
-
 import slick.jdbc.PostgresProfile.api._
+
 import scala.language.higherKinds
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext, Future}
 
 /** database-specific filter operations support */
-object DatabaseQueryExecution {
+object DatabaseQueryExecutionTypes {
```
ivanopagano: Is this useful or needed in some way?

piotrkosecki: avoid naming conflicts  - so here I had to rename it as I've extracted the class, there was also trait named `DatabaseQueryExecution`

src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
```diff
@@ -35,79 +33,86 @@ object Tezos {
...
-            }
-          } ~ pathPrefix("operations") {
-            path("avgFees") {
-                complete(
-                  handleNoneAsNotFound(ApiOperations.fetchAverageFees(filter))
-                )
-            } ~ pathEnd {
-                completeWithJson(ApiOperations.fetchOperations(filter))
+    apiNetworkOperations.getApiOperations(network) { apiOperations =>
+      apiNetworkOperations.getApiFiltering(network, asyncApiFiltersExecutionContext) { apiFiltering =>
```
ivanopagano: Maybe this might change later, but just as a side-note here... if you make
`...getApiFiltering(...) { implicit apiFiltering => ... }`

will the implicit instances in the filtering be available already in scope for the `apiOperations` methods?

So you won't need to later `import apiFiltering._`...

src/main/scala/tech/cryptonomic/conseil/db/DatabaseQueryExecution.scala
ivanopagano: I can't find where the class with conflicting name is?

piotrkosecki: it's conflicting with the trait

ivanopagano: I see no class `DatabaseQueryExecution`, only the trait and the object, I must be missing something...?

src/main/scala/tech/cryptonomic/conseil/tezos/NetworkConfigOperations.scala
piotrkosecki: this function returns networks, the one below, which I've recently added, returns platforms

src/main/scala/tech/cryptonomic/conseil/tezos/NetworkConfigOperations.scala
piotrkosecki: ok, moved

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
piotrkosecki: Approach you are suggesting has issues which are really hard to overcome, service made this way may seem stateless but dependency on external service(DB in this case) that is hardcoded is not a good idea when it comes to testing. Also keeping DB in the one instance accessible by any other service makes reasoning about the architecture harder.

That's why I'm proposing to extract the database from the singleton object and pass it to the classes instead of putting platform and network to each method. @ivanopagano what do you think?

src/main/scala/tech/cryptonomic/conseil/routes/Tezos.scala
piotrkosecki: it won't work, because implicit instances are inside the object

src/main/scala/tech/cryptonomic/conseil/tezos/ApiNetworkOperations.scala
piotrkosecki: I wanted to avoid creating DB instance each time the call is made, also refactored it so now map keeps objects of ApiOperations to avoid recreating them

src/main/scala/tech/cryptonomic/conseil/tezos/ApiNetworkOperations.scala
piotrkosecki: changed

src/main/scala/tech/cryptonomic/conseil/db/DatabaseQueryExecution.scala
piotrkosecki: yes, it was conflicting with the trait, but it is not relevant any more

src/main/resources/application.conf
piotrkosecki: I've changed it the way you suggested

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: I used simple parsing for now, at the moment we have 2 arguments platform and network, in case it will become more complicated then we should think abut library, at the moment I don't think is necessary

src/main/scala/tech/cryptonomic/conseil/tezos/ApiNetworkOperations.scala
```diff
@@ -0,0 +1,52 @@
...
+    apiFilteringMap.get((platform, network)) match {
+      case Some(value) => provide(value)
+      case None => complete(StatusCodes.NotFound)
+    }
+  }
+
+  private def getDatabaseMap[A](fun: Database => A): Map[(String, String), A] = {
+    NetworkConfigOperations.getPlatforms(config).flatMap { platform =>
+      NetworkConfigOperations.getNetworks(config).map { network =>
+        (platform, network.name) -> Try {
```
ivanopagano: We can put the mapping key (in the tuple sense) _within_ the Option, using `map`, to have a `Option[(keys, value)]`.
Using the implicit conversion from Option to List, the `filter` can go away, as well as the final `_.get` on the values

ivanopagano: Or we can even only define the key/value tuple in the final `yield` in a for-comp

```scala
for {
  platform <- ...
  network <- ...
  validA <- Try(...).toOption
} yield (platform, network.name) -> validA
```

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -193,7 +194,7 @@ class TezosNodeOperator(val node: TezosRPCInterface)(implicit executionContext:
     */
   def getBlocksNotInDatabase(network: String, followFork: Boolean): Future[List[(Block, List[AccountId])]] =
     for {
-      maxLevel <- ApiOperations.fetchMaxLevel
+      maxLevel <- ApiOperations(dbHandle).fetchMaxLevel()
```
ivanopagano: Shouldn't this `ApiOperations`instance be memoized? The TezosNodeOperator will always refer to the same db... we could even pass the `apiOperations` instance as a dependency from the outside for the `TezosNodeOperator`

src/test/scala/tech/cryptonomic/conseil/tezos/PlatformDiscoveryOperationsTest.scala
```diff
@@ -27,6 +28,7 @@ class PlatformDiscoveryOperationsTest
   import slick.jdbc.PostgresProfile.api._
   import scala.concurrent.ExecutionContext.Implicits.global
 
+  val pdo = new PlatformDiscoveryOperations(ApiOperations(dbHandler))
```
ivanopagano: Trivial comment: I tend to use "system under test" (i.e. `sut`) as a conventionally defined name for the test subject.
Doing this by convention in situations like this, could remove ambiguity on what we're testing, and avoid the need to lookup what a new acronym stands for (`pdo` in this case)

ivanopagano: additional note: I didn't make the definition up myself, to be clear
https://en.wikipedia.org/wiki/System_under_test

piotrkosecki:
closing as with that amount of changes I will need to start from scratch

PR #216, opened 2018-12-20
missing unit tests

dorians:


src/main/scala/tech/cryptonomic/conseil/util/JsonUtil.scala
```diff
@@ -33,10 +33,6 @@ object JsonUtil {
   mapper.registerModule(DefaultScalaModule)
   mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
 
-  def toJson(value: Map[Symbol, Any]): JsonString = {
```
ba205: I'm assuming this function was checked to not be used by any other module at all, because the bottom one takes precedence?

dorians: Yes, that’s correct.

Issue #121, opened 2018-07-15
Use an argument parsing library to read command line arguments


vishakh:

ivanopagano:
Isn't this closed ??

dorians:
> Isn't this closed ??

Yes, it is.

PR #217, opened 2018-12-27
Feature/213 allow skipping of chain catch up

dorians:


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -26,14 +25,37 @@ object Lorre extends App with TezosErrors with LazyLogging {
...
+
+    opt[Int]("depth")
+      .action((depth, c) => c.copy(d = depth))
+      .validate(it => if (it >= -1) success else failure("Value <depth> must be >= -1") )
+      .text("how many blocks to synchronize starting with head (use -1 for everything and 0 for only new ones)")
+
+    help("help").text("prints this usage text")
+  }
+
+  private val config: Config = parser.parse(args, Config()).getOrElse(sys.exit(1))
```
vishakh: Can we give more descriptive names to `config` and `Config` so command line configurations and file-based configurations are clearly distinguished?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -82,9 +104,14 @@ object Lorre extends App with TezosErrors with LazyLogging {
...
-    logger.info("Taking a nap")
-    Thread.sleep(sleepIntervalInSeconds * 1000)
-    mainLoop(iteration + 1)
+
+    depth match {
+      case Newest =>
+        logger.info("Taking a nap")
+        Thread.sleep(sleepIntervalInSeconds * 1000)
+        mainLoop(iteration + 1)
+      case _ => ()
```
vishakh: Maybe for this case we can leave a log message saying something like "My work is done."?

build.sbt
```diff
@@ -31,6 +31,7 @@ libraryDependencies  ++=  Seq(
   "com.muquit.libsodiumjna" % "libsodium-jna" % "1.0.4" exclude("org.slf4j", "slf4j-log4j12"),
   "com.github.alanverbner" %% "bip39" % "0.1",
   "ch.megard" %% "akka-http-cors" % "0.3.0",
+  "com.github.scopt" %% "scopt" % "3.7.1",
```
vishakh: Very happy we are finally using scopt. :smile:

vishakh:
While I really like how the code is structured here, I have two concerns:
- Sometimes we might need to debug an issue which we know happened between blocks `x` and `y`, maybe one week in the past. We might not want to process one week of blocks in order to debug our issue. Should we therefore support depth parameters of the form `x:y`, e.g. `1998:2001`?
- Currently, it is just assumed we build up our tables of data incrementally, i.e. we either start from scratch or fetch data from the last block which was stored in the database. But now we have a lot more flexibility which can cause problems if there is already data in the database and we end up double recording a lot of the information. What is the best way to prevent this from happening?

@anonymoussprocket @ba205 @ivanopagano A bit-penny for your thoughts also!

ivanopagano:
Would you add a reference to the git issue in the PR description, please? @dorians

build.sbt
ivanopagano: indeed, yet we might want to use the latest fresh release: http://eed3si9n.com/scopt4 ?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -26,14 +25,37 @@ object Lorre extends App with TezosErrors with LazyLogging {
...
   private[this] val shutdownWait = 10.seconds
 
-  private val network =
-    if (args.length > 0) args(0)
-    else {
-      Console.err.println("""
-      | No tezos network was provided to connect to
-      | Please provide a valid network as an argument to the command line""".stripMargin)
-      sys.exit(1)
+  sealed class Depth()
```
ivanopagano: why class and not trait?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -26,14 +25,37 @@ object Lorre extends App with TezosErrors with LazyLogging {
...
+  case class Everything() extends Depth
+  case class Newest() extends Depth
+  case class Custom(depth: Int) extends Depth
+
+  case class Config(private val d: Int = -1, networks: Seq[String] = Seq()) {
+    def depth = {
+      d match {
```
ivanopagano: What about using a string to allow also
- `depth=new`
- `depth=all`
- `depth=42`
?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -187,33 +185,36 @@ class TezosNodeOperator(val node: TezosRPCInterface)(implicit executionContext:
...
+      blocks <- getBlocks(network, headHash, 1, headLevel)
+    } yield blocks
+
+  /**
+    * Gets last n blocks.
+    * @param network    Which Tezos network to go against
+    * @param amount     Amount of latest block to fetch
+    * @return           Blocks and Account hashes involved
+    */
+  def getLastBlocks(network: String, amount: Int): Future[List[(Block, List[AccountId])]] =
```
ivanopagano: cosmetic: may I suggest `getLatestBlocks`?

build.sbt
dorians: Release candidate? I'm excited :)

anonymoussprocket:
@vishakh:
- should we do this with a anchor block hash plus distance to load, block height range as you suggested, block height anchors plus distance to load? In the short term if the original issue spec is implemented, I see no reason not to close this issue and merge the code. We can define range load as a separated task. 
- blockhash should be a PK and FK into related tables which should structurally prevent loading of dupes. You are correct however that we need a way to identify and fill holes. This could be done with math on blockheight looking for gaps and then filled with the range load function you are proposing.

vishakh:
> @vishakh:
> 
>     * should we do this with a anchor block hash plus distance to load, block height range as you suggested, block height anchors plus distance to load? In the short term if the original issue spec is implemented, I see no reason not to close this issue and merge the code. We can define range load as a separated task.
> 
>     * blockhash should be a PK and FK into related tables which should structurally prevent loading of dupes. You are correct however that we need a way to identify and fill holes. This could be done with math on blockheight looking for gaps and then filled with the range load function you are proposing.

@anonymoussprocket 

Yes, agreed on the anchors and also on doing it as part of a subsequent ticket.

We do have the right foreign keys into the `blocks` table but we removed the uniqueness constraint on block hash in the `blocks` table.

vishakh:
> @vishakh @anonymoussprocket
> 
>     1. Running Lorre with depth=n is for debugging purposes I believe. Or am I missing something? On production, we should use depth=0 only which was default behaviour earlier. I tested this code and indexes on DB prevent from inserting the same block twice.

Even when you are debugging an issue locally, you might want to prevent duplicating data. In any case, this is something we can address in the future once our colleagues give us feedback while using this new option.

Overall, I think this PR looks good and is ready to go as soon as the unit tests and any clean up are done.

anonymoussprocket:
This PR also addresses #121.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -36,7 +35,7 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig {
   val shutdownWait = 10.seconds
 
   //whatever happens we try to clean up
-  sys.addShutdownHook(shutdown)
+  sys.addShutdownHook(shutdown())
```
ivanopagano: Interesting, would this actually change behaviour?

src/test/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperatorTest.scala
```diff
@@ -3,210 +3,64 @@ package tech.cryptonomic.conseil.tezos
...
-    blocks.get.length should be (0)
-  }
-
-  "getBlocks" should "handle an invalid block payload" in {
-    val nodeOp: TezosNodeOperator = new TezosNodeOperator(MockTezosNode)
-    val blocks = nodeOp.getBlocks("zeronet", 0, 2, Some("fakeblock"), followFork = false)
-    blocks.isFailure should be (true)
-  }
+    //given
+    (tezosRPCInterface.runAsyncGetQuery _).when("zeronet", "blocks/BLockGenesisGenesisGenesisGenesisGenesis385e5hNnQTe~").returns(Future.successful(TezosResponseBuilder.blockResponse))
```
ivanopagano: Very subjective: I personally try to avoid using externally defined classes to provide canned values for a specific test, or would try to make it very clear that the response-builder is strictly connected to this specific spec.

Rationale: I've seen how creating such test utilities can entice all devs in the team to build up on the same utility class, which will grow in time to a monster file that will be required to somehow provide a consistent view for all possible sorts of tests, becoming a scourge to maintain...
That's why I try to keep such things defined near the spec at hand, or, when needed, create a more generic utility that could be parametrized progressively to suit each case with explicit configuration.
The pain point is returning highly tailored test-specific data in a seemingly innocuous helper class.

I'd appreciate feedback on this as a chance to improve myself in my testing design skills.

src/main/scala/tech/cryptonomic/conseil/config/LorreAppConfig.scala
```diff
@@ -1,29 +1,58 @@
...
+
+  object Int {
+    def unapply(s: String): Option[Int] = util.Try(s.toInt).toOption
+  }
+
+  private case class ArgumentsConfig(private val d: String = "", networks: Seq[String] = Seq()) {
+
+    def depth: Depth = {
+      d match {
+        case Int(-1) => Everything
```
ivanopagano: slight improvement: 
```scala
case Int(-1) | "all" => Everything
case Int(0)  | "new => Newest
case Int(n) => Custom(n)
```

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -179,33 +177,36 @@ class TezosNodeOperator(val node: TezosRPCInterface, batchConf: BatchFetchConfig
...
+
+  /**
+    * Gets last n blocks.
+    * @param network    Which Tezos network to go against
+    * @param amount     Amount of latest block to fetch
+    * @return           Blocks and Account hashes involved
+    */
+  def getLatestBlocks(network: String, amount: Int): Future[List[(Block, List[AccountId])]] =
+    for {
+      blockHead <- getBlockHead(network)
```
ivanopagano: Can we reduce this to a single call for the head and the previous `amount`?

src/test/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperatorTest.scala
```diff
@@ -3,210 +3,64 @@ package tech.cryptonomic.conseil.tezos
...
+    val block: Future[List[(TezosTypes.Block, List[TezosTypes.AccountId])]] = nodeOp.getAllBlocks("zeronet")
 
-  "getAccounts" should "correctly fetch all accounts for a block" in {
-    val nodeOp: TezosNodeOperator = new TezosNodeOperator(MockTezosNode)
-    val accounts = nodeOp.getAccounts("zeronet", "BKiRLq7c2QVr6X428RRvp6JLTJEnWPE4bc4cAQHoo9GuZz9GH38")
-    accounts.get.accounts.size should be (1) //actually 5 in tezos, 1 for testing purposes.
-    val account = accounts.get.accounts.get("tz1btz5Av9BdpoTPnS9zGyPvpgAovmaZ23iN")
-    account.get.balance should be (12000000000000.0)
+    //then
+    Await.result(block, 1.second) should have length 1
```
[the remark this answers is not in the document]
dorians: Yeah, it would be just a dump response from the stub. I personally don't like these tests because it should be an integration test. I created a task for it to make it more reliable: https://github.com/Cryptonomic/Conseil/issues/222

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
dorians: No, it doesn't change behavior. It's just because IntelliJ yielded warning about a parameterless call.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
dorians: I'm not sure what do you mean. I renamed ``amount`` to ``count`` because I suppose that it was misleading :)

dorians: Ok, now I understand. It's probably possible but it affects many other methods as well. I've created a task to investigate this: https://github.com/Cryptonomic/Conseil/issues/229

ivanopagano: It was not a mandatory requirement, I was just asking to know if it could be simplified *as is**.

I didn't meant to imply further work to make it possible, so it's ok like this

PR #219, opened 2018-12-28
Conseil 207 log commit reference

### LINKED ISSUE #207: Expose source reference in Conseil log
Following the "Bonjour..." line print the GitHub hash and branch of the commit from which the binary was built.

During the build process, get the hash of the commit the code is at

`git log --pretty=format:'%h' -n 1`

and write it into a file under src/main/resources, during application startup Conseil will look for that file since it should have been included in the binary. If the file is not found, Conseil should terminate with an error. If the file is found, the contents would be printed in the log.

ivanopagano:
Closes #207 

Will log both app version and commit-hash to the startup message for both Conseil and Lorre

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Conseil.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Conseil.scala
@@ -53,13 +53,23 @@ object Conseil extends App with LazyLogging with EnableCORSDirectives with Conse
       }
 
       val bindingFuture = Http().bindAndHandle(route, server.hostname, server.port)
-      logger.info("Bonjour...")
+      logger.info("""
+        | =========================***=========================
+        |  Conseil v.{}
+        |  {}
+        | =========================***=========================
+        |
+        |  Bonjour...
+        |""".stripMargin,
+        BuildInfo.version,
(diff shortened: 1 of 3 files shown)


src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -53,13 +53,23 @@ object Conseil extends App with LazyLogging with EnableCORSDirectives with Conse
...
+        BuildInfo.gitHeadCommit.fold("")(hash => s"[commit-hash: ${hash.take(7)}]")
+      )
 
       sys.addShutdownHook {
         bindingFuture
-        .flatMap(_.unbind().andThen{ case _ => logger.info("Server stopped...")} )
-        .flatMap( _ => system.terminate())
-        .onComplete(_ => logger.info("We're done here, nothing else to see"))
+          .flatMap(_.unbind().andThen{ case _ => logger.info("Server stopped...")} )
+          .flatMap( _ => system.terminate())
```
dorians: Why do we need this ``flatMap`? Can we just use ``onComplite`` for this final step?

ivanopagano: I need to sequence nested async operations... 

- unbind the server
- terminate the actor system

or did you mean something different?

PR #224, opened 2019-01-08
Conseil-221: query sort support

piotrkosecki:
extended query definition with ordering support

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
@@ -4,7 +4,7 @@ import com.typesafe.scalalogging.LazyLogging
 import slick.jdbc.PostgresProfile.api._
 import slick.jdbc.SQLActionBuilder
 import tech.cryptonomic.conseil.generic.chain.DataTypes.OperationType.OperationType
-import tech.cryptonomic.conseil.generic.chain.DataTypes.{OperationType, Predicate}
+import tech.cryptonomic.conseil.generic.chain.DataTypes.{OperationType, Predicate, QueryOrdering}
 import tech.cryptonomic.conseil.tezos.FeeOperations._
 import tech.cryptonomic.conseil.tezos.TezosTypes._
 import tech.cryptonomic.conseil.util.CollectionOps._
@@ -374,12 +374,14 @@ object TezosDatabaseOperations extends LazyLogging {
     * @param table          name of the table
     * @param columns        list of column names
     * @param predicates     list of predicates for query to be filtered with
+    * @param ordering       list of ordering conditions for the query
     * @return               list of map of [string, any], which represents list of rows as a map of column name to value
     */
-  def selectWithPredicates(table: String, columns: List[String], predicates: List[Predicate])(implicit ec: ExecutionContext):
+  def selectWithPredicates(table: String, columns: List[String], predicates: List[Predicate], ordering: List[QueryOrdering])(implicit ec: ExecutionContext):
   DBIO[List[Map[String, Any]]] = {
      makeQuery(table, columns)
        .addPredicates(predicates)
+       .addOrdering(ordering)
        .as[Map[String, Any]]
        .map(_.toList)
   }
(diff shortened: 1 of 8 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -107,13 +107,11 @@ object TezosPlatformDiscoveryOperations {
   }
 
   /** Checks if columns exist for the given table */
-  def areFieldsValid(tableName: String, queryFields: List[String], predicateFields: List[String]): Boolean = {
-    val fields = (queryFields ++ predicateFields).toSet
-
+  def areFieldsValid(tableName: String, fields: List[String]): Boolean = {
```
ivanopagano: Would it be useful to require a `fields: Set[String]` in the first place, as a parameter?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -408,6 +410,11 @@ object TezosDatabaseOperations extends LazyLogging {
     sql"""SELECT #$cols FROM #$table WHERE true """
   }
 
+  def makeOrdering(ordering: List[QueryOrdering]): SQLActionBuilder = {
```
ivanopagano: The more gets added to this class, the more I have the feeling that it would be better extracted to a general database module/class that builds generic queries on top of Slick.
There is a lot of logic now that is not tied to tezos specifically, and could be more aptly moved to DatabaseUtil, or to a similar QueryBuilder class, that would keep coupled things together

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
piotrkosecki: It could, but if I remember correctly this is the only place I use `Set` features

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
```diff
@@ -1148,6 +1148,124 @@ class TezosDatabaseOperationsTest
...
+      } yield found
+
+      val result = dbHandler.run(populateAndTest.transactionally).futureValue
+      result shouldBe List(
+        Map("level" -> 0, "proto" -> 1, "protocol" -> "protocol", "hash" -> "R0NpYZuUeF"),
+        Map("level" -> 1, "proto" -> 1, "protocol" -> "protocol", "hash" -> "aQeGrbXCmG")
+      )
+    }
+
+    "get map from a block table and sort by hash in ascending order and by hash in ascending order" in {
```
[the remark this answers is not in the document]
piotrkosecki: OK, I've had a wrong description, the idea behind this test is that the table is first sorted by the one column, and as a secondary thing by the other column

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
ivanopagano: what I meant is that putting that as a "type requirement" in the method signature will make for a  more specific and clearly defined function semantic

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
```diff
@@ -685,7 +685,7 @@ class TezosDatabaseOperationsTest
 
       val populateAndTest = for {
         _ <- Tables.Blocks ++= blocksTmp
-        found <- sut.selectWithPredicates(Tables.Blocks.baseTableRow.tableName, columns, List.empty)
+        found <- sut.selectWithPredicates(Tables.Blocks.baseTableRow.tableName, columns, List.empty, List.empty)
```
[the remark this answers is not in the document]
piotrkosecki: yes, probably, but for the next ticket I'll need another parameter, and I don't want to do everything with default ones

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: You are right, I'll refactor this in the new PR with ticket I've got about adding limiting amount of rows returned

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
piotrkosecki: I'm not convinced but I've done it 😄

PR #226, opened 2019-01-11
Conseil 223 query result set limit support

piotrkosecki:
Added support of the `limit` field in the query and refactored a bit moving methods into the `QueryBuilder` object inside of `DatabaseUtil`.  Should be merged after PR #224

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
@@ -12,6 +12,12 @@ import scala.util.Try
   */
 object DataTypes {
 
+  /** Default value of limit parameter */
+  val defaultLimitValue: Int = 10000
+
+  /** Max value of limit parameter */
+  val maxLimitValue: Int = 100000
+
   /** Class required for OperationType enum serialization */
   class OperationTypeRef extends TypeReference[OperationType.type]
 
@@ -39,7 +45,7 @@ object DataTypes {
     fields: List[String] = List.empty,
     predicates: List[Predicate],
     orderBy: List[QueryOrdering] = List.empty,
-    limit: Int = 10000
+    limit: Option[Int] = Some(defaultLimitValue)
   ) {
     /** Method which validates query fields, as jackson runs on top of runtime reflection so NPE can happen if fields are missing */
     def validate: Option[Query] = {
--- a/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
@@ -38,7 +38,8 @@ object DataTypes {
   case class Query(
     fields: List[String] = List.empty,
     predicates: List[Predicate],
-    orderBy: List[QueryOrdering] = List.empty
+    orderBy: List[QueryOrdering] = List.empty,
+    limit: Int = 10000
   ) {
     /** Method which validates query fields, as jackson runs on top of runtime reflection so NPE can happen if fields are missing */
     def validate: Option[Query] = {
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
@@ -143,7 +143,8 @@ object ApiOperations extends DataOperations {
             operation = OperationType.in,
             set = accountDelegates.toList
           )
-        ).filter(_.set.nonEmpty)
+        ).filter(_.set.nonEmpty),
+        limit = limit.getOrElse(defaultLimit)
       )
     }
   }
@@ -392,7 +393,15 @@ object ApiOperations extends DataOperations {
     * */
   override def queryWithPredicates(tableName: String, query: Query)(implicit ec: ExecutionContext): Future[List[Map[String, Any]]] = {
     if (areFieldsValid(tableName, (query.fields ++ query.predicates.map(_.field) ++ query.orderBy.map(_.field)).toSet)) {
-      runQuery(TezosDatabaseOperations.selectWithPredicates(tableName, query.fields, sanitizePredicates(query.predicates), query.orderBy))
+      runQuery(
+        TezosDatabaseOperations.selectWithPredicates(
+          tableName,
+          query.fields,
+          sanitizePredicates(query.predicates),
+          query.orderBy,
+          Math.min(query.limit, 100000)
+        )
+      )
     } else {
       Future.successful(List.empty)
     }
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
@@ -1,7 +1,7 @@
 package tech.cryptonomic.conseil.tezos
 
 import slick.jdbc.PostgresProfile.api._
-import tech.cryptonomic.conseil.generic.chain.DataOperations
+import tech.cryptonomic.conseil.generic.chain.{DataOperations, DataTypes}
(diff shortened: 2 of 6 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -391,8 +392,16 @@ object ApiOperations extends DataOperations {
...
-    if (areFieldsValid(tableName, query.fields, query.predicates.map(_.field))) {
-      runQuery(TezosDatabaseOperations.selectWithPredicates(tableName, query.fields, sanitizePredicates(query.predicates)))
+    if (areFieldsValid(tableName, (query.fields ++ query.predicates.map(_.field) ++ query.orderBy.map(_.field)).toSet)) {
+      runQuery(
+        TezosDatabaseOperations.selectWithPredicates(
+          tableName,
+          query.fields,
+          sanitizePredicates(query.predicates),
+          query.orderBy,
+          Math.min(query.limit, 100000)
```
anonymoussprocket: 100000 should be a constant defined somewhere by name.

anonymoussprocket: How would I get *all* records that are available?

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -23,10 +24,22 @@ object DataTypes {
...
+  class QueryOrderingRef extends TypeReference[OrderDirection.type]
+  case class QueryOrdering(field: String, @JsonScalaEnumeration(classOf[QueryOrderingRef]) direction: OrderDirection)
+
   /** Class representing query */
   case class Query(
     fields: List[String] = List.empty,
-    predicates: List[Predicate]
+    predicates: List[Predicate],
+    orderBy: List[QueryOrdering] = List.empty,
+    limit: Int = 10000
```
anonymoussprocket: 10000 should be a named constant

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
piotrkosecki: In this case there is no way to take all of the records. I thought that's what you meant by `Hard limit, at least for now, should be 100000.` in the ticket description.

piotrkosecki: ok, defined

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: ok, defined

src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
```diff
@@ -2,64 +2,154 @@ package tech.cryptonomic.conseil.util
...
-        })
+  object QueryBuilder {
+    /** Concatenates SQLActionsBuilders
+      * Slick does not support easy concatenation of actions so we need this function based on https://github.com/slick/slick/issues/1161
+      *
+      * @param acc     base SqlActionBuilder with which we want concatenate other actions
+      * @param actions list of actions to concatenate
+      * @return one SQLActionBuilder containing concatenated actions
+      */
+    def concatenateSqlActions(acc: SQLActionBuilder, actions: List[SQLActionBuilder]): SQLActionBuilder = {
```
ivanopagano: I have a feeling that multiple `actions` are never passed in, so we have to uselessly wrap every single builder passed to this method in a `List(_)` constructor...

An alternative to keep it possible to pass multiple builders would be to make it a vararg
`def contatenateSqlActions(first: SQLActionBuilder, actions: SQLActionBuilder*): SQLActionBuilder = ...`

The implementation won't actually change but calling the method would be smoother

src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
```diff
@@ -2,64 +2,154 @@ package tech.cryptonomic.conseil.util
...
+        concatenateSqlActions(action, makePredicates(predicates))
+      }
+
+      /** Method for adding ordering to existing SQLAction
+        *
+        * @param ordering list of QueryOrdering to add
+        * @return new SQLActionBuilder containing ordering statements
+        */
+      def addOrdering(ordering: List[QueryOrdering]): SQLActionBuilder = {
+        val queryOrdering = if (ordering.isEmpty) {
```
ivanopagano: I suggest moving this check within `makeOrdering` as per so-called [postel's law](https://devopedia.org/postel-s-law).

The gist of it, in this case, is to avoid forcing the caller to verify general constraints as pre-conditions to the arguments, while having the target method handle corner cases if it can be made gracefully

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -23,10 +30,22 @@ object DataTypes {
...
+  class QueryOrderingRef extends TypeReference[OrderDirection.type]
+  case class QueryOrdering(field: String, @JsonScalaEnumeration(classOf[QueryOrderingRef]) direction: OrderDirection)
+
   /** Class representing query */
   case class Query(
     fields: List[String] = List.empty,
-    predicates: List[Predicate]
+    predicates: List[Predicate],
+    orderBy: List[QueryOrdering] = List.empty,
+    limit: Option[Int] = Some(defaultLimitValue)
```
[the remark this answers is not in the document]
piotrkosecki: the problem lays within a jackson, I discovered it doesn't support default parameter if it is not an `Option`. https://github.com/FasterXML/jackson-module-scala/issues/140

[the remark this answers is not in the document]
ivanopagano: what a shame... cool then

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -374,53 +372,19 @@ object TezosDatabaseOperations extends LazyLogging {
     * @param table          name of the table
     * @param columns        list of column names
     * @param predicates     list of predicates for query to be filtered with
+    * @param ordering       list of ordering conditions for the query
     * @return               list of map of [string, any], which represents list of rows as a map of column name to value
     */
-  def selectWithPredicates(table: String, columns: List[String], predicates: List[Predicate])(implicit ec: ExecutionContext):
+  def selectWithPredicates(table: String, columns: List[String], predicates: List[Predicate], ordering: List[QueryOrdering], limit: Int)
```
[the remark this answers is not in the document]
piotrkosecki: ok, indented

src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
piotrkosecki: you are right, thats a good idea, changed it

src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
piotrkosecki: ok, I've changed it, but I'm not 100% convinced, because this way I'm creating empty `SQLActionBuilder` and merging it when there is no ordering defined

ivanopagano: You're right, I didn't realize that `makeOrdering` was returning a single action, instead of a list...

piotrkosecki: reverted

PR #227, opened 2019-01-14
Conseil 196 basic metadata caching

piotrkosecki:
Used basic metadata caching to improve the performance of the requests. Please note that the config values may(or even should) change.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -125,10 +125,10 @@ akka {
   # akka-http-caching config described here:
   # https://doc.akka.io/docs/akka-http/current/common/caching.html#frequency-biased-lfu-cache
   http.caching.lfu-cache {
-    max-capacity = 50
-    initial-capacity = 25
-    time-to-live = 60 seconds
-    time-to-idle = 30 seconds
+    max-capacity = 500
+    initial-capacity = 100
+    time-to-live = 5 minutes
+    time-to-idle = 3 minutes
   }
 
 }
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -122,11 +122,13 @@ akka {
     }
   }
 
+  # akka-http-caching config described here:
+  # https://doc.akka.io/docs/akka-http/current/common/caching.html#frequency-biased-lfu-cache
   http.caching.lfu-cache {
     max-capacity = 50
     initial-capacity = 25
-    time-to-live = 20 seconds
-    time-to-idle = 10 seconds
+    time-to-live = 60 seconds
+    time-to-idle = 30 seconds
   }
 
 }
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -122,4 +122,11 @@ akka {
     }
   }
 
+  http.caching.lfu-cache {
+    max-capacity = 50
+    initial-capacity = 25
+    time-to-live = 20 seconds
+    time-to-idle = 10 seconds
+  }
+
 }
--- a/src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
@@ -28,16 +28,23 @@ object PlatformDiscovery {
   * @param apiExecutionContext is used to call the async operations exposed by the api service
   */
 class PlatformDiscovery(config: PlatformsConfiguration)(implicit apiExecutionContext: ExecutionContext, system: ActorSystem) extends LazyLogging with RouteHandling {
-  val defaultCachingSettings: CachingSettings = CachingSettings(system)
 
-  val simpleKeyer: PartialFunction[RequestContext, Uri] = {
+  /** default caching settings*/
+  private val defaultCachingSettings: CachingSettings = CachingSettings(system)
+
+  /** simple partial function for filtering */
+  private val simpleKeyer: PartialFunction[RequestContext, Uri] = {
     case r: RequestContext if r.request.method == HttpMethods.GET => r.request.uri
   }
-  val cachingSettings: CachingSettings =
+
+  /** LFU caching settings */
+  private val cachingSettings: CachingSettings =
     defaultCachingSettings.withLfuCacheSettings(defaultCachingSettings.lfuCacheSettings)
 
-  val lfuCache: Cache[Uri, RouteResult] = LfuCache(cachingSettings)
+  /** LFU cache */
+  private val lfuCache: Cache[Uri, RouteResult] = LfuCache(cachingSettings)
 
+  /** Metadata route */
   val route: Route =
     get {
       cache(lfuCache, simpleKeyer) {
--- a/src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
@@ -1,18 +1,25 @@
 package tech.cryptonomic.conseil.routes
 
+import akka.actor.ActorSystem
+import akka.http.caching.LfuCache
+import akka.http.caching.scaladsl.{Cache, CachingSettings, LfuCacheSettings}
+import akka.http.scaladsl.model.{HttpMethods, Uri}
 import akka.http.scaladsl.server.Directives._
-import akka.http.scaladsl.server.Route
+import akka.http.scaladsl.server.directives.CachingDirectives._
+import akka.http.scaladsl.server.{RequestContext, Route, RouteResult}
 import com.typesafe.scalalogging.LazyLogging
 import tech.cryptonomic.conseil.config.Platforms.PlatformsConfiguration
 import tech.cryptonomic.conseil.tezos.TezosPlatformDiscoveryOperations
 import tech.cryptonomic.conseil.util.JsonUtil._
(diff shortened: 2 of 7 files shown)


src/main/resources/application.conf
```diff
@@ -122,4 +122,13 @@ akka {
     }
   }
 
+  # akka-http-caching config described here:
+  # https://doc.akka.io/docs/akka-http/current/common/caching.html#frequency-biased-lfu-cache
+  http.caching.lfu-cache {
```
piotrkosecki: I need input on what values we would prefer to have here:

vishakh: Any thoughts, @anonymoussprocket ?

vishakh:
I am not familiar with Akka caching so maybe @ivanopagano or @dorians can provide a better review. From my brief look, this looks like a reasonable caching mechanism. Are there any trade offs or limitations we should be aware of?

src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
```diff
@@ -12,7 +17,8 @@ import scala.concurrent.ExecutionContext
 
 /** Companion object providing apply implementation */
 object PlatformDiscovery {
-  def apply(config:  PlatformsConfiguration)(implicit apiExecutionContext: ExecutionContext): PlatformDiscovery = new PlatformDiscovery(config)
+  def apply(config:  PlatformsConfiguration)(implicit apiExecutionContext: ExecutionContext, system: ActorSystem): PlatformDiscovery =
```
ivanopagano: Is it still useful or meaningful to keep the ExecutionContext implicit, here and in the other routes classes?

src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
```diff
@@ -21,36 +27,55 @@ object PlatformDiscovery {
   * @param config              configuration object
   * @param apiExecutionContext is used to call the async operations exposed by the api service
   */
-class PlatformDiscovery(config: PlatformsConfiguration)(implicit apiExecutionContext: ExecutionContext) extends LazyLogging with RouteHandling {
+class PlatformDiscovery(config: PlatformsConfiguration)(implicit apiExecutionContext: ExecutionContext, system: ActorSystem) extends LazyLogging with RouteHandling {
+
+  /** default caching settings*/
+  private val defaultCachingSettings: CachingSettings = CachingSettings(system)
```
ivanopagano: Instead of passing the `system` to indirectly read the caching config, we should first validate the configuration in a wrapping pureconfig object and then pass that down here to use with the `Config` based `CachingSettings.apply`.

For reference, you can look into how HostConnectionPooling configuration is passed [here](https://github.com/Cryptonomic/Conseil/blob/4e6f32bbcab0289542d46d4a795e540e9eddb03a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeInterface.scala#L181), and how it's initially validated

src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
```diff
@@ -21,36 +27,55 @@ object PlatformDiscovery {
...
+
+  /** simple partial function for filtering */
+  private val simpleKeyer: PartialFunction[RequestContext, Uri] = {
+    case r: RequestContext if r.request.method == HttpMethods.GET => r.request.uri
```
ivanopagano: is it necessary to check the method if the cache directive is applied downstream the `get` directive?

src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
```diff
@@ -21,36 +27,55 @@ object PlatformDiscovery {
...
   * @param apiExecutionContext is used to call the async operations exposed by the api service
   */
-class PlatformDiscovery(config: PlatformsConfiguration)(implicit apiExecutionContext: ExecutionContext) extends LazyLogging with RouteHandling {
+class PlatformDiscovery(config: PlatformsConfiguration)(implicit apiExecutionContext: ExecutionContext, system: ActorSystem) extends LazyLogging with RouteHandling {
+
+  /** default caching settings*/
+  private val defaultCachingSettings: CachingSettings = CachingSettings(system)
+
+  /** simple partial function for filtering */
+  private val simpleKeyer: PartialFunction[RequestContext, Uri] = {
```
ivanopagano: Style: I'd rename to something less "example-looking"...
requestCacheKeyer, extractCacheKey, ...

src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
piotrkosecki: I'm not sure about that, `CachingSettings.apply` takes config from whole path `akka.http.caching.lfu-cache`, so I would have to keep whole config there

src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
piotrkosecki: not necessary - I used the one from example,

src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
piotrkosecki: ok, changed

src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
ivanopagano: if you look at the whole process of how the host-pool is defined, you'll find that you can read a "sub-config" from anywhere and then put it at a given path, such as `akka.http.caching.lfu-cache`, as expected by akka.

src/main/resources/application.conf
anonymoussprocket: I was thinking more an in-memory cache of the refdata requests between the database and the service, but this may work just as well.

src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
piotrkosecki: I've removed the checking for `get`

src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
piotrkosecki: I've refactored a bit reading config to be more generic, as I could reuse it

src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
piotrkosecki: not useful when I've added

src/main/resources/application.conf
piotrkosecki: I was thinking about more how many entries we should store, and values for the `time-to-live` and `time-to-idle` params.

src/main/scala/tech/cryptonomic/conseil/util/ConfigUtil.scala
```diff
@@ -119,45 +125,57 @@ object ConfigUtil {
...
+      * @param namespace the path where the custom configuration will be searched-for
+      */
+    def loadAkkaCacheConfig(namespace: String): Either[ConfigReaderFailures, HttpCacheConfiguration] = {
+      // this is where akka searches for the config entry for host connection pool
+      loadConfigForEntryPath(namespace, "akka.http.caching.lfu-cache").map(HttpCacheConfiguration)
+    }
+
+    private def loadConfigForEntryPath(namespace: String, referenceEntryPath: String): Either[ConfigReaderFailures, Config] = {
+      //read a conseil-specific entry into the expected path for the config
+      def loadHostPoolConfig(rootConfig: Config): Either[ConfigReaderFailures, Config] =
```
ivanopagano: I think this might need renaming, since now it's path-agnostic and not specific to host-pool...

src/main/scala/tech/cryptonomic/conseil/util/ConfigUtil.scala
```diff
@@ -119,45 +125,57 @@ object ConfigUtil {
...
+          .left.map {
+          //wraps the error into pureconfig's one
+          t => ConfigReaderFailures(new ThrowableFailure(t, None))
+        }
+
+      //compose the configuration reading outcomes
+      for {
+        akkaConf <- pureconfig.loadConfig[Config]
+        tezosClientConfig <- loadHostPoolConfig(akkaConf)
+      } yield tezosClientConfig
```
ivanopagano: maybe we should now rename `tezosClientConfig` to something more general too

src/main/scala/tech/cryptonomic/conseil/util/ConfigUtil.scala
piotrkosecki: renamed, please tell if it's ok

src/main/scala/tech/cryptonomic/conseil/util/ConfigUtil.scala
piotrkosecki: renamed

PR #232, opened 2019-01-21
Update of Conseil Tezos operations and new ADT

### LINKED ISSUE #237: Tezos manager pubkey field is different in the schema between alphanet and zeronet
On alphanet: `managerPubkey`
On zeronet: `manager_pubkey` (more consistent with the rest)

The new version of operations from PR #232 will count on a specific format, and thus fail if we use the same Conseil between the two networks.

Should we work to handle gracefully the discrepancy or handle it as an error.

@vishakh @anonymoussprocket

ivanopagano:
**WIP - Needs local testing and missing features**

fix #237 

- [x] updates the operations table structure to include new data
- [x] updates the data model definition for tezos operations, using a fine-grained typed ADT
- [x] uses Circe to parse json into the ADT, handling all parsing errors at compile-time
- [x] uses a type class definition to extract the convertion operations and apply it for model-to-db-rows conversions
- [x] adds detailed tests to verify correct write and conversions to persistent data
- [x] remove the legacy design for detailed operations' scala model
- [x] adds balance updates as a separate table (in a separate PR)
- [x] increase Lorre's resilience by handling malformed json with temporary storage for later re-processing (in a separate PR)


src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -0,0 +1,291 @@
+package tech.cryptonomic.conseil.tezos
+
+import tech.cryptonomic.conseil.tezos.TezosTypes._
+import tech.cryptonomic.conseil.model.Model
+import tech.cryptonomic.conseil.tezos.FeeOperations._
+import tech.cryptonomic.conseil.util.Conversion
+import tech.cryptonomic.conseil.util.Conversion.Id
+
+object DatabaseConversions {
```
piotrkosecki: I think this library could help us avoid some of boilerplate code in this object: https://github.com/scalalandio/chimney

What do you think?

ivanopagano: Thanks, I know about the library, of course 😄 
I still didn't use any additional resource to avoid obfuscating the code behind new apis.
This is already introducing a new concept to be handled for anyone not familiar with typeclasses, so I tried to stick to manual and explicit conversion, as long as it's manageable. (part of the code was simply moved from another existing class to here).

To sum it up, I'll have a better look at Chimney to evaluate if it provides substantial help without adding extra complexity

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -128,9 +129,20 @@ class TezosNodeOperator(val node: TezosRPCInterface, batchConf: BatchFetchConfig
     * @param blockHash Hash of the block
     * @return          The `Future` list of operations
     */
-  def getAllOperationsForBlock(network: String, blockHash: BlockHash): Future[List[OperationGroup]] =
+  def getAllOperationsForBlock(network: String, blockHash: BlockHash): Future[List[TezosOperations.Group]] = {
+    import io.circe.parser.decode
```
vishakh: Out of curiosity, how come these imports are not at the top?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -36,6 +60,199 @@ object TezosTypes {
...
+        change: Long,
+        category: Option[String],
+        contract: Option[ContractId],
+        delegate: Option[PublicKeyHash],
+        level: Option[Int]
+      )
+    }
+
+    /** a grouping of operations with common "header" information */
+    final case class Group (
```
vishakh: Can we call this `OperationGroup`?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -128,9 +129,20 @@ class TezosNodeOperator(val node: TezosRPCInterface, batchConf: BatchFetchConfig
...
+    import JsonDecoders.Circe.Operations._
```
vishakh: Is it possible to avoid wildcard imports? It makes the code harder to follow.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -175,38 +185,26 @@ class TezosNodeOperator(val node: TezosRPCInterface, batchConf: BatchFetchConfig
...
     */
-  def getLatestBlocks(network: String, count: Int): Future[List[(Block, List[AccountId])]] =
-    for {
-      blockHead <- getBlockHead(network)
-      headLevel = blockHead.metadata.header.level
-      headHash = blockHead.metadata.hash
-      blocks <- getBlocks(network, headHash, max(1, headLevel - count + 1), headLevel)
-    } yield blocks
+  def getLatestBlocks(network: String, depth: Option[Int] = None): Future[List[(Block, List[AccountId])]] =
+    getBlockHead(network).flatMap {
```
vishakh: Could you explain the nature of the change here?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -243,19 +245,26 @@ class TezosNodeOperator(val node: TezosRPCInterface, batchConf: BatchFetchConfig
...
-    val jsonToOperationsAndAccounts: ((BlockHash, String)) => (BlockHash, List[OperationGroup], List[AccountId]) = {
+    val jsonToOperationsAndAccounts: ((BlockHash, String)) => JsonDecoded[(BlockHash, List[TezosOperations.Group], List[AccountId])] = {
       case (hash, json) =>
-        (hash, jsonToOperationGroups(json), jsonToAccountInvolved(json))
+        jsonToOperationGroups(json).map( groups => (hash, groups, jsonToAccountInvolved(json)))
     }
 
     val isGenesis = (metadata: BlockMetadata) => metadata.header.level == 0
 
+    def decodeOperations(in: List[(BlockHash, String)]): Future[List[(BlockHash, List[TezosOperations.Group], List[AccountId])]] =
```
vishakh: Could we please document this function?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -5,15 +5,39 @@ package tech.cryptonomic.conseil.tezos
...
+
   final case class AccountId(id: String) extends AnyVal
 
+  final case class ChainId(id: String) extends AnyVal
+
+  final case class Nonce(value: String) extends AnyVal
+
+  final case class Secret(value: String) extends AnyVal
+
+  final case class MichelsonV1(expression: String) extends AnyVal
```
vishakh: Michelson is the normal representation of Michelson code. Micheline is the JSON representation. It seems we are storing the latter here.

src/main/scala/tech/cryptonomic/conseil/util/Conversion.scala
```diff
@@ -0,0 +1,27 @@
+package tech.cryptonomic.conseil.util
+
+trait Conversion[F[_], FROM, TO] {
+
+  def convert(from: FROM): F[TO]
+
+}
+
+object Conversion {
```
vishakh: Reminder to document this file when you get to it.

doc/conseil.sql
```diff
@@ -122,21 +122,32 @@ CREATE TABLE public.operation_groups (
 --
 
 CREATE TABLE public.operations (
+    operation_id integer NOT NULL,
```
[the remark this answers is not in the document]
ivanopagano: There was a `level` field already in the table, `block_level` was redundant.
I can rename `level` to `block_level` anytime if better

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -65,18 +65,20 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig {
...
-    val processResult =
+    val attemptedProcessing =
       if (sys.env.get("LORRE_FAILURE_IGNORE").forall(ignore => ignore == "false" || ignore == "no"))
+        processing
+      else
         processing.recover {
-          case f @ AccountsProcessingFailed(_, _) => throw f
-          case _ => () //swallowed
+          //swallow the error and proceed with the default behaviour
+          case f@(AccountsProcessingFailed(_, _) | BlocksProcessingFailed(_, _)) => ()
```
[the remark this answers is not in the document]
ivanopagano: I'll explain with an example

`case binding @ SomeCaseClass(a, b, c) => ... `

Will pattern match against an instance of `SomeCaseClass` based on it's constructor "shape" or the `unapply` method.
It will bind `a`, `b`, `c` to the arguments, and at the same time will bind _the whole instance_ to the `binding` variable

The space can be omitted (but I use it for readability)
`case binding@SomeCaseClass(a, b, c) => ...`

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -111,24 +113,28 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig {
...
-          .andThen {
-            case Success(accountsCount) =>
-              logger.info("Wrote {} blocks to the database, checkpoint stored for{} account updates", blocksWithAccounts.size, accountsCount.fold("")(" " + _))
-            case Failure(e) =>
-              logger.error(s"Could not write blocks or accounts checkpoints to the database because $e")
-          }
-          .map(_ => Done)
-    }.andThen {
+
+        def logOutcome[A](outcome: Future[Option[A]]): Future[Option[A]] = outcome.andThen {
```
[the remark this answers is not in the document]
ivanopagano: I tried to extract that and create a common pattern to use both here and in the accounts, but it needed so much parameterization that it was almost unreadable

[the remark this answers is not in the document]
ivanopagano: Having the `andThen` to log the outcome of the operation was cluttering the code flow.
I extracted the logging effect into this wrapping function, so that the logic was separated from the main operational flow

src/main/scala/tech/cryptonomic/conseil/db/DatabaseQueryExecution.scala
```diff
@@ -644,7 +644,7 @@ trait DatabaseApiFiltering {
         case Some(validAction) =>
           ensuringBlocksExist {
             validAction.distinct
-              .sortBy(_.blockLevel.desc)
+              .sortBy(_.level.desc)
```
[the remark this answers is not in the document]
ivanopagano: I need to double-check, it could be that the bug was always there, or that we missed a place that needed refactor along the road

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ivanopagano: This import is not used to enable visibility of names in the namespace, it's used to bring into scope all implicit decoders needed in scope.

To be honest I've never seen anyone explicitly importing the implicit names. This is how the pattern is meant to be used, to make implicits available

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ivanopagano: I merged two operations was concrete difference was only in a default parameter value.

While java encourages to overload methods with different signatures, it's common in scala to use default parameters to avoid repetition.

In this specific case, I expect `getLatestBlocks` and `getAllBlocks` to essentially do the same thing, while one is limited in the result size. So I opted to put the logic in one place, to avoid duplication whenever the logic changes.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ivanopagano: I need to add documentation in a few places still, got it!

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
ivanopagano: Great, I'll rename

src/main/scala/tech/cryptonomic/conseil/util/Conversion.scala
ivanopagano: Sure, I missed some doc here and there and will cover everything before the final merge

src/main/scala/tech/cryptonomic/conseil/db/DatabaseQueryExecution.scala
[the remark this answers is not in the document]
ivanopagano: I confirm that as I introduced this typeclass pattern for the ApiFiltering, the original endpoint for operations didn't support custom sorting of results.

If that's useful, I suggest we create an issue to add that feature.
Let me know if I should.

src/main/scala/tech/cryptonomic/conseil/model/Model.scala
```diff
@@ -0,0 +1,43 @@
...
+
+object Model {
+
+  /** A class useful to obviate for slick's encoding of a
+    * operation row as a HList object, whose fields are not
+    * as easy to extract, if not by position
+    * Any Query result from Tables.OperationsRow can be converted
+    * to this representation by using `.mapTo[Model.Operation]`
+    */
+  case class Operation(
```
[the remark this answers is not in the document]
ivanopagano: This is actually an extension of the database model, but I agree that having it into the `tezos` package makes a lot of sense.
It reflects conseil's model of the tezos operation, and not the actual tezos model, so I'd keep it in a separate file still.

src/main/scala/tech/cryptonomic/conseil/util/JsonDecoders.scala
```diff
@@ -0,0 +1,168 @@
+package tech.cryptonomic.conseil.util
```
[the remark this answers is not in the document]
ivanopagano: Absolutely, I'll move this

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsData.scala
```diff
@@ -0,0 +1,521 @@
+package tech.cryptonomic.conseil.tezos
```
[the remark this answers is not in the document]
ivanopagano: Let's talk about it, I've added a few classes that doesn't actually describe any test spec, but the data needed to actually run the tests.
Keeping all that code into the test itself was becoming cumbersome and noisy... 

I'll rename these classes to `TestFixtures`, so the tie to tests is more direct.

src/test/scala/tech/cryptonomic/conseil/util/JsonDecodersData.scala
```diff
@@ -0,0 +1,729 @@
+package tech.cryptonomic.conseil.util
```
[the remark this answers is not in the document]
ivanopagano: see the previous comment about `TezosDatabaseOperationData`

src/test/scala/tech/cryptonomic/conseil/util/RandomUtil.scala
```diff
@@ -0,0 +1,24 @@
+package tech.cryptonomic.conseil.util
```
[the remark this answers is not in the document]
ivanopagano: see the previous comment about `TezosDatabaseOperationData`

src/test/scala/tech/cryptonomic/conseil/tezos/DatabaseConversionsData.scala
```diff
@@ -0,0 +1,282 @@
+package tech.cryptonomic.conseil.tezos
```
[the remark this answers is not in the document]
ivanopagano: see the previous comment about `TezosDatabaseOperationData`

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala

src/main/scala/tech/cryptonomic/conseil/util/Conversion.scala
ivanopagano: done

src/main/scala/tech/cryptonomic/conseil/model/Model.scala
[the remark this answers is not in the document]
ivanopagano: moved

src/main/scala/tech/cryptonomic/conseil/tezos/Model.scala
```diff
@@ -0,0 +1,43 @@
+package tech.cryptonomic.conseil.tezos
+
+import scala.math.BigDecimal
+
+object Model {
```
piotrkosecki: I think this name of object is too generic, basically every case class/object used in the code could belong here.

ivanopagano: I guess you're right, I couldn't come with anything better... any suggestion?
It's a container for scala classes mapped to conseil table row

`DBTableModel`, `DBTableMapping`?? Not easy to come up with a valid name

vishakh: I like `DBTableMapping`.

doc/conseil.sql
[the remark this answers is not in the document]
vishakh: I think this was a different `level` field. I suggest having a `block_level` column so it's clear it's a 'foreign' column.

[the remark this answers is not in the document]
ivanopagano: I'm resurrecting the `block_level` field, it's a different think, as you said

src/main/scala/tech/cryptonomic/conseil/tezos/Model.scala
ivanopagano: done

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
ivanopagano: I might consider this, yet in a separate task and PR, this is getting too chubby for me

vishakh:
Did another review and did not observe any issues. Happy to look again if there are further changes.

PR #233, opened 2019-01-23
Conseil 230 v2 query bug fixes

piotrkosecki:
Things I've fixed/added:
- api key verification
- platform verification
- added docs with each operation example
- query and predicate fields validation
- extended values left for sql so value could represent timestamp in ISO8601 format

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
@@ -49,8 +49,6 @@ object DataTypes {
   /** Class representing invalid predicate field */
   case class InvalidPredicateField(message: String) extends QueryValidationError
 
-  case class ErrorFields()
-
   /** Class representing query */
   case class Query(
     fields: List[String] = List.empty,
--- a/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
@@ -1,5 +1,7 @@
 package tech.cryptonomic.conseil.generic.chain
 
+import java.io.Serializable
+
 import com.fasterxml.jackson.core.`type`.TypeReference
 import com.fasterxml.jackson.module.scala.JsonScalaEnumeration
 import tech.cryptonomic.conseil.generic.chain.DataTypes.OperationType.OperationType
@@ -20,7 +22,7 @@ object DataTypes {
   val maxLimitValue: Int = 100000
 
   /** Trait representing query validation errors */
-  sealed trait QueryValidationError {
+  sealed trait QueryValidationError extends Product with Serializable {
     val message: String
   }
 
@@ -47,8 +49,7 @@ object DataTypes {
   /** Class representing invalid predicate field */
   case class InvalidPredicateField(message: String) extends QueryValidationError
 
-  /** Class representing unexpected error */
-  case class UnexpectedError(message: String) extends QueryValidationError
+  case class ErrorFields()
 
   /** Class representing query */
   case class Query(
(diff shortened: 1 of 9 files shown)


src/main/scala/tech/cryptonomic/conseil/util/RouteHandling.scala
```diff
@@ -40,13 +40,14 @@ trait RouteHandling {
...
     * @return Either validated query or bad request if validation failed
     */
-  protected def validateQueryOrBadRequest(fieldQuery: Query): Directive1[Query] = Directive { inner =>
-    fieldQuery.validate match {
-      case Some(value) => inner(Tuple1(value))
-      case None => complete(StatusCodes.BadRequest)
+  protected def validateQueryOrBadRequest(entity: String, fieldQuery: ApiQuery): Directive1[Query] = Directive { inner =>
+    fieldQuery.validate(entity) match {
+      case Right(value) => inner(Tuple1(value))
+      case Left(errors) => complete(StatusCodes.BadRequest -> s"Errors: ${errors.mkString(",")}")
```
ivanopagano: Did we consider returning a json content to parse even when there are errors?

ivanopagano: Possibly using `mkString` with a `\n` separator might be more readable as text

src/main/scala/tech/cryptonomic/conseil/util/RouteHandling.scala
```diff
@@ -62,6 +63,15 @@ trait RouteHandling {
     }
     complete(response)
   }
+
+  /** validates platform */
+  protected def validatePlatform(config: PlatformsConfiguration, platform: String): Directive0 = {
+    getNetworks(config, platform) match {
```
ivanopagano: Idea: would an `if/else` looking for `isEmpty` be easier to follow?

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -4,20 +4,26 @@ import com.fasterxml.jackson.core.`type`.TypeReference
...
+  import io.scalaland.chimney.dsl._
 
   /** Default value of limit parameter */
   val defaultLimitValue: Int = 10000
 
   /** Max value of limit parameter */
   val maxLimitValue: Int = 100000
 
+  /** Trait representing query validation errors */
+  sealed trait QueryValidationError {
```
ivanopagano: extending Product with Serializable might be useful later on, for type inference

src/main/scala/tech/cryptonomic/conseil/util/RouteHandling.scala
piotrkosecki: ok, I've added `\n`

src/main/scala/tech/cryptonomic/conseil/util/RouteHandling.scala
piotrkosecki: I've changed to `if/else`

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: sure, added

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -1,5 +1,7 @@
 package tech.cryptonomic.conseil.generic.chain
 
+import java.io.Serializable
```
ivanopagano: no need, there's an alias in `_root_.scala` package

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -47,8 +49,7 @@ object DataTypes {
   /** Class representing invalid predicate field */
   case class InvalidPredicateField(message: String) extends QueryValidationError
 
-  /** Class representing unexpected error */
-  case class UnexpectedError(message: String) extends QueryValidationError
+  case class ErrorFields()
```
ivanopagano: 🤔 ?

piotrkosecki: removed this ;)

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: I don't know why intellij did not see `Serializable` at that moment but you are right, there it is :)

PR #234, opened 2019-01-24
Conseil paginated block fetching

ivanopagano:
Paginated loading of blocks to keep memory under threshold, for large loads.

 * split the ranges of levels into sub-ranges and sequentially load those in memory for storage in db
 * optimize the account checkpoint loading to avoid reading the whole table

While the process might a bit slower, it now signals progress of block fetching and stores partial results, which means that a failure along the way won't compromise all the progress made so far.

## Database migration
the following index needs be added to running database for compatibility
`CREATE INDEX ix_accounts_checkpoint_block_level ON public.accounts_checkpoint USING btree (block_level);`

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -11,7 +11,7 @@ import scala.concurrent.duration._
 import scala.annotation.tailrec
 import scala.concurrent.{Await, Future}
 import scala.concurrent.duration.Duration
-import scala.util.{Failure, Success}
+import scala.util.{Failure, Success, Try}
 
 /**
   * Entry point for synchronizing data between the Tezos blockchain and the Conseil database.
@@ -116,7 +116,8 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig {
       case Custom(n) => tezosNodeOperator.getLatestBlocks(tezosConf.network, Some(n))
     }
 
-    def storeResultsPage(results: Future[tezosNodeOperator.BlockFetchingResults]): Future[Int] =
+    /* will store a single page of block results */
+    def processBlocksPage(results: Future[tezosNodeOperator.BlockFetchingResults]): Future[Int] =
       results.flatMap {
         blocksWithAccounts =>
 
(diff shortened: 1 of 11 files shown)


ivanopagano:
We need this PR merged before #232 as local testing shows that memory consumption will increase after the operation decoding changes, hence preventing a new conseil bootstrap from scratch.

src/test/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperatorTest.scala
```diff
@@ -37,16 +37,30 @@ class TezosNodeOperatorTest extends FlatSpec with MockFactory with Matchers with
...
+    val blockPages: Future[nodeOp.PaginatedBlocksResults] = nodeOp.getLatestBlocks("zeronet")
 
     //then
-    block.futureValue should have length 1
+    val (pages, total) = blockPages.futureValue
+    total shouldBe 162100
+    val results = pages.toList
+    results should have length 163
+    logger.info("First page is {}", results.head.futureValue)
+    results.head.futureValue should have length 1
```
ivanopagano: I have some issue to figure out an unexpected error during this test.
The error is the same found by the travis CI integration.

Any help is appreciated.

ivanopagano: Indeed, I agree, I made all sorts of mix-and-try this morning to see what combination could work, yet to no avail.

I have not enough expertise on how to use advanced scalamock features, or (as I fear) the framework can't possibly express the needed behaviour in a simple way

piotrkosecki: I've created basic example of a failing test, check if it represents what you wanted to so I could create an issue in the scalamock repo:
```
  class Sth {
    def aaa[A](i: Int): A = ???

    def bbb = aaa[Int](0)
    def ccc = aaa[String](0)
  }

  "X" should "Y" in {
    val sth = stub[Sth]
    (sth.aaa[Int] _).when(*).returns(1)
    (sth.aaa[String] _).when(*).returns("a")

    sth.bbb shouldBe 1
    sth.ccc shouldBe "a"
  }
```

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -160,71 +163,75 @@ class TezosNodeOperator(val node: TezosRPCInterface, batchConf: BatchFetchConfig
...
-        if (headLevel <= maxLevel) {
-          logger.info("No new blocks to fetch from the network")
-          Future.successful(List.empty)
-        } else {
-          if (maxLevel == -1) logger.warn("There were apparently no blocks in the database. Downloading the whole chain..")
-          else logger.info("Found new block head at level {}, currently stored max is {}. Fetching missing blocks", headLevel, maxLevel)
-          getBlocks(network, headHash, maxLevel + 1, headLevel)
-        }
-    } yield blocks
+  def partitionBlocksRanges(levels: Range.Inclusive, pageSize: Int = 1000): Iterator[Range.Inclusive] =
```
vishakh: Please ensure `pageSize` is read from the Lorre conf file.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -110,26 +110,42 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig {
...
-      case Custom(n) => tezosNodeOperator.getLatestBlocks(tezosConf.network, n)
+      case Everything => tezosNodeOperator.getLatestBlocks(tezosConf.network)
+      case Custom(n) => tezosNodeOperator.getLatestBlocks(tezosConf.network, Some(n))
     }
 
-    blocksToSynchronize.flatMap {
-      blocksWithAccounts =>
-        db.run(TezosDb.writeBlocksAndCheckpointAccounts(blocksWithAccounts.toMap))
-          .andThen {
+    def storeResultsPage(results: Future[tezosNodeOperator.BlockFetchingResults]): Future[Int] =
```
vishakh: I think this function name is a bit cryptic. May I suggest something like `processBlocksPage`?

vishakh:
I think this PR is good enough to merge into your primary branch. 

However, could you please describe what happens when there is a failure in block fetching with one of the pages. Will the entire Lorre process die or will we end up with a missing page of blocks in the database?

@ba205 @anonymoussprocket

src/test/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperatorTest.scala
ivanopagano: > I've created basic example of a failing test, check if it represents what you wanted to so I could create an issue in the scalamock repo:
> 
> ```
>   class Sth {
>     def aaa[A](i: Int): A = ???
> 
>     def bbb = aaa[Int](0)
>     def ccc = aaa[String](0)
>   }
> 
>   "X" should "Y" in {
>     val sth = stub[Sth]
>     (sth.aaa[Int] _).when(*).returns(1)
>     (sth.aaa[String] _).when(*).returns("a")
> 
>     sth.bbb shouldBe 1
>     sth.ccc shouldBe "a"
>     sth.bbb shouldBe 1
>     sth.ccc shouldBe "a"
>   }
> ```

is more like what we need

Thanks

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: done

ivanopagano:
> I think this PR is good enough to merge into your primary branch.
> 
> However, could you please describe what happens when there is a failure in block fetching with one of the pages. Will the entire Lorre process die or will we end up with a missing page of blocks in the database?
> 

@vishakh @ba205 @anonymoussprocket

Nice point, I didn't actually realise it would exit the VM with an exception.
I'll change the code to handle that failure gracefully, meaning that:
1. the pages already processed would be safely stored (which happens with the current code too)
2. on error, the current cycle will stop and possibly restart from where it left on the next
3. in any case, no Accounts and no Fee will be stored and computed, before all the blocks are eventually persisted

PR #239, opened 2019-01-29
Conseil 235 tweaks to conseil api v2 - part 1

piotrkosecki:
- [x] The entities route redundantly returns the network name, even though the route looks like <...network/entities...>, so the user already knows by the time they are using this route what network they are on. This should be removed.
**--> network field removed**

- [x] The "timestamp" field has a cardinality when returned by an entities query. This should be removed, because the cardinality of "timestamp" is necessarily the length of the table it is in, and will be expensive to compute.
**--> cardinality field is now optional, will return `null` for high cardinality fields**

- [ ] Right now, we parse the "script" and "params" fields from the Tezos Node using Jackson into a Scala object, and we write that to the blocks/accounts tables. For now, let's write the raw json returned by the Tezos node instead.
**--> I will add it in separate ticket/PR**

- [x] Specifying no fields in the json payload should return all columns for that table. Right now, it only returns some of them when that is the case.
- [x] Certain fields do not work for queries, such as "source" and "destination" for accounts table queries.
**--> both issues above - query was returning only non-null fields, now will return all specified columns or all from table if none of the fields are specified**

- [ ] When trying to query all unique predecessors from the blocks table, the server returned an error instead of returning the specified elements
**--> it is high cardinality field so distinct values were not returned - I can implement different handling of this if needed**


src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
```diff
@@ -668,6 +668,45 @@ class TezosDatabaseOperationsTest
...
+      val columns = List()
+
+      val populateAndTest = for {
+        _ <- Tables.Blocks ++= blocksTmp
+        found <- sut.selectWithPredicates(Tables.Blocks.baseTableRow.tableName, columns, List.empty, List.empty, 3)
+      } yield found
+
+      val result = dbHandler.run(populateAndTest.transactionally).futureValue
+      result shouldBe List(
+        Map("operations_hash" -> None, "timestamp" -> Some(new Timestamp(0)), "context" -> Some("context0"), "proto" -> Some(1),
```
ivanopagano: cosmetic: would you kindly indent these results, to help any further change here with enhanced readability?

piotrkosecki: done

PR #244, opened 2019-02-01
Surgically handle managerPubkey format mismatch within operations

ivanopagano:
* remove the previous coarse-grained solution

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/JsonDecoders.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/JsonDecoders.scala
@@ -8,7 +8,7 @@ object JsonDecoders {
 
     import io.circe.Decoder
     import cats.syntax.functor._
-    import io.circe.{ Error, Errors }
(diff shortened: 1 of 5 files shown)


ivanopagano:
Might get deprecated as only `zeronet` have such discrepancy to handle, and we will probably drop support for it

vishakh:
Closing this PR as `zeronet` is now out of scope.

ivanopagano:
@vishakh I see that actually, a [proposed tezos upgrade](https://gitlab.com/nomadic-labs/tezos/commit/50cf26f012c456248a2a01d59e654fe9993b117b) might actually align the inconsistency we need to handle here

ivanopagano:
May I suggest to re-open this PR?

Even though we might not need it, we do adapt the json, as our internal (conseil) choice of casing is anyway more consistent, and will probably be the same for a possible alphanet/mainnet upgrade.

The code outlined here does the same thing we're doing anyway, but it's a more clean and localized (in terms of code) solution.

@vishakh

PR #246, opened 2019-02-04
Make the results from latest block and block by hash consistent

ivanopagano:


vishakh:
As the `v1` API is generally deprecated and only used by Tezori in production, it's probably best not to change the behavior of the API in case it causes legacy breaks. What do you think, @anonymoussprocket ?

vishakh:
Closing as we have decided not to make further changes to the v1 API.

PR #259, opened 2019-02-15
Conseil typelevel refactor

ivanopagano:
Introduce typelevel community libraries to reduce code clutter with additional utility methods, conversions, type classes, and, in general, a more type-safe and functional approach

The most prominent refactor was to remove the `HasBalanceUpdates` custom type class to use the `Getter` optic (from monocle library), that essentially does the same.


build.sbt
```diff
@@ -6,6 +6,7 @@ val akkaVersion = "2.5.11"
 val slickVersion = "3.2.1"
 val circeVersion = "0.11.0"
 val catsVersion = "1.6.0"
+val monocleVersion = "1.6.0-M1"
```
dorians: M1? You are a brave man ;)

ivanopagano: Using the bare-minimum... will you say that the release cycle will take that long for a stable release?

dorians: I don't how fast they are with releasing new versions but M1 sounds like the early alpha for me.

ivanopagano: I changed it, you're right, I don't know why but I unwillingly read M1 as RC1

ba205:
@ivanopagano From my understanding, lens are used when you have nested records and want to access a subpart of them in a purely functional way while eliminating unnecessary boilerplate.

Right now, it doesn't seem to be adding that much value for this particular use case. What was your reasoning from going from typeclasses to lenses instead of just functions in a module?

ivanopagano:
> @ivanopagano From my understanding, lens are used when you have nested records and want to access a subpart of them in a purely functional way while eliminating unnecessary boilerplate.
> 
> Right now, it doesn't seem to be adding that much value for this particular use case. What was your reasoning from going from typeclasses to lenses instead of just functions in a module?

Apart from full-blown lenses (read and write) there are additional type classes available there that are used to "read-only", "write-only", "read with possibly no value", "read across the entire structure in a single pass", ... and so on, depending on the exact optic used.

The previous custom type class used to read balance-updates from a generic type T, was doing the same work as the `Getter` type class already available in the optics family.
For the sake of code reuse I opted for the existing type class instead of our custom one.

build.sbt
```diff
@@ -41,6 +43,8 @@ libraryDependencies  ++=  Seq(
   "com.github.alanverbner"          %% "bip39"                     % "0.1",
   "com.github.scopt"                %% "scopt"                     % "4.0.0-RC2",
   "io.scalaland"                    %% "chimney"                   % "0.3.0",
+  "com.github.julien-truffaut"      %% "monocle-core"              % monocleVersion,
+  "com.github.julien-truffaut"      %% "monocle-macro"             % monocleVersion,
```
ba205: It might worth considering, either for now or in the future, to add monocle-law, and to consider property-based testing in our codebase in general.

ba205:
@ivanopagano This is a really cool library that may be easier to get the hang of than I initially surmised. I'll approve these changes.

build.sbt
ivanopagano: yes, sounds like a good idea

PR #264, opened 2019-02-19
Conseil 263: Try to fix deadlock with slick connection pooling

### LINKED ISSUE #263: Lorre shutdown after initial blocks downloaded into Postgres
Update: After initial commit, postgres essentially works to build indices while lorre seems to stop, might need to increase number of available threads(speculative) as previously we used to run a custom config file with num_threads = 16, now we default to application.conf which only has 10.  errors posted show that the accounts table seems to crap out once postgres is done building indices and lorre starts picking up the latest blocks.

ivanopagano:
should fix #263 
need to test a whole chain reset on Lorre in dev

 * put lockstep configs in slick's connection
 * upgrade to slick's latest, that should contain a fix for the known
   issue

The issue being referred is dnvriend/akka-persistence-jdbc#177

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,9 @@ conseildb = {
     user = "redacted"
     password = "redacted"
   }
-  numThreads = 10 #please make sure you know what you're doing before changing this
+  #please make sure you know what you're doing before changing these values
+  numThreads = 10
+  maxConnections = 10
 }
 
 #available blockchain platforms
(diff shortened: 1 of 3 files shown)


ivanopagano:
No guarantee of being the fix for database locking down after the block's reload

It's based on error logs and known issues that came out after some quick investigation

PR #265, opened 2019-02-21
Conseil 168 adding endpoints library to metadata and query routes

piotrkosecki:
PR contains changes from https://github.com/Cryptonomic/Conseil/pull/252 and additionally implementation of the metadata endpoints.


piotrkosecki:
created new PR with more changes

PR #267, opened 2019-02-25
Conseil 253: add governance voting data to blocks

### LINKED ISSUE #253: Governance Tables
We want to keep track of Tezos's governance process in our data model. In order to do this, we will create three tables, "proposals", "bakers", and "ballots", as well as add fields to the blocks table.

The schema is attached, and all types and foreign key information will be in there, as well as information about what RPC calls to make in order to obtain the data required. Some notes:

Proposals: Will have the fields "protocol_hash", "block_hash", and "block_level", so we know what proposal is currently be considered during the pseudo "timestamp" that is identified by "block_hash" and "block_level". 

Bakers: Will have fields "pkh", "rolls", "block_hash", "block_level", so we know the account ids of bakers and how many rolls they have. 1 roll = 10k tz, is proportional to the amount of "voting power" the baker has. This is also pseudo "timestamped" by "block_hash" and "block_level". 

Ballots: Will have fields "pkh", "ballot, "block_hash", "block_level", so we know what votes a given user with account id "pkh" votes on a specific proposal. Since this is also pseudo "timestamped" by "block_hash" and "block_level", we can join into the blocks table to see what specific proposal this is for.

BlocksAdd:  We will add the fields "period_kind", "current_expected_quorum", and "active_proposal" to know what proposal is currently under consideration, if any, what section of the proposal review process we are currently in at the time of block creation, and the current_expected_quorum necessary to accept this proposal. The reason for current_expected_quorum is this value changes dynamically as more and more proposals are considered.

ivanopagano:
partially fix #253 

 * adds voting data to each block
 * strict encoding in scala types of the voting data
 * converts data to db columns
 * refactor to handle multiple json requests with decoding in a more uniform way as to not clutter the code to read blocks

- add the following _columns_ to the `blocks` table
  ```
    period_kind character varying,
    current_expected_quorum integer,
    active_proposal character varying
  ```

## Required changes to configuration files
- none


ivanopagano:
Currently marked as `draft` as one test is still not correctly passing

src/main/scala/tech/cryptonomic/conseil/generic/chain/MultiFetchDecoding.scala
```diff
@@ -0,0 +1,136 @@
...
+  /** an effectful function that decodes the json value to an output `Out` */
+  def decodeData: Kleisli[F, Encoded, Out]
+
+  /** using the combination of the provided functions we fetch all data with this general outline:
+    * 1. get json values from the node, given a traversable input collection
+    * 2. keep the correlation with the input and convert every json to a corresponding output
+    * 3. return the traversable collection containing paired input and output values, all wrapped in the effect F
+    */
+  def fetch(implicit app: Monad[F], tr: Traverse[T]): Kleisli[F, T[In], T[(In, Out)]] =
+    fetchBatch.andThen(json => decodeData.second[In].traverse(json))
```
vishakh: For my education..since so much set up was done with Kleislis to potentially set up a neat composition, how come there is no 'higher-order Kleisli' that is able to operate on a traversible of effects so this line simply becomes something like:

`fetchBatch.andThenTraverse(decodeData..)`

?

src/main/scala/tech/cryptonomic/conseil/generic/chain/MultiFetchDecoding.scala
```diff
@@ -0,0 +1,136 @@
...
+  private type Aux[F[_], T[_], I, O, E] = MultiFetchDecoding[F, T] {
+      type In = I
+      type Out = O
+      type Encoded = E
+    }
+
+  /** Adds an extra decoding operation for every intermediate encoded result, returning both the
+    * original output `O` and the new decoded output `O2` as a tuple
+    */
+  def addDecoding[F[_]: Apply, T[_], I, O, O2, E](
```
vishakh: For this function and the ones below, could you add some concrete examples of how they would be useful for fetching blockchain data?

src/main/scala/tech/cryptonomic/conseil/generic/chain/MultiFetchDecoding.scala
ivanopagano: I would say that they are orthogonal concerns, composition of functions (andThen) wrt traversing a structure.
Since the result of the kleisli application (i.e. the result of a function with effect) might have a very general form, you could apply many possible computations to that.
e.g. it could be optional, monoidal, monadic, applicative, alternative, ... each of those type classes provides many operations, on the "content" of the result, and providing custom composition for the kleisli own type classes (arrow, alternative, functor, contravariant-functor, ...) with those of the content would lead to an explosion of custom operators.
This would also be limiting, as new type classes will be defined, and no ad-hoc combination would be available for those.

If you look at `Future.sequence` and `Future.traverse`, that is an ad-hoc commodity feature, that doesn't scale much, because after a while you start asking yourself why only those and why only on Future and not, say Option...
This is my own explanation, the fact that you want to keep concerns separated without tying up separate type classes.

As a side-note, a different reasoning applies if having an additional constraint on the, say, result of a kleisli operation, might allow you to create powerful shortcuts, like having a monoid/semigroup (that is, you can combine many results of the same type to a single value).
This is used if you got a collection of monoid things, that you can immediately reduce to a single value.
Possibly the andThenTraverse combinator is not so common to prove useful yet.

src/main/scala/tech/cryptonomic/conseil/generic/chain/MultiFetchDecoding.scala
ivanopagano: This is a wonderful suggestion, I'll add examples to all cases

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -26,12 +26,15 @@ object TezosNodeOperator {
...
 
+  val isGenesis = (data: BlockData) => data.header.level == 0
+
 }
 
 /**
   * Operations run against Tezos nodes, mainly used for collecting chain data for later entry into a database.
   */
 class TezosNodeOperator(val node: TezosRPCInterface, batchConf: BatchFetchConfiguration)(implicit executionContext: ExecutionContext) extends LazyLogging {
+  import TezosNodeOperator.isGenesis
```
ba205: Is the reason you define isGenesis out of the scope of TezosNodeOperator because it's not intrinsic to interacting with the Tezos Node?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeMultiFetch.scala
```diff
@@ -0,0 +1,164 @@
...
+  type In
+  /** the output type*/
+  type Out
+  /** the json representation type used*/
+  type JsonType
+
+  /** an effectful function from a collection of inputs `T[In]`
+    * to the collection of json, tupled with the corresponding input `T[(In, JsonType)]`
+    */
+  def fetchJsonBatch: Kleisli[F, T[In], T[(In, JsonType)]]
```
ba205: So this is equivalent to T[In] => F[T[(In, JsonType)]]. In is just the raw input type that we collect from the Tezos node, and JsonType is generic for the way we choose to parse/represent the data?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeMultiFetch.scala
```diff
@@ -0,0 +1,164 @@
...
+
+  /** an effectful function that decodes the json value to an output `Out` */
+  def decodeJson: Kleisli[F, JsonType, Out]
```
ba205: This is ambiguous to me. If In is the raw input type from the Tezos Node, and Json type is the representation of the parsing, what would an example of Out be?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeMultiFetch.scala
```diff
@@ -0,0 +1,164 @@
...
+    */
+  object Syntax {
+    implicit class MultiFetchOps[F[_], T[_], I, O, J](m: Aux[F, T, I, O, J]) {
+      def decodeAlso[O2](additionalDecode: Kleisli[F, J, O2])(implicit app: Apply[F]): Aux[F, T, I, (O, O2), J] = addDecoding(m, additionalDecode)
+    }
+  }
+
+}
+
+/** Defines intances of `TezosNodeMultiFetch` for block-related data */
```
ba205: Typo in description. Also, maybe it would make more sense to split up the interface from the actual implementation? In general, the Tezos folder is getting bloated. One option is to have sub folders for database operations, api operations, node operations, and abstractions/interfaces, and place concrete implementations of abstractions in the folder in which those concrete implementations are used.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeMultiFetch.scala
```diff
@@ -0,0 +1,164 @@
...
+    network: String,
+    node: TezosRPCInterface,
+    hashRef: BlockHash,
+    concurrency: Int
+  ) = new TezosNodeMultiFetch[Future, List] {
+    import JsonDecoders.Circe.Blocks._
+
+    type JsonType = String
+    type In = Int //offset from head
+    type Out = BlockData
```
ba205: Oh, okay, so In is any additional information needed to compute with the incoming tezos node data, JsonType is the type of the incoming data, and output is the resultant objects we desire from the computation?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeMultiFetch.scala
```diff
@@ -71,6 +73,21 @@ object TezosNodeMultiFetch {
     def decodeJson = multiFetch.decodeJson.product(additionalDecode)
   }
 
+  /** Combines two multifetch acting on the same inputs to get a pair of the outputs,
+   * considering that the intermediate json should not be the same, as required for `addDecoding`
+   * TODO consider composing with a different function that (,) as to combine more than one with different out shapes
+   */
+  def tupleResults[F[_]: Monad, T[_] : Traverse: FunctorFilter, I, O1, O2, J1, J2](mf1: Aux[F, T, I, O1, J1], mf2: Aux[F, T, I, O2, J2]): Kleisli[F, T[I], T[(I, (O1, O2))]] =
```
ba205: What use case will we need an input type to be mapped to two output types for a single instance of Json?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -38,7 +38,7 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig {
   sys.addShutdownHook(shutdown())
 
   lazy val db = DatabaseUtil.db
-  val tezosNodeOperator = new TezosNodeOperator(new TezosNodeInterface(tezosConf, callsConf, streamingClientConf), batchingConf)
+  val tezosNodeOperator = new TezosNodeOperator(new TezosNodeInterface(tezosConf, callsConf, streamingClientConf), tezosConf.network, batchingConf)
```
ba205: "for adding governance fields to the block table, this change is extraneous. It would have been better to open a ticket and do a separate pr, so we have more traceability into changes to conseil"

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ivanopagano: Just because it's a generic function that can be reused, and I want to define it only in a single place

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeMultiFetch.scala
ivanopagano: Mostly, I've changed the names actually, because we need not be tied to json.

Apart from that yes, the function type is that you said, and the In represent the input on which you call the rest apis (e.g. block offsets or hashes)...

The JsonType is now a more generally defined `Encoded` type, and the decoder is generic from `Encoded` to `Out`

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeMultiFetch.scala
ivanopagano: `In` is our actual input needed to call a single tezos endpoint (e.g. the offset from a specific block, when you call the `blocks` endpoint)

the `JsonType` now renamed, is the representation used for json, in this case `JsonType = String`, very simply

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeMultiFetch.scala
ivanopagano: I did it, I think we mentioned this already somewhere

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeMultiFetch.scala
ivanopagano: yep

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeMultiFetch.scala
ivanopagano: I've added documentation with examples that should clarify why I needed these combinators

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: We discussed this already, you're totally right, but it's not easy to make the right call when a big refactoring is in the working.

I will make a better effort of splitting smaller PRs in the future.
Still is not so easy to do that properly, as
 - it's not always clear during the development, how the final solution will look like, before trying to use it practically
 - doing a review on a "preparation refactor" is not easy if the changes have no clear application

That said, I will anyway make the effeot to improve our process,
thank you

src/main/scala/tech/cryptonomic/conseil/generic/chain/MultiFetchDecoding.scala
```diff
@@ -0,0 +1,209 @@
...
+  * of the operation execution, and represents everything parametrically, to leave
+  * room for future evolutions of the code protocol.
+  * Thus the required implementation is in the form of a `cats.data.Kleisli` type,
+  * parametric on effect `F`, input `In`, output `Out`. As needed, the in/out types are
+  * wrapped in a `T[_]` type constructor that represents multiple values (e.g. a collection type).
+  *
+  * @tparam F the higher-order effect type (e.g. `Try`, `Future`, `IO`, `Either`)
+  * @tparam T a traversable type (used to wrap mutliple requests and corresponding return values)
+  */
+trait MultiFetchDecoding[F[_], T[_]] {
```
vishakh: As discussed on Riot, it would be good to rename the type parameters so they are more descriptive and the code is easy to read.

vishakh: I think the name `MultiFetchDecoding` is too abstract and will scare people away. Can we give it a name which is more friendly and describes is a simple way what this typeclass does?

src/main/scala/tech/cryptonomic/conseil/generic/chain/MultiFetchDecoding.scala
```diff
@@ -0,0 +1,209 @@
...
+  import cats.syntax.foldable._
+  import cats.syntax.functorFilter._
+
+  /* An alias used to define function constraints on the internal dependent types (i.e. `I`, `O`, `E`)
+   * which would be otherwise opaque in any function signature
+   *
+   * Example: the fetcher for tezos blocks might have type `Aux[Future, List, Int, BlockData, String]
+   * where the Int is for the offset from a given block, and String is a representation of a json value.
+   */
+  private type Aux[F[_], T[_], I, O, E] = MultiFetchDecoding[F, T] {
```
vishakh: It would be great to rename `Aux` as well so the code is more friendly to read.

src/main/scala/tech/cryptonomic/conseil/generic/chain/MultiFetchDecoding.scala
vishakh: For my education, could you please describe why you chose to create a typeclass instead of a regular trait? 

When I look at this in the most minimal manner possible, I can certainly see the point of the following:
- Defining a trait with common workflow functions will unite all blockchain implementations.
- Using Kleisli to make the composition of the workflow functions cleaner to read.

However, beyond this, I am not able to see the additional advantage of defining an entire typeclass. Generally, as I understand it, typeclasses allow one to change the behaviour of categories of types without having to change their implementations, e.g. defining `toJSON()` functionality for multiple types like strings, integers, monkeys, bananas, etc. In this case, however, we are not enhancing the behaviour of any existing types but simply defining a few new types for various workflows. If the following were done using typeclasses, it would make intuitive sense to me:

```scala
val levels = Set(773, 774, 775, 776, 777)
val blockData = levels.fetchTezosBlocks()
```

Could you please explain a bit further the additional utility of implementing a typeclass in this specific way? :smile:

doc/conseil.sql
```diff
@@ -86,7 +86,10 @@ CREATE TABLE public.blocks (
     protocol character varying NOT NULL,
     chain_id character varying,
     hash character varying NOT NULL,
-    operations_hash character varying
+    operations_hash character varying,
+    period_kind character varying,
+    current_expected_quorum integer,
+    active_proposal character varying
```
vishakh: Should we not see the new governance tables here?

ivanopagano: next PR on top of this one, will publish _very_ soon for review

src/main/scala/tech/cryptonomic/conseil/generic/chain/MultiFetchDecoding.scala
ivanopagano: In this case we're only using an interface... to be a type class we should have defined implicit instances for the custom types, to use as constraints in other method calls.

In fact here the usage pattern is that of a traditioal OO interface. This is standard information hiding used to provide common higher-level interaction with tezos nodes batched interaction and json parsing, with some reuse.

It allows paves the road to a potential switch to other async models that we can explore soon.
Examples might be 
- using Akka-Streams by changing Future[List[Block]] to Source[Id[Block]] (which is actually `Source[Block]`)
- using monix/cats IO changing Future[List[Block]] to ??? I don't know exactly yet
- using zio using ZIO.Task[ZIO.Stream[Block]] or something alike

src/main/scala/tech/cryptonomic/conseil/generic/chain/MultiFetchDecoding.scala
ivanopagano: I have no trouble with that, but renaming it will lose connection to this technique, which is commonly referred to as the "Aux Pattern"

src/main/scala/tech/cryptonomic/conseil/generic/chain/MultiFetchDecoding.scala
ivanopagano: I will think of something friendlier, but, as I mentioned later on, this is not technically a type class.

ivanopagano: Sure, will do

src/main/scala/tech/cryptonomic/conseil/generic/chain/MultiFetchDecoding.scala
ba205: Are the following normal conventions? F for effects, T for containers, I for input, O for output, E for some intermediary representation? If yes, no worries, but some renames might be worth considering, if we are keeping Aux to represent "Aux Pattern"

src/main/scala/tech/cryptonomic/conseil/generic/chain/MultiFetchDecoding.scala
ba205: So you're saying the advantage here, if I'm understanding correctly, is to generalize over the effects that various workflows over blockchain endpoints may require?

src/main/scala/tech/cryptonomic/conseil/generic/chain/MultiFetchDecoding.scala
ivanopagano: only Aux and F are common, the rest I picked up based on reference to the full word, I input O output E encoded T traversable

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -32,61 +32,81 @@ object TezosNodeOperator {
...
+  val isGenesis = (data: BlockData) => data.header.level == 0
+
+  /** Helper to decode json and convert to a possibly failing future result
+    * This is not running any async operation
+    */
+  def decodeToFuture[A: io.circe.Decoder](json: String) = {
+    import io.circe.parser.decode
+
+    decode[A](json) match {
+      case Left(error) => Future.failed(error)
```
vishakh: We should log the failing JSON string to console. Maybe you can do this in the subsequent PR?

ivanopagano: I'm actually considering how to "percolate" all relevant failure information to the top level, even from multiple threads.
The idea is to have a way to get detailed information on each step failure from the bottom up, traced with some id, such that we can rebuild some narrative of the event through all the layers.

PR #270, opened 2019-03-01
Lorre won't quit if it didn't have chance to connect to Tezos node

piotrkosecki:
So the the current behaviour will be like this:
- if Lorre didn't connect to Tezos node yet, it will keep trying
- if Lorre already connected  and error occurs, Lorre will stop


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -70,7 +70,7 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig {
      * Otherwise, any error will make Lorre proceed as expected by default (stop or wait for next cycle)
      */
     val attemptedProcessing =
-      if (sys.env.get("LORRE_FAILURE_IGNORE").forall(ignore => ignore == "false" || ignore == "no"))
+      if (sys.env.get("LORRE_FAILURE_IGNORE").forall(ignore => ignore == "false" || ignore == "no") || didConnect)
```
ivanopagano: I'm not sure this does what it's needed...
This flag should control wether Lorre stops when processing is failed with one of 
`AccountsProcessingFailed` or `BlocksProcessingFailed`, or if they should be actually ignored...
But this branch of flow should not be the one commonly used anyway... hence probably the connection failures pass through anyway and will stop the process even when didConnect is false

vishakh: Besides what @ivanopagano said, it seems to me the 'boot up' behavior is distinct from the 'run time' behavior, i.e. it might make sense to have a distinct function which is executed at Lorre start up which 'blocks' until the Tezos host is contactable.

ivanopagano: @vishakh: you mean some kind of pre-start sequence, explicitly denoted separately from the main loop?

piotrkosecki: @ivanopagano I've tested it - behaviour I've implemented do not fail lorre until it successfully finishes processing, connection failures do not cause lorre to fail. 

But from what @vishakh said I need to implement it in some way that checks connection to the tezos node. And here I need your opinion on how to do it. I've looked into the documentation https://tezos.gitlab.io/alphanet/api/rpc.html#rpc-index and I'm not sure on which one I can rely that gives me current status of the node - if it is possible to be queried. I could go for the `blocks`, but it would be doubling the the query. Or do we assume that if any endpoint works every other works as well?

piotrkosecki: Ok, I've changed the implementation to run blocking check for blocks head before the main loop recursively so until connection to the tezos node will be established, the `mainLoop` won't start. Added new config value for retry interval.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -103,7 +103,27 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig {
...
   )
 
-  try {mainLoop(0)} finally {shutdown()}
+  /**
+    * Tries to fetch blocks head to verify if connection with Tezos node was successfully established
+    */
+  @tailrec
+  def checkConnection(): Unit = {
+    Try {
+      Await.result(tezosNodeOperator.getBlockHead(tezosConf.network), Duration.Inf)
```
ivanopagano: I would put a limited duration as a timeout, as we actually want to check repeatedly if no timely response comes back

piotrkosecki: it also times out with `Duratinion.Inf`, but the question is, do you have any specific value in mind?

piotrkosecki: after consulting with @ivanopagano added 10 seconds timeout in the Lorre config

src/main/resources/application.conf
```diff
@@ -9,6 +9,8 @@ conseil: {
 # Runtime settings for Lorre process
 lorre: {
   sleepInterval: 60 s
+  retryInterval: 10 s
```
vishakh: Can we give these more specific names like `bootupRetryInterval` and `bootupConnectionCheckTimeout`?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -103,7 +103,27 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig {
     tezosConf.network
   )
 
-  try {mainLoop(0)} finally {shutdown()}
+  /**
+    * Tries to fetch blocks head to verify if connection with Tezos node was successfully established
+    */
+  @tailrec
+  def checkConnection(): Unit = {
```
vishakh: Can we give this a more specific name like `checkTezosConnection`?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -103,7 +103,27 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig {
...
+    Try {
+      Await.result(tezosNodeOperator.getBlockHead(tezosConf.network), lorreConf.connectionCheckTimeout)
+    } match {
+      case Failure(e) =>
+        logger.error("Could not connect to the Tezos node", e)
```
vishakh: How about "Could not make initial connection to Tezos"?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -103,7 +103,27 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig {
...
+        Thread.sleep(lorreConf.retryInterval.toMillis)
+        checkConnection()
+      case Success(_) =>
+        logger.info("Successfully connected to the Tezos node")
```
vishakh: How about "Successfully made initial connection to Tezos"?

piotrkosecki: changed according to suggestion

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: changed according to suggestion

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: changed

src/main/resources/application.conf
piotrkosecki: ok, changed

PR #273, opened 2019-03-08
Conseil improve deploy process

ivanopagano:
Enhance the deployment process by passing extra information on the console at startup, under an additional command line flag `-v` or `--verbose`

Using this will display relevant information to understand configuration of
- database connections
- tezos nodes
- available chain platforms

This PR additionally converts the default `application.conf` file to a `reference.conf`.
Briefly explained, this makes all configuration entries therein visible to the application, even if not explicitly "included" by a custom file.
Additional information is available on [lightbend-config docs](https://github.com/lightbend/config#standard-behavior)

Therefore:

**Important change**
a custom `-Dconfig.file=conf-file` will henceforth *only need to override values*, while the existing entries will *be visible anyway*.
Usage of the `--verbose` flag will be critical to debug potential conflicts in configuration and to understand which valuse are actually used.


vishakh:
The coveralls check is failing. :(

ivanopagano:
> The coveralls check is failing. :(

I'll work to fix this

src/main/scala/tech/cryptonomic/conseil/MainOutputs.scala
```diff
@@ -0,0 +1,169 @@
+package tech.cryptonomic.conseil
```
vishakh: Could we put this file into a specific package? I think the codebase looks much cleaner when only the entrypoints are in the root package.

ivanopagano: I couldn't come out with a valid package idea, I have now

PR #274, opened 2019-03-08
Conseil 254 adding CSV output to query endpoint

piotrkosecki:
~~currently query endpoint supports CSV output with the `Accept: text/plain` header or with wildcard `text/*` or without any `Accept:` header. The issue probably will need to be resolved within the `endpoints` library. Consulted with @anonymoussprocket and it is ok for now.~~
I was wrong, I fixed issue with `text/csv` and it works correctly.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
@@ -12,7 +12,6 @@ import tech.cryptonomic.conseil.tezos.TezosPlatformDiscoveryOperations
   * Classes used for deserializing query.
   */
 object DataTypes {
-
   import io.scalaland.chimney.dsl._
 
   /** Type representing Map[String, Any] */
@@ -104,7 +103,6 @@ object DataTypes {
     val json, csv = Value
   }
 
-
   /** Enumeration for order direction */
   object OrderDirection extends Enumeration {
     type OrderDirection = Value
@@ -116,5 +114,4 @@ object DataTypes {
     type OperationType = Value
     val in, between, like, lt, gt, eq, startsWith, endsWith, before, after, isnull = Value
   }
-
 }
--- a/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.core.`type`.TypeReference
 import com.fasterxml.jackson.module.scala.JsonScalaEnumeration
 import tech.cryptonomic.conseil.generic.chain.DataTypes.OperationType.OperationType
 import tech.cryptonomic.conseil.generic.chain.DataTypes.OrderDirection.OrderDirection
+import tech.cryptonomic.conseil.generic.chain.DataTypes.OutputType.OutputType
 import tech.cryptonomic.conseil.tezos.TezosPlatformDiscoveryOperations
 
 
@@ -33,6 +34,9 @@ object DataTypes {
   /** Class required for OperationType enum serialization */
   class OperationTypeRef extends TypeReference[OperationType.type]
 
+  /** Class required for OutputType enum serialization */
+  class OutputTypeRef extends TypeReference[OutputType.type]
+
   /** Class representing predicate */
   case class Predicate(
     field: String,
@@ -45,6 +49,7 @@ object DataTypes {
   /** Class required for Ordering enum serialization */
   class QueryOrderingRef extends TypeReference[OrderDirection.type]
 
+  /** Class representing representing query ordering */
   case class QueryOrdering(field: String, @JsonScalaEnumeration(classOf[QueryOrderingRef]) direction: OrderDirection)
 
   /** Class representing invalid query field */
@@ -58,7 +63,8 @@ object DataTypes {
     fields: List[String] = List.empty,
     predicates: List[Predicate] = List.empty,
     orderBy: List[QueryOrdering] = List.empty,
-    limit: Int = defaultLimitValue
+    limit: Int = defaultLimitValue,
+    output: OutputType = OutputType.json
   )
 
   /** Class representing query got through the REST API */
@@ -66,7 +72,8 @@ object DataTypes {
     fields: Option[List[String]],
     predicates: Option[List[Predicate]],
     orderBy: Option[List[QueryOrdering]],
-    limit: Option[Int]
+    limit: Option[Int],
+    @JsonScalaEnumeration(classOf[OutputTypeRef]) output: Option[OutputType]
   ) {
     /** Method which validates query fields, as jackson runs on top of runtime reflection so NPE can happen if fields are missing */
     def validate(entity: String): Either[List[QueryValidationError], Query] = {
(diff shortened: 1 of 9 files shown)


src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -45,6 +50,7 @@ object DataTypes {
   /** Class required for Ordering enum serialization */
   class QueryOrderingRef extends TypeReference[OrderDirection.type]
 
+  /** Class representing representing query ordering */
```
ivanopagano: typo

src/main/scala/tech/cryptonomic/conseil/routes/DataHelpers.scala
```diff
@@ -20,11 +20,26 @@ trait DataHelpers extends QueryStringListsServer with Validation with akkahttp.s
...
   /** Function validating request for the query endpoint */
-  override def validated[A](response: A => Route, invalidDocs: Documentation): Either[List[QueryValidationError], A] => Route = {
+  override def validated(response: QueryResponseWithOutput => Route, invalidDocs: Documentation): Either[List[QueryValidationError], QueryResponseWithOutput] => Route = {
     case Left(errors) =>
       complete(StatusCodes.BadRequest -> s"Errors: \n${errors.mkString("\n")}")
-    case Right(success) =>
+    case Right(success@QueryResponseWithOutput(_, OutputType.json)) =>
       response(success)
+    case Right(QueryResponseWithOutput(queryResponse, OutputType.csv)) =>
+      complete(StatusCodes.OK -> makeCsv(queryResponse))
```
ivanopagano: Is there a way to make this choice implicit in the way the response is marshsalled?

To avoid explicitly duplicate the response with different calls (response(success) vs. complete(...))?

Also, what about the response mime-type? Shouldn't it appear somewhere here?

src/main/scala/tech/cryptonomic/conseil/routes/DataHelpers.scala
```diff
@@ -20,11 +20,26 @@ trait DataHelpers extends QueryStringListsServer with Validation with akkahttp.s
...
+  }
+
+  def makeCsv(queryResponse: List[QueryResponse]): String = {
```
ivanopagano: Idea: what about re-using the Conversion type class we have to handout the work of transforming the map to  csv in a separate, independent place and context?

ivanopagano: In any case, I would try personally to make some effort to define the csv encoding separately instead of "inlining" that here in the route code

src/main/scala/tech/cryptonomic/conseil/routes/openapi/DataJsonSchemas.scala
```diff
@@ -51,6 +56,9 @@ trait DataJsonSchemas extends algebra.JsonSchemas with generic.JsonSchemas   {
   /** Query response schema schema */
   implicit def queryResponseSchema: JsonSchema[List[QueryResponse]]
 
+  /** Query response schema schema */
+  implicit def queryResponseSchemaWithCsv: JsonSchema[QueryResponseWithOutput]
```
ivanopagano: maybe a better naming? 😁

src/main/scala/tech/cryptonomic/conseil/routes/openapi/OpenApiDoc.scala
```diff
@@ -52,14 +52,21 @@ object OpenApiDoc extends DataEndpoints
...
-  override def validated[A](response: List[OpenApiDoc.DocumentedResponse], invalidDocs: Documentation): List[OpenApiDoc.DocumentedResponse] =
+    * */
+  override def validated(response: List[OpenApiDoc.DocumentedResponse], invalidDocs: Documentation): List[OpenApiDoc.DocumentedResponse] =
     response :+ OpenApiDoc.DocumentedResponse(
       status = 400,
       documentation = invalidDocs.getOrElse(""),
       content = Map(
         "application/json" -> MediaType(schema = Some(Schema.Array(Schema.simpleString)))
       )
+    ) :+ OpenApiDoc.DocumentedResponse(
```
ivanopagano: I'm a bit lost here, is there an easy way to understand what's happening and when each response will be given?

src/main/scala/tech/cryptonomic/conseil/routes/DataHelpers.scala
```diff
@@ -20,11 +20,26 @@ trait DataHelpers extends QueryStringListsServer with Validation with akkahttp.s
   import io.circe.syntax._
 
   /** Function validating request for the query endpoint */
-  override def validated[A](response: A => Route, invalidDocs: Documentation): Either[List[QueryValidationError], A] => Route = {
+  override def validated(response: QueryResponseWithOutput => Route, invalidDocs: Documentation): Either[List[QueryValidationError], QueryResponseWithOutput] => Route = {
```
ivanopagano: Why are we foregoing parametricity to use a concrete type?

Wouldn't pattern matching allow to be specific all the same?

src/main/scala/tech/cryptonomic/conseil/routes/openapi/OpenApiDoc.scala
piotrkosecki: basically its definition for the OpenApi(and swagger) -> it describes what is expected by each response. It is easier to understand how it works using swagger, when you can chose `accept` header for the request and it shows curl request used.

src/main/scala/tech/cryptonomic/conseil/routes/DataHelpers.scala
piotrkosecki: pattern matching was tricky in this case, but I've managed to find workaround - apparently I couldn't match `case Right(success@QueryResponseWithOutput(_, OutputType.json))` and return `response(success)` - compiler told me that the type is wrong. But when I've moved it as a last case and made it `case Right(success@QueryResponseWithOutput)` it compiles correctly.

src/main/scala/tech/cryptonomic/conseil/routes/DataHelpers.scala
piotrkosecki: you are absolutely right, I haven't thought of that. Implemented your way.

src/main/scala/tech/cryptonomic/conseil/routes/DataHelpers.scala
piotrkosecki: Yes it was wrong, I think I needed to take a step back during the weekend and think about it once again. Now it correctly returns `text/csv`!

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: yep, found few more, and fixed ;)

src/main/scala/tech/cryptonomic/conseil/routes/openapi/DataJsonSchemas.scala
piotrkosecki: changed it to `queryResponseSchemaWithOutputType`, maybe you have other suggestions? 😄

src/main/scala/tech/cryptonomic/conseil/routes/DataHelpers.scala
piotrkosecki: And regarding explicitly choosing the way the responses are marshalled -> I could't find a reasonable way to make this choice implicit.

Issue #256, opened 2019-02-14
Save script and parameters in Michelson format in database


vishakh:
The following database columns store Michelson scripts in JSON format:
- `script` in accounts
-  `script` in operations
- `parameters` in operations

Using the new JSON-to-Michelson functionality, let us ensure Lorre saves the contents of the above columns as Michelson strings instead.

PR #277, opened 2019-03-12
#256 Michelson integration - errors fixed

dorians:


src/main/scala/tech/cryptonomic/conseil/tezos/michelson/dto/MichelsonExpression.scala
```diff
@@ -25,7 +25,7 @@ sealed trait MichelsonExpression extends MichelsonElement
 case class MichelsonType(prim: String, args: List[MichelsonExpression] = List.empty) extends MichelsonExpression
 
 /* Class representing an int constant */
-case class MichelsonIntConstant(int: Int) extends MichelsonExpression
+case class MichelsonIntConstant(int: Long) extends MichelsonExpression
```
piotrkosecki: just a thought about naming, you changed `Int` to `Long`, but name of the case class stayed the same, maybe it is worth to change it to something that does not imply specific scala type like that, to something like `MichelsonNumericConstant`, `MichelsonIntegerConstant`? WDYT?

dorians: The problem is that `int` is the name of a field which contains long value encodes as a string ;)
That's what it looks like in Michelson and I agree that it's confusing. I'll look at this again tomorrow with a fresh head.

piotrkosecki: So in this case it makes more sense 👍

src/main/scala/tech/cryptonomic/conseil/tezos/michelson/renderer/MichelsonRenderer.scala
```diff
@@ -7,15 +7,20 @@ object MichelsonRenderer {
...
+      case MichelsonComplexInstruction(prim, args) => s"$prim ${args.map(_.render()).mkString(" ")}"
       case MichelsonInstructionSequence(args) => s"{ ${args.map(_.render()).mkString(" ; ")} }"
+      case MichelsonEmptyInstruction => "{}"
 
+      // expressions
       case MichelsonType(name, List()) => name
       case MichelsonType(name, args) => s"($name ${args.map(_.render()).mkString(" ")})"
       case MichelsonIntConstant(constant) => constant.toString
-      case MichelsonStringConstant(constant) => constant
+      case MichelsonStringConstant(constant) => "\"%s\"".format(constant)
```
ivanopagano: You can alternatively use "raw strings" when you need to embed quoted values, as in
`s""""$constant""""`

It's debatable which is more easy to read, yet I argue that the latter is harder to write incorrectly, because the compiler might catch non-matching quotes, or an incorrect interpolation...
With the `"...".format()` version, you'll find most errors at runtime

src/test/scala/tech/cryptonomic/conseil/tezos/michelson/parser/JsonParserSpec.scala
```diff
@@ -90,8 +90,8 @@ class JsonParserSpec extends FlatSpec with Matchers {
 
     parse[MichelsonInstruction](json) should equal(Right(
       MichelsonInstructionSequence(List(
-        MichelsonComplexInstruction("DIP", MichelsonInstructionSequence(List(
-          MichelsonSimpleInstruction("DUP"))))))))
+        MichelsonComplexInstruction("DIP", List(MichelsonInstructionSequence(List(
```
ivanopagano: Trivial comment: using indentation and formatting might make it easier to read such nesting of Lists.

I don't know if using the `a :: b :: Nil` format might make it all more parseable to the reader

src/test/scala/tech/cryptonomic/conseil/tezos/michelson/parser/JsonParserSpec.scala
```diff
@@ -90,8 +90,8 @@ class JsonParserSpec extends FlatSpec with Matchers {
 
     parse[MichelsonInstruction](json) should equal(Right(
```
ivanopagano: If you prefer, scalatest will provide with combinators to extract `Right` using a DSL

```scala
parse[MichelsonInstruction](json).right should equal (...)
```

src/test/scala/tech/cryptonomic/conseil/tezos/michelson/parser/JsonParserSpec.scala
ivanopagano: The alternatives would look something like
```scala
MichelsonInstructionSequence(
  List(
    MichelsonComplexInstruction("DIP", 
      List(
        MichelsonInstructionSequence(
          List(
            MichelsonSimpleInstruction("DUP")
          )
        )
      )
    )
  )
)
//vs.
MichelsonInstructionSequence(
  MichelsonComplexInstruction("DIP",
    MichelsonInstructionSequence(
      MichelsonSimpleInstruction("DUP") :: Nil
    ) :: Nil
  ) :: Nil
)
```

dorians: I'm refactoring this structure now to be able to add support for LAMBDA instruction and I'll need to change tests anyway so I'll try with ``a :: b :: Nil`` syntax. It might look better indeed.

PR #278, opened 2019-03-12
Conseil 263: paged accounts processing

ivanopagano:
partially improve on #263 

- Paginate the account's fetch&store operation on Lorre, in the same fashion as we do for blocks.
- Add extra logging information and progress updates


ivanopagano:
Since we have several open PRs and some address test coverage too, I'll wait for them to be merged in `master` before aligning this one

PR #284, opened 2019-03-20
Conseil json decoding failure handling

ivanopagano:
I made a few changes to make usage of circe decoding into Future results more regular.

This includes usage in the DataFetchers and allows more regular handling of local errors (during decoding)
The change explicitly adds constraints to the `Eff` in fetchers, so that fetching might include error handling via type class functions.

This is not related to any specific issue, but it's targeted toward more complete logging of failures to trace application unexpected shutdowns

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataFetcher.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataFetcher.scala
@@ -16,8 +16,9 @@ import cats.syntax.semigroupal._
   *
   * @tparam Eff the higher-order effect type (e.g. `Try`, `Future`, `IO`, `Either`)
   * @tparam Coll a traversable type (used to wrap mutliple requests and corresponding return values)
+  * @tparam Err the expected type of possible errors that the Eff-wrapped operations might raise
   */
-trait DataFetcher[Eff[_], Coll[_]] {
+trait DataFetcher[Eff[_], Coll[_], Err] {
   /** the input type, e.g. ids of data */
   type In
   /** the output type, e.g. the decoded block data */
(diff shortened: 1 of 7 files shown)


src/main/scala/tech/cryptonomic/conseil/generic/chain/DataFetcher.scala
```diff
@@ -79,7 +81,7 @@ object DataFetcher {
    * Example: the fetcher for tezos blocks might have type `Aux[Future, List, Int, BlockData, String]
    * where the Int is for the offset from a given block, and String is a representation of a json value.
    */
```
piotrkosecki: I think you could update docs here

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataFetcher.scala
```diff
@@ -79,7 +81,7 @@ object DataFetcher {
...
-  private type Aux[Eff[_], Coll[_], Input, Output, Encoding] = DataFetcher[Eff, Coll] {
+  private type Aux[Eff[_], Coll[_], Err, Input, Output, Encoding] = DataFetcher[Eff, Coll, Err] {
```
piotrkosecki: I think you could update docs here

piotrkosecki:
beside one doc which possibly could be updated, I'm ok with it

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataFetcher.scala
ivanopagano: thanks, missed that

PR #285, opened 2019-03-20
Fix mainnet errors on voting data decoding

ivanopagano:
* Use default when fetching json for voting data gives no result
 * increase the logging detail for every failed fetching operation


ivanopagano:
if the travis tests fails, please wait for me to fix this, I might have accidentally mixed commits from the other PR I'm on for the tests, so I might have to change that tomorrow

dorians:
I really like that now we can run ``Lorre`` against ``mainnet`` but I don't like that we have so many errors. When I run code from this PR is really hard to track progress because the output is so messy. I'm not familiar with all the additional data we try to fetch but what's the reason for having errors only on ``mainnet`` and not on other networks? It looks like there is some divergence of schemas and we're hiding errors.

ivanopagano:
> I really like that now we can run `Lorre` against `mainnet` but I don't like that we have so many errors. When I run code from this PR is really hard to track progress because the output is so messy. I'm not familiar with all the additional data we try to fetch but what's the reason for having errors only on `mainnet` and not on other networks? It looks like there is some divergence of schemas and we're hiding errors.

Indeed it's bothersome to have so much noise. The reason should hopefully be that such data was not there initially when mainnet went live, but was introduced as protocol upgrades of the chain.

We might think of turning off logging of warns until mainnet catches up to blocks with data available.

I wouldn't completely ignore them as they might turn useful for newly baked blocks, where such data should be available and missing response should not occur.
Or it might be that mainnet uses separate endpoint paths for such data and we need to follow suite.

We'll have to hear from @vishakh about this

vishakh:
@dorians Totally understand your concern but, as @ivanopagano said, we are in a difficult place due to the protocol upgrade in November. In the future, we can think about accounting for specific protocol functionalities between block ranges but for now we have to ensure we have false positives so we avoid false negatives.

PR #287, opened 2019-03-21
#256 Michelson integration - annotations support

dorians:
Brings annotations support to Michelson Parser.
Documentation for annotations: http://tezos.gitlab.io/mainnet/whitedoc/michelson.html#x-annotations

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/michelson/parser/JsonParser.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/michelson/parser/JsonParser.scala
@@ -55,11 +55,16 @@ object JsonParser {
    * Empty expression is represented as an empty array in JSON.
    *
    * */
-  case class JsonType(prim: String, args: Option[List[Either[JsonExpression, Nil.type]]]) extends JsonExpression {
-    override def toMichelsonExpression = MichelsonType(prim, args.getOrElse(List.empty).map {
-      case Left(it) => it.toMichelsonExpression
-      case Right(_) => MichelsonEmptyExpression
-    })
+  case class JsonType(prim: String,
+                      args: Option[List[Either[JsonExpression, Nil.type]]],
+                      annots: Option[List[String]] = None) extends JsonExpression {
+    override def toMichelsonExpression = MichelsonType(
+      prim = prim,
+      args = args.getOrElse(List.empty).map {
+        case Left(jsonExpression) => jsonExpression.toMichelsonExpression
+        case Right(_) => MichelsonEmptyExpression
+      },
+      annots.getOrElse(List.empty))
   }
 
   /*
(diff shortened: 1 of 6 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/michelson/parser/JsonParser.scala
```diff
@@ -55,11 +55,16 @@ object JsonParser {
    * Empty expression is represented as an empty array in JSON.
    *
    * */
-  case class JsonType(prim: String, args: Option[List[Either[JsonExpression, Nil.type]]]) extends JsonExpression {
-    override def toMichelsonExpression = MichelsonType(prim, args.getOrElse(List.empty).map {
-      case Left(it) => it.toMichelsonExpression
-      case Right(_) => MichelsonEmptyExpression
-    })
+  case class JsonType(prim: String,
```
ivanopagano: Just for the sake of consistency (which is not always there anyway, I know... 😝 ), can we agree to start args on a separate line when they're not all inlined?

i.e.
```scala
case class JsonType(
  prim: String,
  args: ...
  ...
)
```

src/main/scala/tech/cryptonomic/conseil/tezos/michelson/renderer/MichelsonRenderer.scala
```diff
@@ -9,14 +9,14 @@ object MichelsonRenderer {
     def render(): String = self match {
 
       // instructions
-      case MichelsonSingleInstruction(prim, List()) => prim
-      case MichelsonSingleInstruction(prim, args) => s"$prim ${args.map(_.render()).mkString(" ")}"
+      case MichelsonSingleInstruction(prim, List(), List()) => prim
```
ivanopagano: Just catched this, here you can and will probably want to pattern match against `Nil` which is a case object.

Otherwise we call a method `apply` (or `unapply` not sure) on the `List` object just create the equivalent value.

src/main/scala/tech/cryptonomic/conseil/tezos/michelson/renderer/MichelsonRenderer.scala
```diff
@@ -9,14 +9,14 @@ object MichelsonRenderer {
...
+      case MichelsonSingleInstruction(prim, args, annotations) => s"$prim ${(annotations ++ args.map(_.render())).mkString(" ")}"
       case MichelsonInstructionSequence(args) => s"{ ${args.map(_.render()).mkString(" ; ")} }"
       case MichelsonEmptyInstruction => "{}"
 
       // expressions
-      case MichelsonType(name, List()) => name
-      case MichelsonType(name, args) => s"($name ${args.map(_.render()).mkString(" ")})"
+      case MichelsonType(name, List(), List()) => name
```
ivanopagano: Same as before, please pattern match againt `Nil` instead of a method call.

PR #288, opened 2019-03-22
Conseil 276 extra block header data

### LINKED ISSUE #276: Additional header fields missing on block table
Some follow-ups are necessary:

-We want to update the blocks table with the following fields from block_header_metadata: "baker", "voting_period_kind", "nonce_hash", "consumed_gas", and all of the subfields of "level".

_Originally posted by @ba205 in https://github.com/Cryptonomic/Conseil/pull/243#issuecomment-460440748_

ivanopagano:
Add a few missing bits of information on the blocks table, related to block metadata info.
closes #276 

The information on _current active proposal_ for each block is now read directly from the available metadata, instead of doing a separate additional rest call, as before.

add the following columns to the `blocks` table

```sql
    baker character varying,
    nonce_hash character varying,
    consumed_gas numeric,
    meta_level integer,
    meta_level_position integer,
    meta_cycle integer,
    meta_cycle_position integer,
    meta_voting_period integer,
    meta_voting_period_position integer,
    meta_commitment boolean
```

None


src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -85,7 +87,7 @@ object TezosTypes {
     chain_id: Option[String],
     hash: BlockHash,
     header: BlockHeader,
-    metadata: BlockHeaderMetadata
+    metadata: Either[BlockHeaderMetadata, GenesisMetadata.type]
```
vishakh: As we understand it, `Either` is by convention used to capture and handle error messages using `Left`. Can we instead use a sum type here instead?

doc/conseil.sql
```diff
@@ -89,7 +89,17 @@ CREATE TABLE public.blocks (
...
+    baker character varying,
+    nonce_hash character varying,
+    consumed_gas numeric,
+    meta_level integer,
+    meta_level_position integer,
+    meta_cycle integer,
+    meta_cycle_position integer,
+    meta_voting_period integer,
+    meta_voting_period_position integer,
+    meta_commitment boolean
```
vishakh: Could we rename this to `expected_commitment` please in order to keep the columns consistent with the Tezos node?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
ivanopagano: `Either` can generally be used to represent a common "Sum type" with 2 alternatives.
It's commonly used to represent the sum of possible results of an operation when you can have a successful or failed outcome. But that's not a rule, it's a convention.

An alternative encoding of sum types is by shallow sealed trait hierachies (ADTs).
I will try to swap definition using the latter and see if the result is more palatable, in fact

doc/conseil.sql
ivanopagano: absolutely, will do

ivanopagano:
WIP to rapidly fix the dropped coverall percentage ETA 30 mins

src/test/scala/tech/cryptonomic/conseil/tezos/JsonDecodersTest.scala
```diff
@@ -39,6 +39,23 @@ class JsonDecodersTest extends WordSpec with Matchers with EitherValues {
 
     }
 
+    "decode null fields as empty Options" in {
```
dorians: This test touches circe directly. It looks like we are missing some abstraction here.

src/test/scala/tech/cryptonomic/conseil/tezos/DatabaseConversionsTest.scala
```diff
@@ -46,15 +48,95 @@ class DatabaseConversionsTest
...
+        converted.metaCyclePosition ::
+        converted.metaVotingPeriod ::
+        converted.metaVotingPeriodPosition ::
+        converted.expectedCommitment :: Nil) {
+        _ shouldBe 'empty
+      }
+
+    }
+
+    "convert a tezos block to a database row" in {
```
dorians: I'd prefer to have a declaration and comparison in one place so I can see structure before and after conversion. Right now It looks like repeating production code here in tests.

src/test/scala/tech/cryptonomic/conseil/tezos/DatabaseConversionsTest.scala
```diff
@@ -20,7 +22,7 @@ class DatabaseConversionsTest
     val groupHash = OperationHash("operationhash")
 
     //keep level 1, dropping the genesis block
-    val block = generateSingleBlock(atLevel = 1, atTime = testReferenceDateTime)
+    val (genesis :: block :: Nil) = generateBlocks(toLevel = 1, startAt = testReferenceDateTime)
```
dorians: What happens when ``generateBlocks`` returns one-element collection?

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTestFixtures.scala
```diff
@@ -65,7 +65,21 @@ trait TezosDataGeneration extends RandomGenerationKit {
...
+        level_position = rnd.nextInt(),
+        cycle = rnd.nextInt(),
+        cycle_position = rnd.nextInt(),
+        voting_period = rnd.nextInt(),
+        voting_period_position = rnd.nextInt(),
+        expected_commitment = rnd.nextBoolean()
+      )
+    }
+
+    def generateOne(level: Int, predecessorHash: BlockHash, genesis: Boolean = false): Block =
```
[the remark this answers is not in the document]
ivanopagano: My personal position here is
- don't use the same approach between production and test code
- much of the code is essentially the same, I'm already tempted to rewrite everything to Property-check generators
- hence I don't see much value in duplicating a test-data generation method for the sake of being more explicit, while I agree that being clear on testing is ok, being down-to-the-domain language I consider an unwarranted cost

Given this, let me know how do you see it, I'm open to discussion if it's for the overall process improvement

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
```diff
@@ -306,7 +313,7 @@ class TezosDatabaseOperationsTest
       implicit val randomSeed = RandomSeed(testReferenceTimestamp.getTime)
 
       //generate data
-      val blocks = generateBlockRows(2, testReferenceTimestamp)
+      val blocks @ (last :: first :: genesis :: Nil) = generateBlockRows(toLevel = 2, startAt = testReferenceTimestamp)
```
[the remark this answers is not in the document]
ivanopagano: genesis is kinda "foundation", it's the taken-for-granted root.
Then the regular blocks builds on top of genesis, so first is the first "real" block, and last the latest (actually second)

Please suggest better naming if it might help, keeping in mind that the genesis is "special", and not to be confused with the others, it's a corner case element

src/test/scala/tech/cryptonomic/conseil/tezos/DatabaseConversionsTest.scala
ivanopagano: I feel your case, sure.
This is the result of having random data generated for the test. Which is why I would love to find time to rethink it in terms of generic properties to satisfy, instead of using a single manually made sample to check against.

This test is almost useless per se, as it duplicates code logic, but it might have its reason to be because
- we plan to try and use the Chimney library to do the conversion with less manual copying, in which case the test will be different code-wise (which is not such a strong case, nonetheless)
- I added the case because otherwise the test coverage percentage would treat this as "untested" code, and lower our rating... that's the price to pay if we want to have a public-facing decent coverage

Here too, I'm open to discussion

src/test/scala/tech/cryptonomic/conseil/tezos/DatabaseConversionsTest.scala
ivanopagano: when it happens we need to also rewrite the pattern matching val binding

ivanopagano: To be more precise, the test will fail as soon as you change the parameters that drives the number of  generated blocks.
The code will still compile, though, which is a shame, I agree.

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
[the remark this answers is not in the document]
piotrkosecki: I agree, naming is a bit confusing, but as there are 3 blocks, why not name just `secondBlock :: firstBlock :: genesis :: Nil` - this could avoid confusion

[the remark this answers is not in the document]
ivanopagano: renamed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -100,11 +102,34 @@ object TezosTypes {
...
   )
 
+  sealed trait BlockMetadata extends Product with Serializable
+
+  final case object GenesisMetadata extends BlockMetadata
 
   final case class BlockHeaderMetadata(
-    balance_updates: Option[List[OperationMetadata.BalanceUpdate]]
+    balance_updates: List[OperationMetadata.BalanceUpdate],
+    nonce_hash: Option[NonceHash],
```
piotrkosecki: just wondering possibly for future improvement, how much work would be to make circe decode/encode if we used camel case in the case classes instead of snake case. Would be it worth it? what do you think?

src/test/scala/tech/cryptonomic/conseil/tezos/DatabaseConversionsTest.scala
dorians: Ahhh, ok. I didn't realize that this is a test and I fought that we have no control over what comes from Tezos.

src/test/scala/tech/cryptonomic/conseil/tezos/DatabaseConversionsTest.scala
dorians: Ok, so that's the reason I would love to have nice integration tests which are part of the coverage. Like in hexagonal architecture, where domain code can be tested with unit tests and all the adapters like postresql adapter and all conversions between domain and adapter's dtos are tested with integration tests (as part of the process and no in separate tests). 

Example:

> trait PostgresqlStorage {
>   def save(block: Block): Unit
>   def get(hash: String): Block
> }

Implementation of this adapter can be tested by saving some block and then receiving it back from DB with embedded postresql database run with a test suite. ``Block`` is the domain class and the conversion to database's dto will be tested by the way, as it is specific to a given database.

I think that this approach could solve both problems: testing implementation instead of behavior and coverage issues. What do you think?

src/test/scala/tech/cryptonomic/conseil/tezos/JsonDecodersTest.scala
dorians: This is how I did it with Michelson parsing: https://github.com/Cryptonomic/Conseil/blob/master/src/main/scala/tech/cryptonomic/conseil/tezos/michelson/parser/JsonParser.scala#L208
Circe is used only internally and API doesn't expose it at all. I agree that it requires some code to be written but in my personal opinion, it's usually worth it. On the other hand, if refactor went smooth, maybe, in that case, it's not applicable. 

Anyway, my main point was that I'd prefer tests like "check if that production object with NULL field decodes correctly" rather than "let's check if that library with configuration copied from production code works correctly on some test object", or, even better: "check if this process works correctly if I pass NULL". The goal is to construct tests in a way to test behavior not implementation.

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTestFixtures.scala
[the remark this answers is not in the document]
dorians: That's fine for me as long as we remember to use named parameters for boolean parameters.

src/test/scala/tech/cryptonomic/conseil/tezos/JsonDecodersTest.scala
ivanopagano: I get your point, and this test is a regression check.

Previously it did the same verification using the Jackson library, and I ported the test to circe decoders once I switched to that, to preserve behaviour across implementations.

As you might see I actually tried to somehow keep things neatly separated 
 at least, by defining a `JsonDecoders.Circe` submodule, so that other JsonDecoders can be actually defined, using different library support.

In some sense the issue is that, while writing a Parser trait is pretty straight-forward, I didn't check how much additional work would be needed if we are to add encoding, too.

I'll think again about this, and we might possibly reuse your abstract approach everywhere.
I am missing time to think about this at the moment, so I'll just defer this task to some future moment

src/test/scala/tech/cryptonomic/conseil/tezos/DatabaseConversionsTest.scala
ivanopagano: The actually like the current separation of Conversion logic (adapting the data) from the Storage abstraction (saving and fetching it).

It's easier an modular to both maintain and test in my opinion.

I'm with you as long as we talk about the DatabaseOperator (or anyway we want to call the Repository  module). I foresee the possibility of generalising the interface to the storage and provide parametric and separate implementations. That would also allow separate testing for the compliance of each and every implementation.

ivanopagano: Of course this leave open the fact that testing conversion of a randomly generated instance would always imply some form of duplication of logic... 
Also keep in mind that the testing adapter might become much simpler than the real one, if the conversion logic gets sophisticated.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
ivanopagano: I think it shouldn't be that hard. I didn't check, but I can imagine that the implicit decoding configuration can be simply tailored to adapt the casing (if it doesn't already by itself).

I'm not sure we might want to do that, tho, as I find it very convenient to have the same exact field naming between json and scala.
I've often found that mismatches in casing rules actually make things uselessly harder to work on, only for the sake of conventions that are not necessarily useful.

PR #289, opened 2019-03-22
Fix for the decimal json encoding

piotrkosecki:
The issue with treating some numbers as strings was that decimal type from database was not handled correct way - decimal columns were encoded as strings. Added correct encoding.


dorians:
Don't we have any tests for this?

vishakh:
> Don't we have any tests for this?

Great point! We need some black box tests of our REST interface.

PR #290, opened 2019-03-22
#256 Michelson integration - formatting michelson

dorians:
This PR brings better formatting. IF instructions are now rendered in rows which makes them easier to debug. Example of IF instruction you can see here: https://tzscan.io/KT1LKhiZz9XFM9iRppLAdq58XoHK3eNdhhQ5

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/michelson/renderer/MichelsonRenderer.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/michelson/renderer/MichelsonRenderer.scala
@@ -9,20 +9,27 @@ object MichelsonRenderer {
     def render(): String = self match {
 
       // instructions
-      case MichelsonSingleInstruction(prim, List(), List()) => prim
-      case MichelsonSingleInstruction(prim, args, annotations) => s"$prim ${(annotations ++ args.map(_.render())).mkString(" ")}"
+      case MichelsonSingleInstruction(name, List(sequence1: MichelsonInstructionSequence, sequence2: MichelsonInstructionSequence), _) => {
+        val indent = " " * (name.length + 1)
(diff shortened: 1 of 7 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/michelson/renderer/MichelsonRenderer.scala
```diff
@@ -9,20 +9,27 @@ object MichelsonRenderer {
     def render(): String = self match {
 
       // instructions
-      case MichelsonSingleInstruction(prim, List(), List()) => prim
-      case MichelsonSingleInstruction(prim, args, annotations) => s"$prim ${(annotations ++ args.map(_.render())).mkString(" ")}"
+      case MichelsonSingleInstruction(name, List(sequence1: MichelsonInstructionSequence, sequence2: MichelsonInstructionSequence), _) => {
+        val indent = " " * (name.length + 1)
+        val embeddedIndent = " " * (name.length + 3)
```
ivanopagano: `indent + " " * 2` is more clear? or I might have misinterpreted intent

If the two concept are tied, why not the definitions? 😉

ivanopagano: What about multiple nesting? Does this handle the case?

dorians: I tried with ``def`` but it looked worse for me. It supports multiple nesting. I added a test for that case.

ivanopagano: I didn't meant to use `def`, just to reuse the `ident` val inside the definition of `embeddedIdent`

Issue #293, opened 2019-03-25
"ERROR:  invalid byte sequence for encoding "UTF8": 0x00" in Conseil master


leenpaws:
Using the current master at commit 25aaaba84be153018654b6671364856f3f06bf28 we see an issue in the parallel staging environment which runs against tezos-mainnet, the logs show the following output before lorre shuts down:
```
Mar 25 21:15:08 staging-test conseil-staging[12744]: io.circe.ParsingFailure: exhausted input
Mar 25 21:15:08 staging-test conseil-staging[12744]: 21:15:08.435 [lorre-system-akka.actor.default-dispatcher-19] WARN  t.c.conseil.tezos.TezosNodeOperator - I fetched a proposal protocol json from tezos node that I'm unable to decode:
Mar 25 21:15:08 staging-test conseil-staging[12744]: io.circe.ParsingFailure: exhausted input
Mar 25 21:15:08 staging-test conseil-staging[12744]: 21:15:08.435 [lorre-system-akka.actor.default-dispatcher-19] WARN  t.c.conseil.tezos.TezosNodeOperator - I fetched a proposal protocol json from tezos node that I'm unable to decode:
Mar 25 21:15:08 staging-test conseil-staging[12744]: io.circe.ParsingFailure: exhausted input
Mar 25 21:15:11 staging-test conseil-staging[12744]: 21:15:11.474 [lorre-system-akka.actor.default-dispatcher-18] ERROR tech.cryptonomic.conseil.Lorre$ - Could not write blocks or accounts checkpoints to the database.
Mar 25 21:15:11 staging-test conseil-staging[12744]: org.postgresql.util.PSQLException: ERROR: invalid byte sequence for encoding "UTF8": 0x00
Mar 25 21:15:11 staging-test conseil-staging[12744]: #011at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2477)
Mar 25 21:15:11 staging-test conseil-staging[12744]: #011at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2190)
Mar 25 21:15:11 staging-test conseil-staging[12744]: #011at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:300)
Mar 25 21:15:11 staging-test conseil-staging[12744]: #011at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:428)
[...]

PR #296, opened 2019-03-27
Stripping backspaces from Tezos node JSON.

vishakh:
This addresses #293. 

The previous solutions tried to strip out all control characters, assuming the source strings actually contained those characters. However, the Tezos node data actually contained string-encoded control characters:

image

Thus, stripping out the control characters did not actually remove the anomalous data. This blew up down the line as the database persistence logic somehow converted the `\u0000` into an actual null character, which Postgres did not like.

In this solution, the backspace symbol is replaced by the string `backspace` so no mischief is possible down the line. The replacement string is an actual word so in the future developers can easily trace cases where the database data does not match the data returned by the node, i.e. by grepping the source code for the word `backspace`.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/util/JsonUtil.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/util/JsonUtil.scala
@@ -56,7 +56,7 @@ object JsonUtil {
     lazy val emptyObject = JsonString("{}")
 
     /** add standard cleaning for input json */
-    def sanitize(s: String): String = s.filterNot(_.isControl)
+    def sanitize(s: String): String = s.filterNot(_.isControl).replaceAll("\\\\", "backslash")
 
   }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/tech/cryptonomic/conseil/util/JsonUtil.scala
```diff
@@ -56,7 +56,7 @@ object JsonUtil {
     lazy val emptyObject = JsonString("{}")
 
     /** add standard cleaning for input json */
-    def sanitize(s: String): String = s.filterNot(_.isControl)
+    def sanitize(s: String): String = s.filterNot(_.isControl).replaceAll("\\\\", "backslash")
```
ivanopagano: So it wasn't _backspace_ but backslash 😁  
Apart from useless comments...

You might want to use raw string literals as in
`replaceAll("""\\""", "backslash")`

if it looks better

vishakh: The particular problem was actually _backspace_, but it was in literal form so the earlier code did not strip it out. By replacing _backslash_ we 'disable' any control character.

vishakh:
> I wonder why the coverall regression... thinking as you actually covered the new case as well

:man_shrugging:

PR #297, opened 2019-03-27
Conseil 286 use slick metadata api for metadata route

piotrkosecki:
This PR introduces a lot of changes in the metadata route handling, but also in some parts connected to it. As discussed with @ivanopagano it is supposed to be a middle ground between moving away from  using `Future` in favour of cats `IO`.

- now `TezosPlatformDiscoveryOperations` will be pre-caching data from `Slick` metadata queries about the tables and columns without distinct counts(those will be initialized with the first call) using `init()` method - if initial counts would be included it would take ~4 minutes from the start of the `Conseil` till metadata and query endpoints would be usable

- for implementing this I used cats MVar which provides thread safety, but downside of that I had to use a lot of `unsafeRunSync` and `unsafeToFuture` for mixing `IO`, `Future` and `DBIO`

- removed redundant query fields validation from the code

- removed unused trait `RouteHandling` and few other things

- changed `TezosPlatformDiscoveryOperationsTest` so it would use `Future` not `DBIO` - eventually we should use `IO` but probably not with slick

None

```
   # Runtime settings for Conseil server
   conseil: {
     hostname: "0.0.0.0"
     port: 1337
+    cacheTTL: 15 minutes
   }
(...)
-  # akka-http-caching config described here:
-  # https://doc.akka.io/docs/akka-http/current/common/caching.html#frequency-biased-lfu-cache
   http {
-    caching.lfu-cache {
-      max-capacity = 500
-      initial-capacity = 100
-      time-to-live = 15 minutes
-      time-to-idle = 10 minutes
-    }
     server.request-timeout = 1 minute
   }

```


src/main/resources/reference.conf
```diff
@@ -4,6 +4,7 @@
 conseil: {
   hostname: "0.0.0.0"
   port: 1337
+  cache-ttl: 15 minutes
```
dorians: Shouldn't we have a specific timeout for each operation rather than one global timeout?

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -77,22 +82,41 @@ object DataTypes {
...
+
       val query = Query().patchWith(this)
 
       val invalidQueryFields = query
         .fields
-        .filterNot(field => TezosPlatformDiscoveryOperations.areFieldsValid(entity, Set(field)))
-        .map(InvalidQueryField)
+        .map(field => tezosPlatformDiscovery.areFieldsValid(entity, Set(field)).map(_ -> field))
+        .sequence
+        .map(_.filterNot(_._1).map(fieldPair => InvalidQueryField(fieldPair._2)))
```
dorians: Please, use named parameters instead of expressions like ``_._1``. It's hard to realize what actually happens in this code.

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -77,22 +82,41 @@ object DataTypes {
...
+    Future[Either[List[QueryValidationError], Query]] = {
+      import cats.implicits._
+
       val query = Query().patchWith(this)
 
       val invalidQueryFields = query
         .fields
-        .filterNot(field => TezosPlatformDiscoveryOperations.areFieldsValid(entity, Set(field)))
-        .map(InvalidQueryField)
+        .map(field => tezosPlatformDiscovery.areFieldsValid(entity, Set(field)).map(_ -> field))
```
dorians: I don't understand this code. Method ``areFieldsValid`` return ``Future[Boolean]`` but you ignore the result and map it to its original value. The same is in the method below.

src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
```diff
@@ -92,15 +71,15 @@ class PlatformDiscovery(config: PlatformsConfiguration, caching: HttpCacheConfig
...
     case (((platform, network, entity), attribute, filter), _) =>
-      TezosPlatformDiscoveryOperations.listAttributeValues(entity, attribute, Some(filter)).map { attributes =>
+      tezosPlatformDiscoveryOperations.listAttributeValues(entity, attribute, Some(filter)).map { attributes =>
         ConfigUtil.getNetworks(config, platform).find(_.network == network).map { _ =>
           attributes
         }
       }
   }
 
   /** Concatenated metadata routes with cache */
```
dorians: Is this comment still valid?

vishakh:
The approach looks generally good to me. However, the details are outside of my areas of expertise. Happy to merge after @dorians and @ivanopagano approve.

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: Look again at the arrow ;)

src/main/resources/reference.conf
piotrkosecki: actually we have only 2 endpoints using timeout, so I don't feel that it is necessary to have different TTLs for those endpoints -  but this could change when I get feedback about real usage, now its like fortune telling if this value will be ok, or if we will need different timeouts for different endpoints

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
dorians: hahaha :)

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala

src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
piotrkosecki: nope, its not, changed

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -35,8 +40,20 @@ object Conseil extends App with LazyLogging with EnableCORSDirectives with Conse
 
       val tezosDispatcher = system.dispatchers.lookup("akka.tezos-dispatcher")
       lazy val tezos = Tezos(tezosDispatcher)
-      lazy val platformDiscovery = PlatformDiscovery(platforms, caching)(tezosDispatcher)
-      lazy val data = Data(platforms)(tezosDispatcher)
+
+      // This part is a temporary middle ground between current implementation and moving code to use IO
+      implicit val contextShift: ContextShift[IO] = IO.contextShift(executionContext)
+      val attributesCache = MVar[IO].empty[Map[String, (Long, List[Attribute])]].unsafeRunSync()
```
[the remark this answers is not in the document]
piotrkosecki: because in the end we would want to have all of the IO operations to compose, so probably we would have one big for comprehension which would help avoid `unsafeRunSync` call. At this point I had a choice to put `unsafeRunSync` in the  `PlatformDicoveryOp` or in here. I chose `Conseil` object to show the direction in which we should go.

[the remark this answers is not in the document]
ivanopagano: Awesome 👍

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -77,22 +82,41 @@ object DataTypes {
     @JsonScalaEnumeration(classOf[OutputTypeRef]) output: Option[OutputType]
   ) {
     /** Method which validates query fields, as jackson runs on top of runtime reflection so NPE can happen if fields are missing */
-    def validate(entity: String): Either[List[QueryValidationError], Query] = {
+    def validate(entity: String, tezosPlatformDiscovery: TezosPlatformDiscoveryOperations)(implicit ec: ExecutionContext):
```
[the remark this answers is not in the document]
piotrkosecki: thats an outdated comment

PR #298, opened 2019-03-28
Conseil 291: handle errors on account processing

### LINKED ISSUE #291: Lorre fails to correctly store Account updates
Might be related to #263 

Some failure in the accounts updates fetch and db write process seems to fail the operation as a whole with the fairly relevant possibility of losing such account recent values, especially on Lorre bootstrap of data.

ivanopagano:
closes #291 
Now the accounts are taken using a data-fetcher and decoded with a custom circe decoder that would not try to parse micheline scripts, too.
This allows to separate that into a second step, after accounts are loaded.
Failure to parse the script, would simply log a warn and store the micheline as-is instead of converting to michelson format.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
@@ -47,12 +47,12 @@ object DatabaseConversions {
         Tables.AccountsRow(
           accountId = id.id,
           blockId = hash.value,
-          manager = manager,
+          manager = manager.value,
           spendable = spendable,
           delegateSetable = delegate.setable,
-          delegateValue = delegate.value,
+          delegateValue = delegate.value.map(_.value),
           counter = counter,
-          script = script.map(_.toString),
+          script = script.map(_.code.toString),
           balance = balance,
           blockLevel = level
         )
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
@@ -3,7 +3,8 @@ package tech.cryptonomic.conseil.tezos
 import cats._
 import cats.data.Kleisli
 import com.typesafe.scalalogging.LazyLogging
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Try
 import scala.util.control.NonFatal
 import tech.cryptonomic.conseil.generic.chain.DataFetcher
 import tech.cryptonomic.conseil.util.JsonUtil
@@ -15,7 +16,6 @@ import TezosTypes._
 trait BlocksDataFetchers {
   //we require the cabability to log
   self: LazyLogging =>
-  import scala.concurrent.Future
   import cats.instances.future._
   import cats.syntax.applicativeError._
   import cats.syntax.applicative._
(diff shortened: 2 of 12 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -47,12 +47,12 @@ object DatabaseConversions {
         Tables.AccountsRow(
           accountId = id.id,
           blockId = hash.value,
-          manager = manager,
+          manager = manager.value,
```
ba205: Were the changes in this file due to the schema being wrong/optional fields?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
```diff
@@ -291,4 +291,69 @@ trait BlocksDataFetchers {
...
+  import cats.syntax.applicativeError._
+  import cats.syntax.applicative._
+  import JsonDecoders.Circe.decodeLiftingTo
+
+  implicit def fetchFutureContext: ExecutionContext
+
+  /* reduces repetion in error handling */
+  private def logWarnOnJsonDecoding[Encoded](message: String): PartialFunction[Throwable, Future[Unit]] = {
+    case decodingError: io.circe.Error =>
+      logger.warn(message, decodingError).pure[Future]
```
ba205: circe errors are, in general, vague. Is the message here descriptive enough to compensate for that?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -57,11 +57,13 @@ object TezosNodeOperator {
...
 class TezosNodeOperator(val node: TezosRPCInterface, val network: String, batchConf: BatchFetchConfiguration)(implicit val fetchFutureContext: ExecutionContext)
   extends LazyLogging
-  with BlocksDataFetchers {
+  with BlocksDataFetchers
+  with AccountsDataFetchers {
   import TezosNodeOperator.isGenesis
   import batchConf.{accountConcurrencyLevel, blockOperationsConcurrencyLevel, blockPageSize}
 
   override val fetchConcurrency = blockOperationsConcurrencyLevel
+  override val accountsFetchConcurrency = accountConcurrencyLevel
```
ba205: how do we tune concurrency levels for blocks vs accounts again?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -326,34 +327,37 @@ object TezosTypes {
...
-                    spendable: Boolean,
-                    delegate: AccountDelegate,
-                    script: Option[String],
-                    counter: Int
-                    )
+    manager: PublicKeyHash,
+    balance: scala.math.BigDecimal,
+    spendable: Boolean,
+    delegate: AccountDelegate,
+    script: Option[AccountScript],
```
ba205: I really like this pattern of making type synonyms for very specific instances of data. This isn't just an Option of a String, this needs to be an Option of an AccountScript. The benefit of the compiler catching these bugs reduces a lot of overhead, but one thing I wonder is: have we not seen errors of this kind because they've been completely eliminated, or is the compiler not giving us enough information when things like happen indirectly, and if the latter is the case, how do we fix this?

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
ivanopagano: I added a wrapping class definition, i.e.
`manager: String` -> `manager: PublicKeyHash`

So now to read the pkh string you need to extract the `.value` 🙂

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
ivanopagano: This is one of the things we discussed the other day, I plan to spend some time parsing circe's errors to see if they can be translated to accurate and readable messages

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ivanopagano: To be honest this difference is so subtle that I don't even know anymore if the different tuning is worth it.

The parallelism involved is _only_ used when the http responses are streamed back and the utf8 body content is extracted eagerly.
The processing is quite minimal at this point, and I'm not even sure if the concurrency is getting us any real gain.

But since we could change things later to do more processing in parallel, I'd leave it as is for the moment

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
ivanopagano: Not sure if I get this right, but if you're asking how we forgot to "type" some of the contents with more precise definitions than `String` or `Long`, the answer is that the code is evolving.
Hence we're migrating piece by piece.

I first used this approach to decode block operations, because it was necessary for the decoders to apply specific rules of validation to the content of different fields, based on the wrapping custom type. E.g. when we decode a PublicKeyHash or a BlockHash, the parsing will fail if the field string is not correctly base58check-encoded.

The current PR needs pushed me to switch from jackson to circe, just so that I could apply a custom decoder, to the script field content.
I'll explain because it's important that this is clear.
The script contains a json-object and the parsers won't accept to convert that format to a json-string, because they're syntactically different, right?
So I customized the decoder (with circe is easy to do it if I create a specific target type, like `AccountScript`) that will read a json-object, and then convert it to a string that is put inside the `AccountScript` value class.

Does this answer your doubt?

PR #303, opened 2019-03-28
Made substitution string for sanitization simpler

vishakh:
Quick fix in response to testing issues.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/util/JsonUtil.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/util/JsonUtil.scala
@@ -56,7 +56,7 @@ object JsonUtil {
     lazy val emptyObject = JsonString("{}")
 
     /** add standard cleaning for input json */
-    def sanitize(s: String): String = s.filterNot(_.isControl).replaceAll("""\\u[[a-zA-Z0-9]]{1,4}""", "REDACTED_BY_CONSEIL")
+    def sanitize(s: String): String = s.filterNot(_.isControl).replaceAll("""\\u[a-fA-F0-9]{1,4}""", "REDACTED_BY_CONSEIL")
 
   }
 
--- a/src/main/scala/tech/cryptonomic/conseil/util/JsonUtil.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/util/JsonUtil.scala
@@ -56,7 +56,7 @@ object JsonUtil {
     lazy val emptyObject = JsonString("{}")
 
     /** add standard cleaning for input json */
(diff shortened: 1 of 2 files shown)


src/main/scala/tech/cryptonomic/conseil/util/JsonUtil.scala
```diff
@@ -56,7 +56,7 @@ object JsonUtil {
     lazy val emptyObject = JsonString("{}")
 
     /** add standard cleaning for input json */
-    def sanitize(s: String): String = s.filterNot(_.isControl).replaceAll("""\\u[[a-zA-Z0-9]]{1,4}""", "[REDACTED_BY_CONSEIL]")
+    def sanitize(s: String): String = s.filterNot(_.isControl).replaceAll("""\\u[[a-zA-Z0-9]]{1,4}""", "REDACTED_BY_CONSEIL")
```
ivanopagano: Why the double brackets? Did I suggest that?

Shouldn't the regex be `"""\\u[a-zA-Z0-9]{1,4}"""` ?

ivanopagano: And now that I come to look better at it

`"""\\u[a-fA-F0-9]{1,4}"""` should be more lenient, we where redacting things that couldn't have been unicode values

src/test/scala/tech/cryptonomic/conseil/util/JsonUtilTest.scala
```diff
@@ -71,7 +71,7 @@ class JsonUtilTest extends WordSpec with Matchers with JsonMatchers {
 
     "sanitize unwanted input, by removing any string-encoded ISO control chars from the json string" in {
       val invalid: String = "demo\\u0000"
-      (JsonString sanitize invalid) shouldBe "demo[REDACTED_BY_CONSEIL]"
+      (JsonString sanitize invalid) shouldBe "demoREDACTED_BY_CONSEIL"
```
ivanopagano: Would you care to replace this with a more through test I prepared anyway?

```scala
    "sanitize unwanted input, by removing any string-encoded ISO control chars from the json string" in {
      val invalid = Set(
        "demo\\u0",
        "demo\\u00",
        "demo\\u000",
        "demo\\u0000",
        )

      forAll(invalid)(JsonString.sanitize(_) shouldBe "demoREDACTED_BY_CONSEIL")
    }

    "not sanitize char sequences that only resemble ISO control chars from the json string" in {
      val valid = "demo\\uGO"
      val trailing = "demo\\u0000F"

      JsonString sanitize valid shouldEqual valid
      //preserve the extra char
      JsonString sanitize trailing shouldBe "demoREDACTED_BY_CONSEILF"
    }

```

PR #305, opened 2019-03-29
Format Conseil automatically using Scalafmt

ivanopagano:
To get a consistent format and eliminate the problem of personal taste on readability we introduce an automatic-formatting tool: scalafmt.

It can be used through sbt using [tasks](https://scalameta.org/scalafmt/docs/installation.html#task-keys) (e.g. scalafmtAll, scalafmtSbt) or, as strongly suggested, adding automatic format on [save on intellij](https://scalameta.org/scalafmt/docs/installation.html#format-on-save).
This usually helps out a lot, by removing the need to think on proper formatting while writing code.

The current formatting is based on the values in the `.scalafmt.conf` file
You can see the outcome in the diff.

If you want to try a different set, tweaking the config, [here is the doc](https://scalameta.org/scalafmt/docs/configuration.html), then run the sbt task on the project.

I found that some of the configuration combinations can make the tool unstable, so please check everything works before.


src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -46,46 +51,48 @@ object Conseil extends App with LazyLogging with EnableCORSDirectives with Conse
...
       lazy val platformDiscovery = PlatformDiscovery(platforms, tezosPlatformDiscoveryOperations)(tezosDispatcher)
       lazy val data = Data(platforms, tezosPlatformDiscoveryOperations)(tezosDispatcher)
 
       val route = cors() {
-        enableCORS {
-          validateApiKey { _ =>
-            logRequest("Conseil", Logging.DebugLevel) {
-              tezos.route ~
-              AppInfo.route
+          enableCORS {
```
piotrkosecki: I told you at some point that auto code rearrangement will break those routes 😉

piotrkosecki: by break I meant rearrange it the different way than you proposed in one of my PRs - maybe scalafmt has something special prepared for the akka routes?

ivanopagano: I couldn't figure out if there was a substantial change when I reformatted those.
In any case, I simply thought that we have `//format: off` as our weapon for critical sections where formatting is an art more than a science 😄

ivanopagano:
superseded by #430

Issue #307, opened 2019-03-29
Accounts data is not formatted properly


umurb:
Here are the results I get in production and in dev:

**Production**
![Screen Shot 2019-03-29 at 4 42 06 PM](https://user-images.githubusercontent.com/37525692/55262390-0c5a2980-5244-11e9-88f7-cc783ba8ceec.png)

**Dev**
![Screen Shot 2019-03-29 at 4 43 52 PM](https://user-images.githubusercontent.com/37525692/55262407-0fedb080-5244-11e9-9e3f-1a6b06d6eec3.png)

umurb:
Fixed with #308

Issue #306, opened 2019-03-29
Average Fees are not working


umurb:
1 - I am unable to pick `operation_kind`. Conseil keeps on returning `endorsement`.
2 - I only get 0 as high, medium, and low fee.

umurb:
Fixed with #308

PR #308, opened 2019-03-30
Bug fixes and backwards compatibility measures for v1 API

vishakh:
As released versions of the Tezori / Galleon wallet for Tezos are using the v1 Conseil API, we must ensure the v1 API is working as expected in the current codebase. This PR addresses three issues.

For #307, a missing JSON encoder for accounts has been added (37015f1439bf443d3c5842c844cf6f5c60ac8bd9). 

Related to #307, I found the encoding of operation groups and operations was incorrect and fixed them (a2f0e3633b2e129fd93d025220ee482921bf5f9f).

For #306, I reverted the query parameter names back to the old ones so released wallets don't break when this code is pushed out (b5fd2fd9b7cf61688aa272e7192b55117dfe12d3). 

I had some trouble with one implicit function and left a corresponding comment for others' review.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/routes/openapi/DataJsonSchemas.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/routes/openapi/DataJsonSchemas.scala
@@ -44,10 +44,6 @@ trait DataJsonSchemas extends algebra.JsonSchemas with generic.JsonSchemas {
   implicit def operationGroupsRowSchema: JsonSchema[OperationGroupsRow] =
     genericJsonSchema[OperationGroupsRow]
 
-  /** Operation groups row schema */
-  implicit def operationsRowSchema: JsonSchema[OperationsRow] =
-    genericJsonSchema[OperationsRow]
-
   /** Average fees schema */
   implicit def avgFeeSchema: JsonSchema[AverageFees] =
     genericJsonSchema[AverageFees]
@@ -64,8 +60,8 @@ trait DataJsonSchemas extends algebra.JsonSchemas with generic.JsonSchemas {
   /** AnyMap schema */
   implicit def blocksByHashSchema: JsonSchema[AnyMap]
 
-  /** Operation schema */
-  implicit def operationSchema: JsonSchema[OperationsRow] =
+  /** Operation groups row schema */
+  implicit def operationsRowSchema: JsonSchema[OperationsRow] =
     genericJsonSchema[OperationsRow]
 
   /** Accounts row schema */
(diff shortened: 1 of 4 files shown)


src/main/scala/tech/cryptonomic/conseil/routes/openapi/TezosEndpoints.scala
```diff
@@ -80,15 +80,10 @@ trait TezosEndpoints extends algebra.Endpoints with DataJsonSchemas with ApiFilt
...
-    endpoint(
-      request = get(
-        url = commonPath / "operations" / "avgFees" /? qsFilter,
-        headers = header("apiKey")),
-      response = jsonResponse[AverageFees](docs = Some("Endpoint for average fees")).orNotFound(Some("Not found")),
-      tags = List("Fees")
-    )
+  /** Operation groups row schema */
+  /** Pasted here because somehow it is not being pulled in from DataJsonSchemas at all. */
+  override implicit def operationsRowSchema: JsonSchema[OperationsRow] =
```
vishakh: For some reason, `operationsRowSchema()` from `DataJsonSchemas` does not get pulled into the implicit context here. Removing this line causes errors and I cannot figure out why `operationGroupsRowSchema()` works and `operationsRowSchema()` does not. Any ideas, @piotrkosecki @dorians and @ivanopagano ?

piotrkosecki: I've investigated it and I think error message is wrong - it should be rather `ambiguous implicit`. There is already implicit for `OperationsRow`(`implicit def operationSchema: JsonSchema[OperationsRow]`, line 68), I removed it locally and it works. I saw that you used the new it explicitly in the encoder, so you can remove the old one, also it was poorly named, so thats why you didn't find it.

vishakh: Wow, I am not very impressed with the Scala compiler!

src/main/scala/tech/cryptonomic/conseil/routes/openapi/DataJsonSchemas.scala
```diff
@@ -60,8 +60,8 @@ trait DataJsonSchemas extends algebra.JsonSchemas with generic.JsonSchemas {
   /** AnyMap schema */
   implicit def blocksByHashSchema: JsonSchema[AnyMap]
 
-  /** Operation schema */
-  implicit def operationSchema: JsonSchema[OperationsRow] =
+  /** Operation groups row schema */
```
piotrkosecki: minor thing in the comment, I think you meant `Operations row schema`

Issue #130, opened 2018-07-23
Add ability to run Lorre for a block interval


vishakh:
At the moment, Lorre can only be run from a given block (head or manually specified) back to the genesis block (level zero). This is because all blocks in the database are expected to have a predecessor. 

We should:
- [ ] Relax the database constraint and allow blocks without predecessors
- [ ] Make the start (defaults to head) and end (defaults to level 0) block hashes command line parameters

anonymoussprocket:
Superseded by #213.

vishakh:
Reopening ticket so it rides off #213. 

We now want to existing `depth` command line parameter to Lorre to accept block ranges, e.g. `1000:2500'.

vishakh:
I would now recommend not changing the behaviour of the `depth` argument. Instead, there should be a new argument (e.g. `--headHash`) which allows the optional specification of the hash of the maximum desired block. In combination with `depth` the new argument should allow us to run Lorre for any desired range. 

For example, if I wanted to run Lorre from block 127,000 to 127,100 I would set the the head hash parameter to `BMCUFw5J1ocXibgZifMLGgTeqezZUWtWuZp1j3PLhm2fLnShLqb` and the depth parameter to `100`.

PR #310, opened 2019-03-31
#130 head hash parameter

dorians:
I followed @vishakh's advice to create ``--headHash`` parameter because I couldn't find a method in Tezos API to get a block by level.

This PR introduces ``--headHash`` to Lorre option which changes the default head block to a given one. Alongside with ``--depth`` it enables to fetch a given range of blocks in the middle of the chain.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/config/LorreAppConfig.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/config/LorreAppConfig.scala
@@ -23,13 +23,17 @@ trait LorreAppConfig {
     case _ => None
   }
 
-  private case class ArgumentsConfig(val depth: Depth = Newest, verbose: Boolean = false, network: String = "")
+  private case class ArgumentsConfig(depth: Depth = Newest, verbose: Boolean = false, headHash: Option[String] = None, network: String = "")
 
   private val argsParser = new OptionParser[ArgumentsConfig]("lorre") {
     arg[String]("network")
       .required()
       .action( (x, c) => c.copy(network = x))
       .text("which network to use")
+
+    opt[Option[String]]("headHash")
+      .action( (x, c) => c.copy(headHash = x))
+      .text("from which block to start. Default to actual head")
 
     opt[Option[Depth]]('d', "depth")
       .validate{
@@ -56,13 +60,13 @@ trait LorreAppConfig {
 
     val loadedConf = for {
       args <- readArgs(commandLineArgs)
-      ArgumentsConfig(depth, verbose, network) = args
+      ArgumentsConfig(depth, verbose, headHash, network) = args
       lorre <- loadConfig[LorreConfiguration](namespace = "lorre")
       nodeRequests <- loadConfig[NetworkCallsConfiguration]("")
       node <- loadConfig[TezosNodeConfiguration](namespace = s"platforms.tezos.$network.node")
       streamingClient <- loadAkkaStreamingClientConfig(namespace = "akka.tezos-streaming-client")
       fetching <- loadConfig[BatchFetchConfiguration](namespace = "batchedFetches")
-    } yield CombinedConfiguration(lorre, TezosConfiguration(network, depth, node), nodeRequests, streamingClient, fetching, VerboseOutput(verbose))
+    } yield CombinedConfiguration(lorre, TezosConfiguration(network, depth, node, headHash), nodeRequests, streamingClient, fetching, VerboseOutput(verbose))
 
     //something went wrong
     loadedConf.left.foreach {
--- a/src/main/scala/tech/cryptonomic/conseil/config/LorreAppConfig.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/config/LorreAppConfig.scala
@@ -2,12 +2,12 @@ package tech.cryptonomic.conseil.config
 
 import tech.cryptonomic.conseil.config.Platforms._
 import tech.cryptonomic.conseil.util.ConfigUtil.Pureconfig.loadAkkaStreamingClientConfig
-import pureconfig.{CamelCase, ConfigFieldMapping, loadConfig}
-import pureconfig.ConfigReader
+import pureconfig.{CamelCase, ConfigFieldMapping, ConfigReader, loadConfig}
 import pureconfig.error.ConfigReaderFailures
 import pureconfig.generic.{EnumCoproductHint, ProductHint}
 import pureconfig.generic.auto._
 import scopt.{OptionParser, Read}
+import tech.cryptonomic.conseil.tezos.TezosTypes.BlockHash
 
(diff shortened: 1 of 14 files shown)


src/main/scala/tech/cryptonomic/conseil/config/LorreAppConfig.scala
```diff
@@ -23,14 +23,18 @@ trait LorreAppConfig {
...
-  private case class ArgumentsConfig(val depth: Depth = Newest, verbose: Boolean = false, network: String = "")
+  private case class ArgumentsConfig(depth: Depth = Newest, verbose: Boolean = false, headHash: Option[String] = None, network: String = "")
 
   private val argsParser = new OptionParser[ArgumentsConfig]("lorre") {
     arg[String]("network")
       .required()
       .action( (x, c) => c.copy(network = x))
       .text("which network to use")
 
+    opt[Option[String]]("headHash")
```
ivanopagano: please provide shorthand form (1 letter)

src/main/scala/tech/cryptonomic/conseil/config/Platforms.scala
```diff
@@ -52,7 +52,7 @@ object Platforms extends LazyLogging {
   }
 
   /** collects all config related to a tezos network */
-  final case class TezosConfiguration(network: String, depth: Depth = Newest, nodeConfig: TezosNodeConfiguration) extends PlatformConfiguration
+  final case class TezosConfiguration(network: String, depth: Depth = Newest, nodeConfig: TezosNodeConfiguration, headHash: Option[String] = None) extends PlatformConfiguration
```
ivanopagano: To be honest, I had the feeling that `depth` didn't belong here, and more-so now that we're adding `headHash`.
I'll explain myself: the Tezos configuration was meant to contain all values needed to define how to connect to the tezos nodes.
I would argue that `LorreConfiguration` is a best choice here, since it keeps all relevant values concerned with how Lorre is executed, independently of the platform it connects to.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -325,19 +326,24 @@ class TezosNodeOperator(val node: TezosRPCInterface, val network: String, batchC
...
-      head =>
-        val headLevel = head.data.header.level
-        val headHash = head.data.hash
-        val minLevel = depth.fold(1)(d => max(1, headLevel - d + 1))
-        val pagedResults = partitionBlocksRanges(minLevel to headLevel).map(
-          page => getBlocks((headHash, headLevel), page)
-        )
-        (pagedResults, headLevel - minLevel + 1)
-    }
+  def getLatestBlocks(depth: Option[Int] = None, headHash: Option[String] = None): Future[PaginatedBlocksResults] = {
```
ivanopagano: We might want `headHash: Option[BlockHash]` to use a better-typed value.

Consider also that you will find the `blockHeadHash` constant in TezosTypes as a useful default in that case, with no need for an `Option`-wrapped argument

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -325,19 +326,24 @@ class TezosNodeOperator(val node: TezosRPCInterface, val network: String, batchC
...
+    headHash
+      .map(it => getBlock(BlockHash(it)))
+      .getOrElse(getBlockHead())
+      .map {
+        head =>
```
ivanopagano: This should better be renamed, to avoid confusion.
Something like `reference`, `max`, `from`, ...

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -15,8 +15,8 @@ object TezosTypes {
     private val operationGroups = GenLens[Block](_.operationGroups)
     private val operations = GenLens[OperationsGroup](_.contents)
 
-    private val origination = Prism.partial[Operation, Origination]{case it: Origination => it}(identity)
-    private val transaction = Prism.partial[Operation, Transaction]{case it: Transaction => it}(identity)
+    private val origination = GenPrism[Operation, Origination]
```
ivanopagano: nice catch! 👌

src/test/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperatorTest.scala
```diff
@@ -136,6 +136,63 @@ class TezosNodeOperatorTest extends FlatSpec with MockFactory with Matchers with
...
+    //given
+    val tezosRPCInterface = stub[TezosRPCInterface]
+    val blockResponse = Future.successful(TezosResponseBuilder.blockResponse)
+    val operationsResponse = Future.successful(TezosResponseBuilder.operationsResponse)
+    val votesPeriodKind = Future.successful(TezosResponseBuilder.votesPeriodKind)
+    val votesQuorum = Future.successful(TezosResponseBuilder.votesQuorum)
+    val votesProposal = Future.successful(TezosResponseBuilder.votesProposal)
+
+    (tezosRPCInterface.runAsyncGetQuery _)
+      .when("zeronet", "blocks/BLJKK4VRwZk7qzw64NfErGv69X4iWngdzfBABULks3Nd33grU6c~")
```
ivanopagano: Would it make sense to use any different block hash for testing this call, as to make it plainly different from the one that refers to the "head" block?

src/main/scala/tech/cryptonomic/conseil/config/LorreAppConfig.scala
dorians: done

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
dorians: done

src/test/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperatorTest.scala
dorians: I don't get it. Looking at builders this is a "head" block only by convention.

ivanopagano: What I meant is that looking at these two test cases, it's not clear what they are doing differently. I'm always more worried about the fact that reading the code makes the logic plain to see, than the code works itself.
Of course when doing reviews it strikes out immediately if something is not so evident.

In the end I'm just highlighting the fact that it took me some effort to identify any difference in the test code, as to check if there might've been some error in the logic.

PR #311, opened 2019-04-01
#256 bytes support

dorians:
During tests against ``mainnet`` I found following structure which wasn't handled by Michelson parser:
```
{
  "prim": "PUSH",
  "args": [
    {
      "prim": "bytes"
    },
    {
      "bytes": "0500"
    }
  ]
}
```

This PR adds support for this data type.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/michelson/parser/JsonParser.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/michelson/parser/JsonParser.scala
@@ -89,6 +89,16 @@ object JsonParser {
   }
 
   /*
+   * Wrapper for bytes constant
+   *
+   * {"bytes": "0500"}
+   *
+   * */
+  case class JsonBytesConstant(bytes: String) extends JsonExpression {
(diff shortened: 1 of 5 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/michelson/parser/JsonParser.scala
```diff
@@ -88,6 +88,16 @@ object JsonParser {
     override def toMichelsonExpression = MichelsonStringConstant(string)
   }
 
+  /*
+   * Wrapper for bytes constant
+   *
+   * {"bytes": "0500"}
+   *
+   * */
+  case class JsonBytesConstant(bytes: String) extends JsonExpression {
```
ivanopagano: Would it be useful to wrap the `bytes` String in a Value class and only parse formally correct byte values with a custom parser/decoder, as we do with JsonDecoders for TezosTypes?

dorians: I was thinking about it but decided not to do it because it would require guessing validation rules and since I have only a few examples, the guessing might not match the reality.

PR #314, opened 2019-04-03
Conseil cleanup dependencies

ivanopagano:
Bumps all stale versions in the dependencies, and "excludes" all conflicting cases.

- The change opened the opportunity to remove some custom code for the endpoints definitions and to make some return value more precisely typed.
- There were conflicting mix-in if we tried to define the open-api main file and to expose it as json in the same endpoint interpreter, so now we have splitted the publishing in a separately-defined route interpreter

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/routes/openapi/DataJsonSchemas.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/routes/openapi/DataJsonSchemas.scala
@@ -10,31 +10,31 @@ import tech.cryptonomic.conseil.tezos.Tables.{AccountsRow, BlocksRow, OperationG
 trait DataJsonSchemas extends generic.JsonSchemas {
 
   /** API query schema */
-  implicit def queryRequestSchema: JsonSchema[ApiQuery] =
+  implicit lazy val queryRequestSchema: JsonSchema[ApiQuery] =
     genericJsonSchema[ApiQuery]
 
   /** Query predicate schema */
-  implicit def queryPredicateSchema: JsonSchema[Predicate] =
+  implicit lazy val queryPredicateSchema: JsonSchema[Predicate] =
     genericJsonSchema[Predicate]
 
   /** Query ordering operation schema */
-  implicit def queryOrderingOperationSchema: JsonSchema[OperationType.Value] =
+  implicit lazy val queryOrderingOperationSchema: JsonSchema[OperationType.Value] =
     enumeration(OperationType.values.toSeq)(_.toString)
 
   /** Query ordering schema */
-  implicit def queryOrderingSchema: JsonSchema[QueryOrdering] =
+  implicit lazy val queryOrderingSchema: JsonSchema[QueryOrdering] =
     genericJsonSchema[QueryOrdering]
 
   /** Query ordering direction schema */
-  implicit def queryOrderingDirectionSchema: JsonSchema[OrderDirection.Value] =
+  implicit lazy val queryOrderingDirectionSchema: JsonSchema[OrderDirection.Value] =
     enumeration(OrderDirection.values.toSeq)(_.toString)
 
   /** Query output schema */
-  implicit def queryOutputSchema: JsonSchema[OutputType.Value] =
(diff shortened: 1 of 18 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -327,9 +331,7 @@ object ApiOperations extends DataOperations with MetadataOperations {
 
     dbHandle.run(fetchOperation).map{
       accounts =>
-        accounts.headOption.map { account =>
-          Map("account" -> account)
-        }
+        accounts.headOption.map(AccountResult(_))
```
piotrkosecki: why not `accounts.headOption.map(AccountResult)`?

ivanopagano: Will check if the inference works, maybe I didn't even try this time (it failed so many times in my past experience 😢 )

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -423,6 +422,9 @@ class TezosNodeSenderOperator(override val node: TezosRPCInterface, network: Str
   import com.muquit.libsodiumjna.{SodiumKeyPair, SodiumLibrary, SodiumUtils}
   import TezosNodeOperator._
 
+  /** Type representing Map[String, Any] */
```
piotrkosecki: just out of curiosity - can you describe your approach to the type aliases regarding where to put them? I always tried to put them in some object with `*Types` in name because at the beginning when I join I saw that way being used somewhere in the project regarding case classes. I have no clear preferences, but maybe we should think of extracting models/types to separate packages, or maybe even objects. WDYT?

src/main/scala/tech/cryptonomic/conseil/routes/openapi/DataJsonSchemas.scala
```diff
@@ -1,51 +1,57 @@
...
+    xmapJsonSchema[Long, java.sql.Timestamp](
+      implicitly[JsonSchema[Long]],
+      millisFromEpoch => new java.sql.Timestamp(millisFromEpoch),
+      ts => ts.getTime
+    )
+  }
 
   /** Blocks row schema */
-  implicit def blocksRowSchema: JsonSchema[BlocksRow] =
+  implicit lazy val blocksRowSchema: JsonSchema[BlocksRow] =
```
ba205: what was the impetus behind making these values lazy?

ivanopagano: They've always been (they where methods).
The idea was to compute them only once, as I expect the generically derived implementation might carry some extra weight.
I made them lazy to keep the lazy semantic nonetheless

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ivanopagano: I certainly see your point here, though I didn't get the naming convention before.
I must admit that "Data" never sounds like a great identifier to me, I tend to avoid any generic naming like "manager", "process", "data", "map", "value" as they can mean all and nothing at the same time.
So I might've missed the idea that `DataTypes` meant type definitions for the chain.DataX objects.
I agree that a consistent naming rule like this makes a lot of sense.

On the other hand, I simply moved the type alias out of the main DataTypes object, because after the refactor, it's only used in two specific contexts (i.e. this file and another).
I wanted to somehow "demote" it from being a globally shared type name, to a local shortcut.
The reason behind my choice was to reduce the chances of any-typed definitions to spread in other parts of the code, unless where strictly necessary.

Yet I'm not too strong on this position, so if you find good reasons for it to be restored as a globally shared public type definition, I have no issue to put it back there

piotrkosecki: No, I don't have any good reason to move it. I just wanted to know about your position about this for future as I'm introducing aliases in my PR 😄

PR #318, opened 2019-04-04
315 - store a storage field

dorians:
We do need to store storage in PostgreSQL.

add the following column to the `operations` table

```sql
    storage character varying,
```

None


piotrkosecki:
Please mention in the PR description changes to the schema/configuration like @ivanopagano did here https://github.com/Cryptonomic/Conseil/pull/288. Otherwise LGTM

doc/conseil.sql
```diff
@@ -33,6 +33,8 @@ COMMENT ON EXTENSION plpgsql IS 'PL/pgSQL procedural language';
 -- Name: truncate_tables(character varying); Type: FUNCTION; Schema: public; Owner: -
 --
 
+DROP FUNCTION IF EXISTS public.truncate_tables(username character varying);
```
ivanopagano: Out of curiosity: did you add this manually?

AFAIK this script is not supposed to be run on existing instances. I'm not sure if we may prefer for it to fail if that happens.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -398,17 +399,17 @@ class TezosNodeOperator(val node: TezosRPCInterface, val network: String, batchC
 
   val UNPARSABLE_CODE_PLACEMENT = "Unparsable code: "
 
-  private def toMichelsonScript[T <: MichelsonElement:Parser](json: Any): String = {
+  private def toMichelsonScript[T <: MichelsonElement](json: Any)(implicit parser: Parser[T], tag: ClassTag[T]): String = {
```
ivanopagano: You can add more than one type bound, if you like to, or mix them

`toMichelsonScript[T <: MichelsonElement: Parser](json: Any)(implicit tag: ClassTag[T]): String`

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -398,17 +399,17 @@ class TezosNodeOperator(val node: TezosRPCInterface, val network: String, batchC
...
     Some(json).collect {
       case t: String => convert[T](t)
       case t: Micheline => convert[T](t.expression)
     } match {
       case Some(Right(value)) => value
       case Some(Left(t)) =>
-        logger.error(s"Error during converting Michelson format: $json", t)
+        logger.error(s"Error during converting Michelson format to (${tag.runtimeClass}): $json", t)
```
ivanopagano: Detail: I think the sentence is wrong, but may be mistaken

"Error while converting ..." or "Error during conversion of ..." should be correct?

doc/conseil.sql
dorians: Yes, I added this manually because I needed to remove this function manually each time I created a new database.

PR #320, opened 2019-04-04
Conseil 301: Interleave the account processing with block batches

### LINKED ISSUE #301: Lorre should print out processing information for accounts
Lorre should print out processing information for accounts just as it does for blocks. Especially after a full-chain download, it's difficult to see how far account fetching has progressed. Here is some example output:

```
18:57:16.997 [lorre-system-akka.actor.default-dispatcher-14] INFO  tech.cryptonomic.conseil.Lorre$ - Completed processing 99.52% of total requested blocks
18:57:16.997 [lorre-system-akka.actor.default-dispatcher-14] INFO  tech.cryptonomic.conseil.Lorre$ - Estimated time to finish is around 2 minutes
18:58:22.674 [lorre-system-akka.actor.default-dispatcher-1453] INFO  tech.cryptonomic.conseil.Lorre$ - Wrote 395 blocks to the database, checkpoint stored for account updates
18:59:45.535 [lorre-system-akka.actor.default-dispatcher-59] INFO  tech.cryptonomic.conseil.Lorre$ - Wrote 0 voting data records to the database
18:59:45.535 [lorre-system-akka.actor.default-dispatcher-14] INFO  tech.cryptonomic.conseil.Lorre$ - Completed processing 100.00% of total requested blocks
18:59:45.535 [lorre-system-akka.actor.default-dispatcher-14] INFO  tech.cryptonomic.conseil.Lorre$ - Processing latest Tezos data for updated accounts...
18:59:45.563 [lorre-system-akka.actor.default-dispatcher-14] INFO  tech.cryptonomic.conseil.Lorre$ - Selecting all accounts touched in the checkpoint table, this might take a while...
```

The time is now 00:45 the day after but it's impossible to say how account fetching is doing. The only evidence I see for active processing is Postgres is maxing out a CPU core.

ivanopagano:
fix #301 
Instead of handling accounts after all blocks are synced, now we do it after each page of new blocks is done.

This way, for long-running syncs -- as in bootstrapping a new db -- we process accounts more often than actually needed, overwriting the results as each new block page is synced.
Yet the marginal impact should be relatively minor, considered the upsides:
1. querying for all updated accounts is now a reasonable operation, given the size is limited to how many accounts might be updated for a single block page (500 blocks, but configurable)
2. after each block batch is done, accounts are processed too, so an interrupted execution can probably leave the database consistent with the block level reached up-to-failure, including account data
3. lemma of 2 is that the accounts are updated in real-time while lorre is executing, which might give a visual cue to anyone running a client, instead of an empty balance
4. the progress shown by the logs can now be considered valid for the whole sync operation, which will complete soon after blocks are done 100%, and will give an ETA on the whole completion

Marginal Note: I left the paged processing for accounts too, considering that
- they can be a consistent number for each block batch too, probably growing with more recent blocks
- the block paging size is configurable, so the accounts loaded per step can grow too

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -79,7 +79,6 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
     val noOp = Future.successful(())
     val processing = for {
       _ <- processTezosBlocks()
-      _ <- processTezosAccounts()
       _ <-
         if (iteration % lorreConf.feeUpdateInterval == 0)
           FeeOperations.processTezosAverageFees(lorreConf.numberOfFeesAveraged)
@@ -126,6 +125,9 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
     * Additionally stores account references that needs updating, too
     */
   private[this] def processTezosBlocks(): Future[Done] = {
+    import cats.instances.future._
+    import cats.syntax.apply._
+
     logger.info("Processing Tezos Blocks..")
 
     val blockPagesToSynchronize = tezosConf.depth match {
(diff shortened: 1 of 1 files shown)


dorians:
Good one :). I'll be easier to debug too.

PR #321, opened 2019-04-05
Aggregation fixes + doc update

piotrkosecki:
Fixed issues with not sorting correctly by the field not existing in the aggregation and with not ordering by the aggregated field, added tests for those cases and updated the doc

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
@@ -75,11 +75,11 @@ object DatabaseUtil {
         * @param ordering list of QueryOrdering to add
         * @return new SQLActionBuilder containing ordering statements
         */
-      def addOrdering(ordering: List[QueryOrdering]): SQLActionBuilder = {
+      def addOrdering(ordering: List[QueryOrdering], aggregation: Option[Aggregation]): SQLActionBuilder = {
         val queryOrdering = if (ordering.isEmpty) {
           List.empty
         } else {
-          List(makeOrdering(ordering))
+          List(makeOrdering(ordering, aggregation))
         }
         concatenateSqlActions(action, queryOrdering:_*)
       }
@@ -101,7 +101,7 @@ object DatabaseUtil {
         */
       def addGroupBy(aggregation: Option[Aggregation], columns: List[String]): SQLActionBuilder = {
         val columnsWithoutAggregation = columns.filterNot(col => aggregation.exists(_.field == col))
-        if(columnsWithoutAggregation.isEmpty) {
+        if(aggregation.isEmpty) {
           action
         } else {
           concatenateSqlActions(action, makeGroupBy(columnsWithoutAggregation))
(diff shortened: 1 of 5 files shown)


src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
```diff
@@ -101,7 +101,7 @@ object DatabaseUtil {
         */
       def addGroupBy(aggregation: Option[Aggregation], columns: List[String]): SQLActionBuilder = {
         val columnsWithoutAggregation = columns.filterNot(col => aggregation.exists(_.field == col))
-        if(columnsWithoutAggregation.isEmpty) {
+        if(aggregation.isEmpty) {
```
ivanopagano: If the check is now on the optional argument I think we can make this a lot simpler
```scala
aggregation.fold(action) {
  aggregates =>
    val col... = columns.filterNot(_ == aggregates.field)
    concatenateSqlActions(....)
}
```

src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
```diff
@@ -144,8 +144,12 @@ object DatabaseUtil {
       * @param ordering list of ordering parameters
       * @return SQLAction with ordering
       */
-    def makeOrdering(ordering: List[QueryOrdering]): SQLActionBuilder = {
-      val orderingBy = ordering.map(x => s"${x.field} ${x.direction}").mkString(",")
+    def makeOrdering(ordering: List[QueryOrdering], aggregation: Option[Aggregation]): SQLActionBuilder = {
+      val aggregatedOrdering = ordering.map {
+        case ord if aggregation.map(_.field).contains(ord.field) => ord.copy(field = mapAggregationToSQL(aggregation.get.function, aggregation.get.field))
```
ivanopagano: Can do `aggregation.exists(_.field == ord.field)` if you like

ivanopagano: I might propose 
```scala
val orderingBy = ordering.map {
  ord =>
    val ordField =
      if (aggregation.exists(_.field == ord.field)) mapAggreagationToSQL(..., ...)
      else ord.field
    s"$ordField ${ord.direction}"
}.mkString(",")
```

IMO, this kind of approach reduces the need to "enter/exit" within Lists/Options content using multiple mappings.
It gets to the exact data needed and produces the end result where it's available, so that no extra transformation is done in the calling context if not needed.

src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
piotrkosecki: changed, I haven't thought of that :/

src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
piotrkosecki: changed according to your suggestion

PR #324, opened 2019-04-08
Aggregation and DateTime fixes

piotrkosecki:
There were two issues:

- all aggregation functions were only handling Numeric/DateTime fields, but `COUNT` function should be available for all the fields - after this PR it will be.

- predicates with DateTime field type were compared in the DB with the timestamp representation in milliseconds - after this PR predicate with DateTime type will be have `set` mapped to `Instant` and then to sting so it could be correctly recognised in DB.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
@@ -1,5 +1,7 @@
 package tech.cryptonomic.conseil.generic.chain
 
+import java.sql.Timestamp
+
 import tech.cryptonomic.conseil.generic.chain.DataTypes.AggregationType.AggregationType
 import tech.cryptonomic.conseil.generic.chain.DataTypes.OperationType.OperationType
 import tech.cryptonomic.conseil.generic.chain.DataTypes.OrderDirection.OrderDirection
@@ -19,19 +21,29 @@ object DataTypes {
   import cats.implicits._
   import io.scalaland.chimney.dsl._
 
+
   /** Type representing Map[String, Option[Any]] for query response */
   type QueryResponse = Map[String, Option[Any]]
-  /** Method checks if type can be aggregated */
-  def canBeAggregated(dataType:DataType, aggregationType: AggregationType): Boolean = {
-      if(aggregationType != AggregationType.count) {
-        Set(DataType.Decimal, DataType.Int, DataType.LargeInt, DataType.DateTime)(dataType)
-      } else true
-  }
-
   /** Default value of limit parameter */
   val defaultLimitValue: Int = 10000
   /** Max value of limit parameter */
   val maxLimitValue: Int = 100000
+
+  /** Replaces timestamp represented as Long in predicates with one understood by the SQL */
+  private def replaceTimestampInPredicates(entity: String, query: Query, tezosPlatformDiscoveryOperations: TezosPlatformDiscoveryOperations)
+    (implicit executionContext: ExecutionContext): Future[Query] = {
+    query.predicates.map { predicate =>
+      tezosPlatformDiscoveryOperations.getTableAttributesWithoutUpdatingCache(entity).map { maybeAttributes =>
+        maybeAttributes.flatMap { attributes =>
+          attributes.find(_.name == predicate.field).map {
+            case attribute if attribute.dataType == DataType.DateTime =>
+              predicate.copy(set = predicate.set.map(x => new Timestamp(x.toString.toLong).toInstant.toString))
+            case _ => predicate
+          }
+        }.toList
+      }
+    }.sequence.map(pred => query.copy(predicates = pred.flatten))
+  }
 
   /** Helper method for finding fields used in query that don't exist in the database */
   private def findNonExistingFields(query: Query, entity: String, tezosPlatformDiscovery: TezosPlatformDiscoveryOperations)
@@ -68,6 +80,13 @@ object DataTypes {
         }
       }
       .map(_.flatten.collect { case (false, fieldName) => InvalidAggregationFieldForType(fieldName) }.toList)
+  }
+
+  /** Method checks if type can be aggregated */
+  def canBeAggregated(dataType: DataType, aggregationType: AggregationType): Boolean = {
+    if (aggregationType != AggregationType.count) {
+      Set(DataType.Decimal, DataType.Int, DataType.LargeInt, DataType.DateTime)(dataType)
+    } else true
   }
 
   /** Trait representing query validation errors */
@@ -149,15 +168,16 @@ object DataTypes {
       val nonExistingFields = findNonExistingFields(query, entity, tezosPlatformDiscovery)
       val invalidTypeAggregationField = findInvalidAggregationTypeFields(query, entity, tezosPlatformDiscovery)
 
-      for {
+      val result = for {
         invalidNonExistingFields <- nonExistingFields
         invalidAggregationFieldForTypes <- invalidTypeAggregationField
       } yield {
         invalidNonExistingFields ::: invalidAggregationFieldForTypes match {
-          case Nil => Right(query)
-          case wrongFields => Left(wrongFields)
+          case Nil => Right(replaceTimestampInPredicates(entity, query, tezosPlatformDiscovery))
+          case wrongFields => Left(Future.successful(wrongFields))
         }
       }
+      result.map(_.bisequence).flatten
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -58,13 +75,20 @@ object DataTypes {
...
           }
         }
       }
       .map(_.flatten.collect { case (false, fieldName) => InvalidAggregationFieldForType(fieldName) }.toList)
   }
 
+  /** Method checks if type can be aggregated */
+  def canBeAggregated(dataType: DataType, aggregationType: AggregationType): Boolean = {
+    if (aggregationType != AggregationType.count) {
+      Set(DataType.Decimal, DataType.Int, DataType.LargeInt, DataType.DateTime)(dataType)
```
ivanopagano: _Tag: performance_
I would try to make this result a constant, since it will be rebuilt each time the method is called.

Alternatively the method could be somehow converted to a function value?

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -19,15 +21,30 @@ object DataTypes {
...
-  lazy val canBeAggregated: DataType => Boolean = Set(DataType.Decimal, DataType.Int, DataType.LargeInt, DataType.DateTime)
   /** Default value of limit parameter */
   val defaultLimitValue: Int = 10000
   /** Max value of limit parameter */
   val maxLimitValue: Int = 100000
 
+  /** Replaces timestamp represented as Long in predicates with one understood by the SQL */
+  private def replaceTimestampInPredicates(entity: String, query: Query, tezosPlatformDiscoveryOperations: TezosPlatformDiscoveryOperations)
+    (implicit executionContext: ExecutionContext): Future[Query] = {
+    query.predicates.map { predicate =>
```
ivanopagano: _Tag: possible improvements_
I think some form of optic/lens can be written to ligthen this complex nested modification

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -19,15 +21,30 @@ object DataTypes {
...
+      tezosPlatformDiscoveryOperations.getTableAttributesWithoutUpdatingCache(entity).map { maybeAttributes =>
+        maybeAttributes.flatMap { attributes =>
+          attributes.find(_.name == predicate.field).map {
+            case attribute if attribute.dataType == DataType.DateTime =>
+              predicate.copy(set = predicate.set.map(x => new Timestamp(x.toString.toLong).toInstant.toString))
```
ivanopagano: Are we sure that in both places we use `.toString`, the result would always be the one that we have now?

- `x.toString` seems quite plausible
- I'm not too keen on the `.toInstant.toString`, I'm worried that 
  - the default toString of `Instant` might change in the future
  - we're not considering timezone-independence of this code (in particular wrt. the database's own TZ)

for the latter I can think of using an explicit iso format defined in java.time that would
- make the assumptions and conventions clear
- take the timezone into account

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -19,15 +21,30 @@ object DataTypes {
...
+            case _ => predicate
+          }
+        }.toList
+      }
+    }.sequence.map(pred => query.copy(predicates = pred.flatten))
```
ivanopagano: not sure what the last `flatten` is achieving, could this automatically ignore any formally invalid datetime constraint in the set?

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -144,15 +168,16 @@ object DataTypes {
...
 
-      for {
+      val result = for {
         invalidNonExistingFields <- nonExistingFields
         invalidAggregationFieldForTypes <- invalidTypeAggregationField
       } yield {
         invalidNonExistingFields ::: invalidAggregationFieldForTypes match {
-          case Nil => Right(query)
-          case wrongFields => Left(wrongFields)
+          case Nil => Right(replaceTimestampInPredicates(entity, query, tezosPlatformDiscovery))
```
ivanopagano: Suggestion: replace timestamps directly where `query` is defined, and make the val `lazy`

ivanopagano: else you can simply put the `replaceTimestamp...` in the for-comp, to remove the extra layer to peel out with `bisequence` down here

src/test/scala/tech/cryptonomic/conseil/tezos/DataTypesTest.scala
```diff
@@ -197,6 +215,60 @@ import scala.concurrent.ExecutionContext.Implicits.global
 
       result.futureValue.left.get should contain theSameElementsAs List(InvalidAggregationField("invalid"), InvalidAggregationFieldForType("invalid"))
     }
+
+    "correctly validate aggregation when COUNT function is used" in {
```
ivanopagano: Not sure if this is confirming that the aggregation is valid or if the validate function works, the test phrasing is not too clear for me.
> Correctly validate
seems to say that we're testing that validate performs correctly.

May I suggest "successfully validates aggregation for any dataType when ..."?

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: yes, I started with that, but I didn't like how it looked :P I'll change that ;)

ivanopagano: It's not that critical, anyway

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: I've got `List[Future[List[Predicate]]]`, so I used sequence to get `Future[List[List[Predicate]]]` and I needed to flatten it

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: you are right, I was overthinking it so to be safe I wanted to make `replaceTimestampInPredicates` a last call so if there were validation errors this method would not have a chance to fail, but it still wouldn't

src/test/scala/tech/cryptonomic/conseil/tezos/DataTypesTest.scala
piotrkosecki: changed

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: don't know, never used lens, I'll need to read about it

ivanopagano: You can leave that for a later refactor, possibly adding a generic task to say that

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
ivanopagano: I lost myself there, thanks

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -147,15 +171,25 @@ object DataTypes {
...
-          case Nil => Right(query)
+          case Nil => Right(updatedQuery)
           case wrongFields => Left(wrongFields)
         }
       }
+
     }
   }
 
+  private def dateToIso(date: Date): String = {
```
ivanopagano: `java.time` package provides ready-made-apis for this kind of things (I encourage everyone to study those)

```scala
import java.time.{Instant, ZoneOffset}
import java.time.format.DateTimeFormatter.ISO_INSTANT

def formatToIso(epochMillis: Long): String =
  Instant.ofEpochMilli(epochMillis)
    .atZone(ZoneOffset.UTC)
    .format(ISO_INSTANT)
```

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
ivanopagano: I outlined a different way to do that, using java.time apis

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: ok, done

PR #331, opened 2019-04-10
Fixes for between operation being used on the datetime fields and fix for transforming milis to Timestamp format

piotrkosecki:
fixed behaviour of `between` operation, so it can be used on the DateTime fields, changed the datetime formatting because of inconsistency I've found between results from DB and trying to query them. Probably I need to do some more fixes as timezones in my local container and in the Travis do not match. 
Probably we should think of using `timestamp with timezone` in the DB types.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
@@ -1,8 +1,6 @@
 package tech.cryptonomic.conseil.generic.chain
 
 import java.sql.Timestamp
-import java.time.format.DateTimeFormatter
-import java.time.ZoneOffset
 
 import tech.cryptonomic.conseil.generic.chain.DataTypes.AggregationType.AggregationType
 import tech.cryptonomic.conseil.generic.chain.DataTypes.OperationType.OperationType
@@ -47,7 +45,7 @@ object DataTypes {
         maybeAttributes.flatMap { attributes =>
           attributes.find(_.name == predicate.field).map {
             case attribute if attribute.dataType == DataType.DateTime =>
-              predicate.copy(set = predicate.set.map(x => formatToIso(x.toString.toLong)))
+              predicate.copy(set = predicate.set.map(x => new Timestamp(x.toString.toLong).toString))
             case _ => predicate
           }
         }.toList
@@ -55,10 +53,6 @@ object DataTypes {
     }.sequence.map(pred => query.copy(predicates = pred.flatten))
   }
 
-  /** Method formatting millis to ISO format */
-  def formatToIso(epochMillis: Long): String = {
-    DateTimeFormatter.ISO_INSTANT.format(new Timestamp(epochMillis).toInstant)
-  }
   /** Helper method for finding fields used in query that don't exist in the database */
   private def findNonExistingFields(query: Query, entity: String, tezosPlatformDiscovery: TezosPlatformDiscoveryOperations)
     (implicit ec: ExecutionContext): Future[List[QueryValidationError]] = {
--- a/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
@@ -57,11 +57,7 @@ object DataTypes {
 
   /** Method formatting millis to ISO format */
   def formatToIso(epochMillis: Long): String = {
-    // time needs to be adjusted with toLocalDateTime
-    new Timestamp(epochMillis)
-      .toLocalDateTime
-      .atZone(ZoneOffset.UTC)
-      .format(DateTimeFormatter.ISO_INSTANT)
+    DateTimeFormatter.ISO_INSTANT.format(new Timestamp(epochMillis).toInstant)
   }
   /** Helper method for finding fields used in query that don't exist in the database */
   private def findNonExistingFields(query: Query, entity: String, tezosPlatformDiscovery: TezosPlatformDiscoveryOperations)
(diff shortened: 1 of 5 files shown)


piotrkosecki:
Please don't merge it yet, because there are inconsistencies between running tests locally and on Travis.

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -55,11 +56,9 @@ object DataTypes {
   }
 
   /** Method formatting millis to ISO format */
-  def formatToIso(epochMillis: Long): String =
-    Instant.ofEpochMilli(epochMillis)
-      .atZone(ZoneOffset.UTC)
-      .format(ISO_INSTANT)
-
+  def formatToIso(epochMillis: Long): String = {
+    DateTimeFormatter.ISO_INSTANT.format(new Timestamp(epochMillis).toInstant)
```
ivanopagano: I can't figure out why passing through Timestamp should change things, wrt. creating the Instant.

Did you see a difference?

src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
```diff
@@ -186,7 +186,7 @@ object DatabaseUtil {
     /** maps operation type to SQL operation */
     private def mapOperationToSQL(operation: OperationType, inverse: Boolean, vals: List[String]): SQLActionBuilder = {
       val op = operation match {
-        case OperationType.between => sql"BETWEEN #${vals.head} AND #${vals(1)}"
+        case OperationType.between => sql"BETWEEN '#${vals.head}' AND '#${vals(1)}'"
```
ivanopagano: Will this behave as expected when comparing other columns types?

Will it not actually compare number lexicographically?

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: I think I got to the bottom of my issue and the solution was to use just `new Timestamp(_).toString`: 
- database does not care about the time zone
- instant does
So what problems did I have there - pick 2 out of three - Travis, tests, running locally 😄. Two of them worked and third not when I tried to configure it with `Instant`. `Timestamp` is a generic solution.

src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
piotrkosecki: I tested it with `<` and `>` and it works correctly.

piotrkosecki:
I think it is ready for review/merge. I don't understand how coveralls work. I've removed some lines and changed some + added test for the case which failed and it thinks that coverage decreased...

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
ivanopagano: so, we assume that
- the DB will always receive timestamps from the TZ that we choose to use to create the value
- it returns the timestamp as-is, with no notion of TZ
- if we read an Instant from the Timestamp, then something bad happens?

src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
ivanopagano: What about writing a test for between with a numeric type, just to be sure?

Sorry but I don't follow you when you say that you tested on `<`, `>`

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: First of all, we don't have much control on what we will receive from DB.
Next if we read Instant from the Timestamp and try to use it against DB we will get incorrect results. Instant would need to be used with TZ which changes - best example is when you try running tests locally and on Travis. 
We are relying on the consistent usage of timestamp across Conseil and Lorre. Maybe it would be better to have `Timestamp with timezone` in the DB. But it would require using Instant everywhere.

src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
piotrkosecki: I've just added test for your case 😉

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
ivanopagano: What I mean is that consistency between different environment should be guaranteed exactly by using timezones correctly.
When Lorre stores a timestamp in DB, it should be expressed relative to a standard timezone (UTC being the most common one).
Hence every time we read timestamps from there we should know that they need be expressed as epoch relative to that timezone.
It's reasonable that running a query from a different timezone returns different results, unless we normalise the query input too...

If you're in Warsaw and ask for 14.00-15.00, would you expect to have data timestamped on db with 14.00-15.00 or 12.00-13.00 ?

piotrkosecki: But we are not using timezones explicitly anywhere, so I don't see a reason to use it here. I agree with your argument that we probably should, that's why I mentioned changing DB columns to timestamp with timezone. My understanding is that frontend will be handling time zones thats why Mike told on Riot that he expects timestamp in millis.

ivanopagano: ok, I agree that, as long as the client knows what it's asking for (in terms of date-strings) we're cool

ivanopagano:
> I think it is ready for review/merge. I don't understand how coveralls work. I've removed some lines and changed some + added test for the case which failed and it thinks that coverage decreased...

You can look into the link shown in the failing coverage, there you get detailed reporting on which code files have decreased coverage and how much

vishakh:
I fixed the tag in `master` but this build is still failing.

piotrkosecki:
I'm closing this PR as I created second branch.... this one seems corrupted somehow.

PR #332, opened 2019-04-11
Fixes for between operation and timestamp

piotrkosecki:
Apparently previous branch was somehow corrupted and it couldn't generate correct version in `BuildInfo`


piotrkosecki:
I'll probably need help from @ivanopagano with this, master has correct tagging, when I create a branch it breaks...

PR #335, opened 2019-04-15
Conseil 330: operation results storage size

### LINKED ISSUE #330: Consume paid_storage_size_diff field with operations
Invoking and deploying a smart contract sometimes requires a payment to be made to increase the storage size. This information is listed under operation metadata with the name `paid_storage_size_diff`. This information is not recorded in Conseil. Can we please add? Also adding `storage_size` would be very nice.

Here is an example of an origination operation. Note that **some** transaction operations have the same field.

**Operation Group ID:** oot6P2D7TSKbwjtDbPBsETXvJnZCWkRAk3ZqfwPb1BrnKUz3yJ5

The result Tezos Node returns:

The result in Conseil:

Screen Shot 2019-04-09 at 5 04 04 PM

ivanopagano:
closes #330
Store the missing values, when available, for operations

## Required changes to DB Schema

the new _columns_ are added to the `operations` table

```sql
   storage_size numeric,
   paid_storage_size_diff numeric,
 ```

### CODE REVIEW

File: src/test/scala/tech/cryptonomic/conseil/util/ConfigUtilTest.scala [hunk 1]
```diff
@@ -5,6 +5,37 @@ import com.typesafe.config.ConfigFactory
...
+  "the config.Natural matcher" should {
+
+    "match a valid positive integer string" in {
+      "10" match {
+        case Natural(value) => value shouldBe 10
+        case _ => fail("the matcher didn't correcly identify an integer")
+      }
+    }
+
+    "refuse a zero integer string" in {
```
piotrkosecki: why "0" is not a `Natural`?

ivanopagano: Because the definition is not unambiguous and I needed a name to say the positive integers (the zero is already handled in a separate case handler for our usage of this method)

PR #336, opened 2019-04-15
316 - supplementary metadata

### LINKED ISSUE #316: Conseil should accept supplementary metadata from a file
Conseil should accept supplementary content in files for its metadata routes, e.g. `displayName` for the `attributes` routes. The metadata can be maintained in a file called `attributes.xml` which could have the following format:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<platforms>
	<platform name="tezos" visible="true" displayName="Tezos" description="The Tezos blockchain">
		<networks>
			<network name="alphanet" visible="true" displayName="Alphanet" description="Official Tezos testnet">
				<entities>
					<entity name="blocks" visible="true" displayName="Blocks" description="Blocks">
						<attributes>
							<attribute name="level" visible="true" displayName="Block level" description="Height of block in blockchain" />
						</attributes>
					</entity>
				</entities>
			</network>
		</networks>
	</platform>
</platforms>
```

Platforms, networks, entities and attributes should not be returned by the metadata routes **unless** they appear in this file and their corresponding `visible` flag is set to `true`. The schema returned by the metadata routes should be enhanced to accept all the new information in the XML file. 

An XML schema should be created for the above content and any provided file should be validated upfront for errors.

Conseil should also accept override metadata files as command-line parameters.

dorians:
This PR resolves #316.


src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
```diff
@@ -0,0 +1,71 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+class MetadataService(config: PlatformsConfiguration,
+                      tezosPlatformDiscoveryOperations: TezosPlatformDiscoveryOperations,
+                      metadataOverrides: MetadataOverridesConfiguration) {
+
+  def getPlatforms: List[Platform] = {
+    ConfigUtil
+      .getPlatforms(config)
+      .flatMap(platform => PlatformData(PlatformPath(platform.name), platform).apply(metadataOverrides))
```
ivanopagano: I see a general pattern here apparently, e.g. for a given platform `layer`(platform, network, entity, attribute)

`layer => LayerData(pathBuilder(layer.name), layer).apply(metadataOverrides)`

this makes me think that 
1. The "LayerData" is a short-lived storage of data that will be passed to a function call
2. often, if not always, `layer.name` is passed along with the `layer` to the `LayerData`, hence we could pass `layer` only

Given `2`, we can think of converting `LayerData(...).apply(overrides)` to a function of the same name that simply gets all arguments together, e.g.

```scala
object Data {
  def platform(platform: Platform, overrides: OverridesConf): Option[Platform] = {
    val path = PlatformPath(platform.name)
    when(overrides.isVisible(path)) {
      val nameOverride = overrides.platform(path).map(_.displayName)
      nameOverride.fold(platform){
        overridden => platform.copy(displayName = overridden)
      }
    }
  }
```
 evaluate the code above, if it makes sense.

I also approached the code to avoid copying the layer if not needed. The logic might or might not be easier to follow

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  def getAttributeValues(platform: String, network: String, entity: String, attribute: String)(implicit apiExecutionContext: ExecutionContext): Future[Option[List[String]]] = {
+    if (exists(NetworkPath(platform, network)))
+      tezosPlatformDiscoveryOperations.listAttributeValues(entity, attribute).map(Some(_))
+    else
+      successful(None)
+  }
+
+
+  def getAttributesValuesWithFilterRoute(platform: String, network: String, entity: String, attribute: String, filter: String)(implicit apiExecutionContext: ExecutionContext): Future[Option[List[String]]] = {
```
ivanopagano: Does this needs to be a separate method from `getAttributeValues` ?

Would an Optional param suffice?

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
```diff
@@ -0,0 +1,71 @@
...
+import tech.cryptonomic.conseil.generic.chain.PlatformDiscoveryTypes.{Attribute, Entity, Network, Platform}
+import tech.cryptonomic.conseil.metadata.Data.{AttributeData, EntityData, NetworkData, PlatformData}
+import tech.cryptonomic.conseil.tezos.TezosPlatformDiscoveryOperations
+import tech.cryptonomic.conseil.util.OptionUtil.when
+import tech.cryptonomic.conseil.util.ConfigUtil
+
+import scala.concurrent.Future.successful
+import scala.concurrent.{ExecutionContext, Future}
+
+class MetadataService(config: PlatformsConfiguration,
```
ivanopagano: I find the code in this class a little complicated, and I'm trying to figure out if 
- we could easily separate the override logic from the configuration reading
- the nesting of different maps/flatMaps might be simplified

I'll forego a deeper analysis now, as long as there will be tests to guarantee correctness of the code.
If some idea to simplify comes along, I'll be happy with that, but for now it's good to go for me, as long as it works.

src/main/scala/tech/cryptonomic/conseil/metadata/Path.scala
```diff
@@ -0,0 +1,31 @@
...
+  def addLevel(nextLabel: String): Path
+}
+
+case class EmptyPath() extends Path {
+  override def up: Path = this
+  override def addLevel(nextLabel: String): Path = PlatformPath(nextLabel)
+}
+
+case class PlatformPath(platform: String) extends Path {
+  override def up: EmptyPath = EmptyPath()
```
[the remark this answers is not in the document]
dorians: I was thinking about it but I was afraid that constructing those Paths will be difficult. But I changed it right now and it's not that bad :)

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
dorians: I must provide ``Path`` too, but I refactored it in a similar way.

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
dorians: Nice catch!

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
dorians: I refactored it a little bit. Hope you find it more clear now. I spend a lot of time on it. Mostly because there is a lot of types like ``Future[Option[List[...]]]``.

ivanopagano: Did you consider using some of cats' MonadT? Or `Nested`?

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
```diff
@@ -0,0 +1,63 @@
...
+  def getAttributeValues(platform: String, network: String, entity: String, attribute: String, filter: Option[String] = None)
+                        (implicit apiExecutionContext: ExecutionContext): Future[Option[List[String]]] = {
+    if (exists(NetworkPath(network, PlatformPath(platform))))
+      tezosPlatformDiscoveryOperations.listAttributeValues(entity, attribute, filter).map(Some(_))
+    else
+      successful(None)
+  }
+
+  private def exists(path: NetworkPath): Boolean = {
+    getNetworks(path.up).getOrElse(List()).exists(_.network == path.network)
```
ivanopagano: minor: using the `empty` method consistently should always get better results with type-inference, and avod temporary allocation of an empty wrapper

e.g. `List.empty`, `Option.empty`, `Set.empty`...

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  def getEntities(path: NetworkPath)(implicit apiExecutionContext: ExecutionContext): Future[Option[List[Entity]]] = {
+    if (exists(path) && metadataOverrides.isVisible(path))
+      tezosPlatformDiscoveryOperations
+        .getEntities
+        .embeddedMap(entity => overrideEntity(entity, path.addLevel(entity.name)))
+        .map(Some(_))
+    else
+      successful(None)
```
ivanopagano: This is a constant value, you might want to store it as a named `val` e re-use it, with the additional possible side-effect of improving readability, too?

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
ivanopagano: Doesn't look bad at all, 👍

src/main/scala/tech/cryptonomic/conseil/util/CollectionOps.scala
```diff
@@ -37,5 +39,13 @@ object CollectionOps {
     def onBounds(f: (T, T) => R): Option[R] = applyOnBounds(list)(f)
   }
 
+  implicit class ExtendedFutureWithOption[T](val value: Future[Option[List[T]]]) {
```
ivanopagano: Can it be made a Value Class? 🤔

src/main/scala/tech/cryptonomic/conseil/util/CollectionOps.scala
```diff
@@ -37,5 +39,13 @@ object CollectionOps {
...
+    def embeddedMap(function: T => Option[T])(implicit apiExecutionContext: ExecutionContext): Future[Option[List[T]]] =
+      value.map(_.map(_.flatMap(function(_))))
```
ivanopagano: not sure if it's better, 

```scala
import cats.implicits._

value.nested.map(_.flatMap(function)).value
```

src/main/scala/tech/cryptonomic/conseil/util/CollectionOps.scala
```diff
@@ -37,5 +39,13 @@ object CollectionOps {
     def onBounds(f: (T, T) => R): Option[R] = applyOnBounds(list)(f)
   }
 
+  implicit class ExtendedFutureWithOption[T](val value: Future[Option[List[T]]]) {
+    def embeddedMap(function: T => Option[T])(implicit apiExecutionContext: ExecutionContext): Future[Option[List[T]]] =
```
ivanopagano: minor: the name to me sounds like a data instance, possibly `mapNested`/`mapNested` could make it more evident?

src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
```diff
@@ -8,72 +8,53 @@ import endpoints.akkahttp
 import endpoints.algebra.Documentation
 import tech.cryptonomic.conseil.config.Platforms.PlatformsConfiguration
```
piotrkosecki: I think this import is no longer in use

src/main/scala/tech/cryptonomic/conseil/routes/PlatformDiscovery.scala
```diff
@@ -8,72 +8,53 @@ import endpoints.akkahttp
...
-    case ((platform, network, entity), _) =>
-      tezosPlatformDiscoveryOperations.getTableAttributes(entity).map { attributes =>
-        ConfigUtil.getNetworks(config, platform).find(_.network == network).flatMap { _ =>
-          attributes
-        }
-      }
+    case ((platform, network, entity), _) => metadataService.getTableAttributes(EntityPath(entity, NetworkPath(network, PlatformPath(platform))))
   }
 
   import cats.instances.future._
```
piotrkosecki: those can be removed as well

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
ivanopagano: Isn't the `PlatformPath` derived from the platform itself?

dorians: Yes, it is. But you cannot create a ``Path`` based on an ``Entity`` or an ``Attribute`` so I provide ``Path`` everywhere to be consistent.

src/main/resources/reference.conf
```diff
@@ -143,4 +143,2022 @@ akka {
...
+    visible: true
+    networks {
+      mainnet {
+        visible: true,
+        entities {
+          accounts {
+            visible: true,
+            attributes {
+              account_id {
+                visible: true
```
anonymoussprocket: Missing `description` field as specified in #316

src/main/resources/reference.conf
```diff
@@ -143,4 +143,2022 @@ akka {
...
+              block_id {
+                visible: true
+              }
+            }
+          }
+          operations {
+            visible: true,
+            attributes {
+              operation_id: {
+                visible: true
```
anonymoussprocket: visible should be false here

dorians: For ``operation_id`` for all networks, I assume?

src/main/resources/reference.conf
dorians: This PR is already approved, so I added this to https://github.com/Cryptonomic/Conseil/pull/344 to simplify code review.

vishakh:
@dorians The PR overall looks good. However, as a practice, we leave ScalaDoc comments on all classes and functions. Could you please add documentation for all new classes and functions?

dorians:
> @dorians The PR overall looks good. However, as a practice, we leave ScalaDoc comments on all classes and functions. Could you please add documentation for all new classes and functions?

@vishakh: done

PR #337, opened 2019-04-16
Conseil 322-1: parametric tezos rpc

ivanopagano:
### It is not meant to be merged as-is but to provide material for progressive review of the whole refactor

The "pluggable-backend" approach will make it easier to try streaming designs, based on different implementations than akka-stream and plain scala futures
The long-term plan is to open a discussion to evaluate the switch to cats-effects/monix/fs2

---
The current step introduces a type class design for the Tezos RPC calls, enabling a simpler interface that can accommodate different implementations.
This makes it easy to create sync, async, streaming rpc calls.

We define RPC callers for both `Future` and `akka.stream`-`Source` (the latter supports only GET calls actually)

The future rpc is used directly by Node Operator to individually fetch data from single endpoints.
The streaming rpc is now internally required by the async DataFetchers, providing multiple endpoint calls in the NodeOperator.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -4,7 +4,8 @@ import akka.actor.ActorSystem
 import akka.Done
 import mouse.any._
 import com.typesafe.scalalogging.LazyLogging
-import tech.cryptonomic.conseil.tezos.{TezosTypes, FeeOperations, ShutdownComplete, TezosErrors, TezosNodeInterface, TezosNodeOperator, TezosDatabaseOperations => TezosDb}
+import tech.cryptonomic.conseil.tezos.{TezosTypes, FeeOperations, ShutdownComplete, TezosErrors, TezosNodeOperator, TezosDatabaseOperations => TezosDb}
+import tech.cryptonomic.conseil.tezos.TezosRemoteInstances.Akka.RemoteContext
 import tech.cryptonomic.conseil.tezos.TezosTypes.BlockAccounts
 import tech.cryptonomic.conseil.io.MainOutputs.LorreOutput
 import tech.cryptonomic.conseil.util.DatabaseUtil
@@ -35,6 +36,7 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
   //the dispatcher is visible for all async operations in the following code
   implicit val system: ActorSystem = ActorSystem("lorre-system")
   implicit val dispatcher = system.dispatcher
+  implicit val nodeContext = RemoteContext(tezosConf, callsConf, streamingClientConf)
 
   //how long to wait for graceful shutdown of system components
   val shutdownWait = 10.seconds
@@ -43,15 +45,14 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
   sys.addShutdownHook(shutdown())
 
   lazy val db = DatabaseUtil.db
-  val tezosNodeOperator = new TezosNodeOperator(new TezosNodeInterface(tezosConf, callsConf, streamingClientConf), tezosConf.network, batchingConf)
+  val tezosNodeOperator = new TezosNodeOperator(tezosConf.network, batchingConf)
 
   /** close resources for application stop */
   private[this] def shutdown(): Unit = {
     logger.info("Doing clean-up")
     db.close()
     val nodeShutdown =
-      tezosNodeOperator.node
-        .shutdown()
+      nodeContext.shutdown()
         .flatMap((_: ShutdownComplete) => system.terminate())
 
     Await.result(nodeShutdown, shutdownWait)
--- a/src/main/scala/tech/cryptonomic/conseil/generic/chain/RemoteRpc.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/generic/chain/RemoteRpc.scala
@@ -1,0 +1,32 @@
+package tech.cryptonomic.conseil.generic.chain
+
+trait RemoteRpc[Eff[_], Req[_], Res[_]] {
+
+  type CallConfig
+  type PostPayload
+
+  def runGetCall[CallId](callConfig: CallConfig, request: Req[CallId], commandMap: CallId => String): Eff[Res[CallId]]
+
+  def runPostCall[CallId](callConfig: CallConfig, request: Req[CallId], commandMap: CallId => String, payload: Option[PostPayload]): Eff[Res[CallId]]
+
+}
+
+object RemoteRpc {
+
+  type Aux[Eff[_], Req[_], Res[_], Conf, Payload] = RemoteRpc[Eff, Req, Res] { type CallConfig = Conf; type PostPayload = Payload }
+
+  def apply[Eff[_], Req[_], Res[_], CC, PL](implicit remote: Aux[Eff, Req, Res, CC, PL]): Aux[Eff, Req, Res, CC, PL] = remote
+
+  def runGet[Eff[_], Req[_], Res[_], CallConfig, CallId](
+    callConfig: CallConfig, request: Req[CallId], commandMap: CallId => String
+  )(
+    implicit remote: Aux[Eff, Req, Res, CallConfig, _]
+  ): Eff[Res[CallId]] = remote.runGetCall(callConfig, request, commandMap)
+
+  def runPost[Eff[_], Req[_], Res[_], CallConfig, CallId, Payload](
+    callConfig: CallConfig, request: Req[CallId], commandMap: CallId => String, payload: Option[Payload] = None
+  )(
+    implicit remote: Aux[Eff, Req, Res, CallConfig, Payload]
+  ): Eff[Res[CallId]] = remote.runPostCall(callConfig, request, commandMap, payload)
+
+}
--- a/src/main/scala/tech/cryptonomic/conseil/generic/chain/RemoteRpc.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/generic/chain/RemoteRpc.scala
@@ -7,26 +7,72 @@ trait RemoteRpc[Eff[_], Req[_], Res[_]] {
 
   def runGetCall[CallId](callConfig: CallConfig, request: Req[CallId], commandMap: CallId => String): Eff[Res[CallId]]
 
-  def runPostCall[CallId](callConfig: CallConfig, request: Req[CallId], commandMap: CallId => String, payload: Option[PostPayload]): Eff[Res[CallId]]
+  def runPostCall[CallId](callConfig: CallConfig, request: Req[CallId], commandMap: CallId => String, payload: Option[PostPayload] = None): Eff[Res[CallId]]
 
 }
 
 object RemoteRpc {
 
+  import cats.{Id, Functor}
+  import cats.data.Const
+  import cats.syntax.functor._
+
   type Aux[Eff[_], Req[_], Res[_], Conf, Payload] = RemoteRpc[Eff, Req, Res] { type CallConfig = Conf; type PostPayload = Payload }
 
-  def apply[Eff[_], Req[_], Res[_], CC, PL](implicit remote: Aux[Eff, Req, Res, CC, PL]): Aux[Eff, Req, Res, CC, PL] = remote
-
-  def runGet[Eff[_], Req[_], Res[_], CallConfig, CallId](
-    callConfig: CallConfig, request: Req[CallId], commandMap: CallId => String
-  )(
-    implicit remote: Aux[Eff, Req, Res, CallConfig, _]
-  ): Eff[Res[CallId]] = remote.runGetCall(callConfig, request, commandMap)
-
-  def runPost[Eff[_], Req[_], Res[_], CallConfig, CallId, Payload](
-    callConfig: CallConfig, request: Req[CallId], commandMap: CallId => String, payload: Option[Payload] = None
-  )(
+  def apply[Eff[_], Req[_], Res[_], CallConfig, Payload](
     implicit remote: Aux[Eff, Req, Res, CallConfig, Payload]
-  ): Eff[Res[CallId]] = remote.runPostCall(callConfig, request, commandMap, payload)
+  ): Aux[Eff, Req, Res, CallConfig, Payload] = remote
+
+  def runGet[Eff[_], CallId, Req[_], Res[_], CallConfig](
+    callConfig: CallConfig,
+    request: Req[CallId],
+    commandMap: CallId => String
+  )(implicit remote: Aux[Eff, Req, Res, CallConfig, _]): Eff[Res[CallId]] =
+    remote.runGetCall(callConfig, request, commandMap)
+
+  def runGet[Eff[_], CallId, Req[_], Res[_]](
+    request: Req[CallId],
+    commandMap: CallId => String
+  )(implicit remote: Aux[Eff, Req, Res, Any, _]): Eff[Res[CallId]] =
+    remote.runGetCall((), request, commandMap)
+
+  def runGet[Eff[_]: Functor, CallId, Res](
+    request: CallId,
+    commandMap: CallId => String
+  )(implicit remote: Aux[Eff, Id, Const[Res, ?], Any, _]): Eff[Res] =
+    remote.runGetCall((), request, commandMap).map(_.getConst)
+
+  def runGet[Eff[_]: Functor, Res](
+    command: String,
+  )(implicit remote: Aux[Eff, Id, Const[Res, ?], Any, _]): Eff[Res] =
+    remote.runGetCall((), (), (_: Any) => command).map(_.getConst)
+
+  def runPost[Eff[_], CallId, Req[_], Res[_], CallConfig, Payload](
+    callConfig: CallConfig,
+    request: Req[CallId],
+    commandMap: CallId => String,
+    payload: Option[Payload] = None
+  )(implicit remote: Aux[Eff, Req, Res, CallConfig, Payload]): Eff[Res[CallId]] =
+    remote.runPostCall(callConfig, request, commandMap, payload)
+
+  def runPost[Eff[_], CallId, Req[_], Res[_], Payload](
+    request: Req[CallId],
+    commandMap: CallId => String,
+    payload: Option[Payload]
+  )(implicit remote: Aux[Eff, Req, Res, Any, Payload]): Eff[Res[CallId]] =
+    remote.runPostCall((), request, commandMap, payload)
+
+  def runPost[Eff[_]: Functor, CallId, Res, Payload](
+    request: CallId,
+    commandMap: CallId => String,
+    payload: Option[Payload]
+  )(implicit remote: Aux[Eff, Id, Const[Res, ?], Any, Payload]): Eff[Res] =
+    remote.runPostCall((), request, commandMap, payload).map(_.getConst)
+
+  def runPost[Eff[_]: Functor, Res, Payload](
+    command: String,
+    payload: Option[Payload]
+  )(implicit remote: Aux[Eff, Id, Const[Res, ?], Any, Payload]): Eff[Res] =
+    remote.runPostCall((), (), (_: Any) => command, payload).map(_.getConst)
 
 }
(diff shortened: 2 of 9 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/TezosRemotes.scala
```diff
@@ -0,0 +1,202 @@
+package tech.cryptonomic.conseil.tezos
+
+import tech.cryptonomic.conseil.config.{HttpStreamingConfiguration, NetworkCallsConfiguration}
+import tech.cryptonomic.conseil.config.Platforms.TezosConfiguration
+import tech.cryptonomic.conseil.generic.chain.RemoteRpc
+import tech.cryptonomic.conseil.util.JsonUtil.JsonString
+
+
+object TezosRemoteInstances {
```
ba205: The object name might be a good file name too.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosRemotes.scala
```diff
@@ -0,0 +1,202 @@
...
+    import com.typesafe.scalalogging.Logger
+    import cats.Monoid
+    import scala.concurrent.Future
+    import akka.actor.ActorSystem
+    import akka.stream.ActorMaterializer
+    import akka.stream.scaladsl.Source
+    import akka.http.scaladsl.Http
+    import akka.http.scaladsl.model._
+
+    case class RemoteContext(
```
ba205: Maybe the entire file/object should be TezosConfiguration/TezosContext? I think a rename is in order.

src/main/scala/tech/cryptonomic/conseil/generic/chain/RemoteRpc.scala
```diff
@@ -0,0 +1,78 @@
+package tech.cryptonomic.conseil.generic.chain
+
+trait RemoteRpc[Eff[_], Req[_], Res[_]] {
```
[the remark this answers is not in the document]
ivanopagano: The naming is totally free to change. The trait corresponds and replaces the existing RPCInterface.
About Req, Res, they are there to accomodate for multi-data requests, like
- sending a batch of inputs (e.g. a List[Offset])
- returning a result value (a representation of the REST content, e.g. json) wrapped in some extra type, to allow a result that contains the request argument. An example is the type `(Offset, Json)`, so that `Res[_] = (Offset, _)`, which has a single type whole to contain the actual result type.

Instead the `Future/IO` effect is represented by the `Eff[_]` type

src/main/scala/tech/cryptonomic/conseil/generic/chain/RemoteRpc.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  type CallConfig
+  type PostPayload
+
+  def runGetCall[CallId](callConfig: CallConfig, request: Req[CallId], commandMap: CallId => String): Eff[Res[CallId]]
```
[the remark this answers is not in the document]
ivanopagano: Sure, no problem, I simply found weird to say that you're "querying" an endpoint with a POST call... which is essentially an update.

src/main/scala/tech/cryptonomic/conseil/generic/chain/RemoteRpc.scala
```diff
@@ -0,0 +1,78 @@
+package tech.cryptonomic.conseil.generic.chain
+
+trait RemoteRpc[Eff[_], Req[_], Res[_]] {
+
+  type CallConfig
```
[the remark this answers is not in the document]
ivanopagano: This type is a wild-card to contain any extra information needed by the specific implementation of the RemoteRpc/RPCHandler to be passed in when making the remote query.
It's not necessarily credentials, it's... CallMetadata?

E.g. for multi-data requests, we need to pass the akka-http-client concurrency-level (an Int), but we might be needing more than one extra arg, so this type changes dramatically based on the implementation.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
```diff
@@ -60,19 +63,20 @@ trait BlocksDataFetchers {
 
     def makeUrl = (offset: Offset) => s"blocks/${hashRef.value}~${String.valueOf(offset)}"
```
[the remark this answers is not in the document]
ivanopagano: Not so easy, I guess, but I will try. I think I didn't gave this too much thought, in fact

src/main/scala/tech/cryptonomic/conseil/tezos/TezosRemotes.scala
ivanopagano: Good one

src/main/scala/tech/cryptonomic/conseil/tezos/TezosRemotes.scala
ivanopagano: 🤔 ... ok, what about TezosRemoteContext, or TezosRpcContext ? It has to do with how to connect to the node to make a REST call.
Still we need to think of something that will make sense in perspective, also for Eth and other platforms

src/main/scala/tech/cryptonomic/conseil/generic/chain/RemoteRpc.scala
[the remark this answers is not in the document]
ba205: I think it may be useful for our codebase, whenever we use an abstract effect type, to list 2-3 concrete examples in the documentation, as it may be non obvious to newcomers to the codebase what we're generalizing over.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -35,6 +36,7 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
   //the dispatcher is visible for all async operations in the following code
   implicit val system: ActorSystem = ActorSystem("lorre-system")
   implicit val dispatcher = system.dispatcher
+  implicit val nodeContext = RemoteContext(tezosConf, callsConf, streamingClientConf)
```
ba205: RemoteContext is an object under TezosRemote Instances, not a priority right now, but maybe in the future we want this to be a level higher? I do like this call though, because it's clear that tezosConf determines what remote context were accessing.

src/main/scala/tech/cryptonomic/conseil/generic/chain/RemoteRpc.scala
```diff
@@ -0,0 +1,78 @@
...
+  )(implicit remote: Aux[Eff, Id, Const[Res, ?], Any, Payload]): Eff[Res] =
+    remote.runPostCall((), request, commandMap, payload).map(_.getConst)
+
+  def runPost[Eff[_]: Functor, Res, Payload](
+    command: String,
+    payload: Option[Payload]
+  )(implicit remote: Aux[Eff, Id, Const[Res, ?], Any, Payload]): Eff[Res] =
+    remote.runPostCall((), (), (_: Any) => command, payload).map(_.getConst)
+
+}
```
ba205: In general, I like this a lot, especially seeing TezosNodeOperator now. Giving concrete examples and descriptive documentation is all that's necessary.

ivanopagano:
@ba205 I'm taking all your comments into account, yet I'm doing this parallel work on spare time, extra work... so progress will not be plainly obvious.
I eventually rewrote most traits/typeclasses to simplify everything, removing complicated generalizations with many type parameters that existed only to keep old and new designs both under the same abstraction.
I have a working solution that streams all data using fs2 streams and cats IO.

I need more time yet to finish any missing pieces, rewrite the tests that need updates, tune it to make it fast as the current solution.
I also would like to have an integration testing suite like Mike mentioned to verify that the refactor didn't lose any information.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ba205: Minor nitpick, maybe callsConf should be renamed networkTimeoutConf, and the class associated with it should be renamed as such? It seems the config is only used to update that aspect of network calls.

ivanopagano: I'll look into that, the final version of this whole re-work is quite different, might be that I've no more need of those.

No worry though, I'll incorporate your feedbacks in the latest proposal for this work, on which I hope to spend some time soon, it's almost ready.

src/main/scala/tech/cryptonomic/conseil/generic/chain/RemoteRpc.scala
```diff
@@ -0,0 +1,78 @@
...
+    request: Req[CallId],
+    commandMap: CallId => String
+  )(implicit remote: Aux[Eff, Req, Res, Any, _]): Eff[Res[CallId]] =
+    remote.runGetCall((), request, commandMap)
+
+  def runGet[Eff[_]: Functor, CallId, Res](
+    request: CallId,
+    commandMap: CallId => String
+  )(implicit remote: Aux[Eff, Id, Const[Res, ?], Any, _]): Eff[Res] =
+    remote.runGetCall((), request, commandMap).map(_.getConst)
```
vishakh: For my education, what does `getConst()` do?

ivanopagano: `Const[A, B]` is the type analogous to the function 
```scala
//a function taking two params that always returns the first
// it's some sort of hi-jacked function, you give it the first value and you get back a 
// another function that returns such value, independently of the passed argument 
def const[A, B](a: A)(b: B) = a
```
Now to the Const type, and its `getConst` attribute:
I had need to define the Rpc call to be able to return a value that might be tied to the input arguments (e.g. a tuple of the input/output), hence the parametric `Res[_]` return type, that can be used to return a `Res[CallId]` when multiple (lists) inputs (of type CallId) are passed as a batch.
In the case where we want the output to actually be independent of the `CallId` type, we can use the type
`Const[A, B]` which ignores the `B` type and is actually an alias for `A`
For the `runGet` that wants to return a simple result for a single request, the type `Const[Res, ?]` is a type constructor that takes one type, as expected by the generic form of RemoteRpc.Aux, but ignores it.

tl;dr; It all boils down to having a valid way to return a `Eff[Res]` that is independent of the single request type input `CallId`.
Even more, the evolution of this code is a huge simplification of all this, I found this encoding too general and ambitious, so I made it much simpler, at the cost of not being able to accomodate multiple designs in a single "form".
I'll try to push a PR out for you all to evaluate as soon as I can, based on my spare time

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: Ok, I've verified and what you suggested makes perfect sense, I renamed the config in the end result, I'll publish in a separate PR

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
```diff
@@ -60,19 +63,20 @@ trait BlocksDataFetchers {
...
 
+    // lazy val failureHandler = (offset: Offset, error: Throwable) =>
```
vishakh: How come these failure handlers are commented out? How are HTTP errors being handled now?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeInterface.scala
```diff
@@ -23,21 +23,28 @@ object ShutdownComplete extends ShutdownComplete
   * Interface into the Tezos blockchain.
   */
 trait TezosRPCInterface {
```
vishakh: Is this used anymore apart from tests? Has `RemoteRpc` replaced this functionality?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosRemotes.scala
```diff
@@ -0,0 +1,202 @@
...
+
+import tech.cryptonomic.conseil.config.{HttpStreamingConfiguration, NetworkCallsConfiguration}
+import tech.cryptonomic.conseil.config.Platforms.TezosConfiguration
+import tech.cryptonomic.conseil.generic.chain.RemoteRpc
+import tech.cryptonomic.conseil.util.JsonUtil.JsonString
+
+
+object TezosRemoteInstances {
+
+  object Akka {
```
vishakh: Could we give this a more descriptive name, e.g. `AkkaRemoteInstance`?

src/main/scala/tech/cryptonomic/conseil/generic/chain/RemoteRpc.scala
vishakh: How are HTTP errors handled now?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosRemotes.scala
```diff
@@ -0,0 +1,202 @@
...
+
+          implicit val system = context.system
+          implicit val dispatcher = system.dispatcher
+          implicit val materializer = ActorMaterializer()
+
+          type CallConfig = Any
+          type PostPayload = JsonString
+
+          def runGetCall[CallId](
+            callConfig: Any = (),
```
vishakh: Why not just a `Unit`?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosRemotes.scala
```diff
@@ -0,0 +1,202 @@
...
+            } yield {
+              val responseBody = strict.data.utf8String
+              logger.debug("Query results: {}", responseBody)
+              Const(JsonString sanitize responseBody)
+            }
+          }
+        }
+    }
+
+    object Streams {
```
vishakh: This does not look like a typical companion object. Does it make more sense to give it a slightly different name?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosRemotes.scala
```diff
@@ -0,0 +1,202 @@
...
+        import tezosConfig.nodeConfig._
+        s"$protocol://$hostname:$port/${pathPrefix}chains/main/$command"
+      }
+
+      def shutdown(): Future[ShutdownComplete] = {
+        rejectingCalls.compareAndSet(false, true)
+        Http(system).shutdownAllConnectionPools().map(_ => this.ShutdownComplete)(system.dispatcher)
+      }
+
+      def withRejectionControl[T, Container[_]](call: => Container[T])(implicit mono: Monoid[Container[T]]): Container[T] =
```
vishakh: Could you explain what this does and how it works? Why does it need a monoid to the extent that we are creating 'unnatural' monoids on futures?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
ivanopagano: In this version they're temporarily set off, yet this branch was never meant to get merged and deployed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeInterface.scala
ivanopagano: I eventually removed it in favour of RemoteRpc (which I renamed `RpcHandler` after @ba205 's suggestion)

src/main/scala/tech/cryptonomic/conseil/tezos/TezosRemotes.scala
ivanopagano: I'm not sure it would help, as this "module objects" are named to make instances visible through a local import that would look like
`import TezosRemotesInstances.Akka.Futures._` or
`import TezosRemotesInstances.Akka.Streams._` or 
`import TezosRemotesInstances.Cats.IO._`

I'm not sure if a more descriptive name will help here
`import TezosRemotesInstances.AkkaRemoteInstance.FutureInstances._` or something alike...
what do you think?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosRemotes.scala
ivanopagano: Sure, the weirdness can be explained like this:
- the original code provided a default failed reply for rejection on Future, namely Future.failed(...)
- to define a Rejection result for more generic "effects", we need a failed case default instance for `Container[T]`
- to get a "default" as a type class, Cats offers a specific type to provide such "empty" instance only in a secondary project called alleycats, as in the core project, empty is only available for a Monoid (as the neutral element for the monoid operation "append"). I didn't want to add an additional dependency only for this feature
- after thinking about it, this is not even the correct solution, because when rejecting the call it would be better to mark the response as a "failed" one, thus an `ApplicativeError` or `MonadError` is a better matched choice for this case. I'm in favour of changing the requirement, so that a rejection calls "raiseError"

src/main/scala/tech/cryptonomic/conseil/tezos/TezosRemotes.scala
ivanopagano: because here we're actually implementing the `trait`'s interface.

It should more aptly be defined as `override def runGetCall ...`

The method type signature is already set in stone. Since this parameter is actually ignored by this instance, the choice is to use a completely general input type (i.e. Any), so that you can pass whatever you want to the method call, as it will be ignored anyway.
As a default argument value, we could've used literally anything, so I picked `()` as the more memory efficient object to pass in and ignore.

To sum it up, if I changed `callConfig` to expect `Unit` the signature would not have matched and the `futureInstance` class would've failed to implement `RemoteRpc`

src/main/scala/tech/cryptonomic/conseil/tezos/TezosRemotes.scala
ivanopagano: Maybe we want to define some rule here about naming.

I am accustomed to use `object` in scala as a sort of locally-defined piece of modularity, that is, a _module_.

In this case, the name choice is driven by its expected usage in locally-scoped `import` directives.

It's much like cats defines (package) objects like `cats.instances.yourclassname._`, `cats.syntax.typeclassname._` and so on.

These objects are not meant to be called for the behaviour provided, but to be imported for the instances they bring in scope, they're more akin to namespaces than singleton static instances.

This was a design goal of the language, along with mixin-composition, to provide modular extensible behaviour, as outlined in [this post](http://lambda-the-ultimate.org/node/3735), whose last reference should point to [this paper by Odersky & Zenger](https://lampwww.epfl.ch/~odersky/papers/ScalableComponent.pdf)

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
```diff
@@ -274,17 +290,21 @@ trait BlocksDataFetchers {
...
-        node.runBatchedGetQuery(network, blocks, makeUrl, fetchConcurrency)
-          .onError { case err =>
-            logger.error("I encountered problems while fetching ballot votes from {}, for blocks {}. The error says {}",
-              network,
-              blocks.map(_.data.hash.value).mkString(", "),
-              err.getMessage
-            ).pure[Future]
-          }
-    )
+    // lazy val failureHandler = (block: Block, error: Throwable) =>
```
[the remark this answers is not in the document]
ivanopagano: Consider that all you see here is totally out-of-date 😛

src/main/scala/tech/cryptonomic/conseil/tezos/TezosRemotes.scala
```diff
@@ -0,0 +1,202 @@
...
+    import akka.http.scaladsl.model._
+
+    case class RemoteContext(
+      tezosConfig: TezosConfiguration,
+      requestConfig: NetworkCallsConfiguration,
+      streamingConfig: HttpStreamingConfiguration
+    )(implicit val system: ActorSystem) {
+
+      object ShutdownComplete extends ShutdownComplete
+      private val rejectingCalls = new java.util.concurrent.atomic.AtomicBoolean(false)
```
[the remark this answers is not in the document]
ivanopagano: Ok, good question, let me know if you have ideas because the concept lived on...

When the system starts shutting down, it might happen that Lorre is still running its batch of requests (tens or hundreds of those).
What happened was that those requests were run pointlessly and returned an akka http stream response that would generate lots of noisy errors about the body not being consumed within the given timeout (If I remember correctly).

So I introduced (time ago) this sort of short-circuiting state that would immediately reject any node request from lorre from when the system initiate its shutdown process (even if fired by an OS signal, like Ctrl-C)

ivanopagano:
overridden by #375

PR #338, opened 2019-04-16
Conseil 322-2: tagless final node operator

ivanopagano:
### It is not meant to be merged as-is but to provide material for progressive review of the whole refactor

The "pluggable-backend" approach will make it easier to try streaming designs, based on different implementations than akka-stream and plain scala futures
The long-term plan is to open a discussion to evaluate the switch to cats-effects/monix/fs2

---
The current step defines an alternative interface for the `NodeOperator` that is parametric on effects.
The parametric methods use implicit constraints (type classes) to decide what kind of actual effect can be used concretely, .e.g Monad, Functor, Appicative, ...
In addition the NodeOperator calls now will require available instances of data-fetchers and remote-rpc callers tailored to the specific effect.

We also rewrite Lorre to use the new NodeOperator, and provide the implicit instances before each call.
The resulting code replaces the exisiting future-based one, but doesn't currently implement the "batching logic", thus being unfitted yet for production deployment.
The plan for next step is to re-introduce the batching logic, in a design that decouples it from the NodeOperator itself.
We also introduce here a finer control on error logging for each requested rpc call.

Finally, the node test is no more based on an async model. The required fetchers and remote-rpc are provided as dummy/stubs, and only the actual node logic should be tested, arguably for best and more fine-grained checks.

The current version doesn't provide yet a substitute implementation for the "write-side" of the node, which was actually never used yet (but will be provided in the following PRs)

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -2,11 +2,20 @@ package tech.cryptonomic.conseil
 
 import akka.actor.ActorSystem
 import akka.Done
-import mouse.any._
 import com.typesafe.scalalogging.LazyLogging
-import tech.cryptonomic.conseil.tezos.{TezosTypes, FeeOperations, ShutdownComplete, TezosErrors, TezosNodeOperator, TezosDatabaseOperations => TezosDb}
+import tech.cryptonomic.conseil.tezos.{
+  AccountsDataFetchers,
+  ApiOperations,
+  BlocksDataFetchers,
+  FeeOperations,
+  NodeOperator,
+  ShutdownComplete,
+  TezosDatabaseOperations => TezosDb,
+  TezosErrors,
+  TezosRemoteInstances
+}
 import tech.cryptonomic.conseil.tezos.TezosRemoteInstances.Akka.RemoteContext
-import tech.cryptonomic.conseil.tezos.TezosTypes.BlockAccounts
+import tech.cryptonomic.conseil.tezos.TezosTypes.Block
 import tech.cryptonomic.conseil.io.MainOutputs.LorreOutput
 import tech.cryptonomic.conseil.util.DatabaseUtil
 import tech.cryptonomic.conseil.config.{Custom, Everything, LorreAppConfig, Newest}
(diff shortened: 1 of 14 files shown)


ivanopagano:
For additional info on the Final Tagless design, based on type classes:
https://scalac.io/tagless-final-pattern-for-scala-code/
https://scalac.io/typelevel-ecosystem-overview/

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -45,7 +57,26 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
-  val tezosNodeOperator = new TezosNodeOperator(tezosConf.network, batchingConf)
+
+  /* this is needed later to connect api logic to the fetcher,
+   * being dependent on the specific effect, i.e. Future, we can't
+   * devise a more generic solution, yet...
+   */
+  private[this] val fetchMaxLevel = ApiOperations.fetchMaxLevel _
+
+  //get all DataFetcheres and RemoteRpc instances needed in scope, for later use by the node methods
+  import TezosRemoteInstances.Akka.Futures._
```
ba205: IntelliJ lists this as unused, how does Node find this instance in scope?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -113,102 +145,35 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+      import cats.syntax.traverse._
+      import cats.syntax.foldable._
+      import cats.syntax.semigroup._
+      import cats.instances.list._
+      import cats.instances.option._
+      import cats.instances.int._
+      import slickeffect.implicits._
+
+      node.getVotingDetails[Future](blocks).flatMap {
+        case (proposals, bakersBlocks, ballotsBlocks) =>
```
ba205: writeVotingProposals and writeVotingBakers currently have errors saying "PostgresProfile.api.DBIO[Option[Int]] doesn't conform to the expected type G_[B_]"

ivanopagano: weird, I'll possibly disregard these errors as I will publish the final version, so this actual PR will never be actually merged...

Right now I don't even know how hard it is to apply current master changes to the WIP branch 😛

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: I guess IDEA is wrong, you can try commenting out the line, and the ensuing error should point to the point of usage 😉

ivanopagano:
overridden by #375

PR #339, opened 2019-04-16
Conseil 323 - default values for the predicates

piotrkosecki:
Because of lack of handling of default parameters in `endpoints` library I had to write a bit more boilerplate code to make default parameters work as expected. If anyone has any suggestions how to make it less complicated I'm open to any suggestions.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
@@ -206,7 +206,7 @@ object DataTypes {
     def validate(entity: String, tezosPlatformDiscovery: TezosPlatformDiscoveryOperations)(implicit ec: ExecutionContext):
     Future[Either[List[QueryValidationError], Query]] = {
 
-      val patchedPredicates = this.predicates.toList.flatten.map(_.toPredicate)
+      val patchedPredicates = predicates.getOrElse(List.empty).map(_.toPredicate)
       val query = this.into[Query]
         .withFieldConst(_.fields, fields.getOrElse(List.empty))
         .withFieldConst(_.predicates, patchedPredicates)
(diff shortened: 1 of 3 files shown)


src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -161,17 +196,25 @@ object DataTypes {
...
+    aggregation: Option[ApiAggregation] = None
   ) {
     /** Method which validates query fields */
     def validate(entity: String, tezosPlatformDiscovery: TezosPlatformDiscoveryOperations)(implicit ec: ExecutionContext):
     Future[Either[List[QueryValidationError], Query]] = {
 
-      val query = Query().patchWith(this)
+      val patchedPredicates = this.predicates.toList.flatten.map(_.toPredicate)
+      val query = this.into[Query]
+        .withFieldConst(_.fields, fields.getOrElse(List.empty))
```
ivanopagano: This is some chimney api, right?

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -161,17 +196,25 @@ object DataTypes {
...
     output: Option[OutputType],
-    aggregation: Option[Aggregation] = None
+    aggregation: Option[ApiAggregation] = None
   ) {
     /** Method which validates query fields */
     def validate(entity: String, tezosPlatformDiscovery: TezosPlatformDiscoveryOperations)(implicit ec: ExecutionContext):
     Future[Either[List[QueryValidationError], Query]] = {
 
-      val query = Query().patchWith(this)
+      val patchedPredicates = this.predicates.toList.flatten.map(_.toPredicate)
```
ivanopagano: I think we can do it easily with
```scala
val patchedPredicates = predicates.getOrElse(List.empty).map(_.toPredicate)
```

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: yes it is

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: yes, changed

PR #340, opened 2019-04-17
Conseil recursive sqlbuilder

ivanopagano:
This is a tiny patch to make concatenation of multiple sql actions using a recursive algorithm.
A test is added to verify the correctness of results

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
@@ -44,10 +44,12 @@ object DatabaseUtil {
         case Nil =>
           concatenateSqlActions(content, sql")")
         case head :: tail =>
-          val next = concatenateSqlActions(content, sql",'#$head'")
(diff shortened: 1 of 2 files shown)


src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
```diff
@@ -38,15 +38,18 @@ object DatabaseUtil {
...
       * @return SqlActionBuilder with values from parameter
       */
-    def insertValuesIntoSqlAction[T](xs: Seq[T]): SQLActionBuilder = {
-      var b = sql"("
-      var first = true
-      xs.foreach { x =>
-        if (first) first = false
-        else b = concatenateSqlActions(b, sql",")
-        b = concatenateSqlActions(b, sql"'#$x'")
+    def insertValuesIntoSqlAction[T](values: Seq[T]): SQLActionBuilder = {
```
piotrkosecki: please change the param name in the javadoc

ivanopagano: sure, done, nice catch

Issue #343, opened 2019-04-26
Supplementary metadata continued


anonymoussprocket:
Related to issue #316. Please add `placeholder` and `dataformat` fields to the metadata override file at attribute level. For example timestamp under operations might look like:

```
timestamp: {
  dataformat: 'yyyy MM dd HH:mm Z',
  visible: true
}
```

anonymoussprocket:
Supports https://github.com/Cryptonomic/Arronax/issues/166

PR #344, opened 2019-04-29
#343 - Supplementary metadata continued

dorians:
Added support for ``dataformat`` and ``placeholder `` on ``attributes`` level and for ``description`` on all levels.

Fixes: #343 and #316 (``description`` support)


src/main/scala/tech/cryptonomic/conseil/metadata/UnitTransformation.scala
```diff
@@ -7,30 +7,50 @@ import tech.cryptonomic.conseil.util.OptionUtil.when
 class UnitTransformation(overrides: MetadataOverridesConfiguration) {
 
   def overridePlatform(platform: Platform, path: PlatformPath): Option[Platform] = when(overrides.isVisible(path)) {
-    platform.copy(displayName = overrides
-      .platform(path)
-      .flatMap(_.displayName)
-      .getOrElse(platform.displayName))
+    val overridePlatform = overrides.platform(path)
```
piotrkosecki: I'm wondering if chimney would help here(it is already in the project)?: https://scalalandio.github.io/chimney/

src/main/scala/tech/cryptonomic/conseil/config/MetadataOverridesConfiguration.scala
```diff
@@ -29,10 +29,10 @@ case class MetadataOverridesConfiguration(metadataOverrides: Map[PlatformName, P
   def attribute(path: AttributePath): Option[AttributeConfiguration] = entity(path.up).flatMap(_.attributes.get(path.attribute))
 }
 
-case class PlatformConfiguration(displayName: Option[String], visible: Option[Boolean], networks: Map[NetworkName, NetworkConfiguration] = Map.empty)
```
ivanopagano: Is it me, or this looks like a good place to define a parametric case class?

```scala
case class OverridesConfiguration[N <: Name, C <: Configuration](displayName: ..., visible: ..., description: ..., overrides: Map[N, C])
```

src/main/scala/tech/cryptonomic/conseil/metadata/UnitTransformation.scala
ivanopagano: yes, it should be in the dependencies, so it might be worthwhile to try it out and see if it reads better

src/test/scala/tech/cryptonomic/conseil/metadata/MetadataServiceTest.scala
```diff
@@ -62,9 +71,22 @@ class MetadataServiceTest extends WordSpec with Matchers with ScalatestRouteTest
...
+      val result = sut(overriddenConfiguration).getNetworks(PlatformPath("tezos"))
+
+      // then
+      result shouldBe Some(List(Network("mainnet", "overwritten name", "tezos", "mainnet", Some("description"))))
+    }
+
     "filter out a hidden network" in {
       // given
-      val overriddenConfiguration = Map("tezos" -> PlatformConfiguration(None, Some(true), Map("mainnet" -> NetworkConfiguration(None, Some(false)))))
+      val overriddenConfiguration = Map("tezos" -> PlatformConfiguration(None, Some(true), None, Map("mainnet" -> NetworkConfiguration(None, Some(false)))))
```
ivanopagano: would it be better to use named parameters with defaults to avoid writing all those `None`, `Some` which doesn't look so obvious to follow?

src/main/scala/tech/cryptonomic/conseil/generic/chain/PlatformDiscoveryTypes.scala
```diff
@@ -24,7 +24,10 @@ object PlatformDiscoveryTypes {
     dataType: DataType,
     cardinality: Option[Int],
     keyType: KeyType,
-    entity: String
+    entity: String,
+    description: Option[String] = None,
+    placeholder: Option[String] = None,
+    dataformat: Option[String] = None
```
anonymoussprocket: `displayName` is camel-case, `dataformat` isn't. Pick one please.

src/main/scala/tech/cryptonomic/conseil/metadata/UnitTransformation.scala
dorians: It is not possible right now without complicating the code. 
I proposed a ticket to Chimney: https://github.com/scalalandio/chimney/issues/106

PR #348, opened 2019-05-03
Pretty docs

anonymoussprocket:
This will absorb and replace /doc

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -368,6 +368,23 @@ Send this query to `/v2/data/tezos/<network>/accounts`
 }
 ```
 
+#### Top 20 Bakers by roll count
+
+Send this query to `/v2/data/tezos/<network>/bakers`
+
+```json
+{
+    "fields": ["pkh", "rolls"],
+    "predicates": [],
+    "orderBy": [
+        { "field": "block_level", "direction": "desc" },
+        { "field": "rolls", "direction": "desc" }
+    ],
+    "limit": 20,
+    "output": "csv"
+}
+```
+
 #### All originated accounts which are smart contracts
 
 Send this query to `/v2/data/tezos/<network>/accounts`
--- a/docs/README.md
+++ b/docs/README.md
@@ -319,13 +319,13 @@ Here's a collection of interesting datasets. Run these examples as
 
 ```bash
 curl -H 'Content-Type: application/json' -H 'apiKey: <API key>' \
--X POST '<conseil-host>/v2/data/tezos/alphanet/blocks/' \
+-X POST '<conseil-host>/v2/data/<platform>/<network>/<entity>/' \
 -d '<example>'
 ```
 
 #### Bakers by block count in April 2019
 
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -0,0 +1,288 @@
...
+...
+```
+
+#### Extended metadata
+
+As mentioned in the [Datasource configuration](#datasource-configuration) section, it's possible to provide additional details for the items Conseil manages. 
+
+### Tezos Chain Data Query
+
+Data requests in Conseil v2 API are sent via `POST` command and have a prefix of `/v2/data/<platform>/<network>/<entity>`. The items in curly braces match up to the `name` property of the related metadata results. There additional header to be set in the request: `Content-Type: application/json`.
```
ivanopagano: >The items in _curly braces_ match up...

actually _angle brackets_ ?

docs/README.md
```diff
@@ -0,0 +1,288 @@
...
+
+The query grammar allows for some fine-grained data extraction.
+
+#### `fields`
+
+It is possible to narrow down the fields returned in the result set by listing just the ones that are necessary. The `fields` property of the query is an array of `string`, these values are the `name`s from the `/v2/metadata/<platform>/<network>/<entity>/attributes/` metadata response.
+
+#### `predicates`
+
+`predicates` is an array containing objects. The inner object as several properties:
```
ivanopagano: >The inner object _has_ several properties:

typo

docs/README.md
```diff
@@ -0,0 +1,288 @@
...
+- `field` – attribute name from the metadata response.
+- `operation` – one of: in, between, like, lt, gt, eq, startsWith, endsWith, before, after.
+- `set` – an array of values to compare against. `in` requires two or more elements, `between` must have exactly two, the rest of the operations require a single element.
+- `inverse` – boolean, setting it to `true` applied a `NOT` to the operator
```
ivanopagano: I'm starting to wonder (late for doing so) if `negate` would better pass the meaning across?

On the other hand a boolean value named "negate" might become confusing

docs/README.md
```diff
@@ -0,0 +1,288 @@
...
+- `precision` – for numeric field matches this specifies the number of decimal places to round to.
+
+#### `limit`
+
+Specifies the maximum number of records to return.
+
+#### `orderBy`
+
+Sort condition, multiple may be supplied for a single query. Inner obejct(s) will contain `field` and `direction` properties. The former is `name` from the `/v2/metadata/<platform>/<network>/<entity>/attributes/` metadata response, the latter is one of 'asc', 'desc'.
```
ivanopagano: I see an small inconsistency in how keywords are emphasised like the predicate operation enumerations, which are plain text, the same used as an example for the predicates' set field, and here with the ordering values in single quotes.
Using a single consistent notation might help the reader

ivanopagano: this applies to the aggregation's functions

docs/README.md
vishakh: Interesting point. Perhaps `invert` might also work?

PR #360, opened 2019-05-09
351 - increasing stack size in readme

dorians:


ivanopagano:
There's a "run locally" section in the document that actually calls sbt to run conseil and lorre. There too we'd need to increase the stack size, because probably the cold run of sbt will force a compilation...

Issue #357, opened 2019-05-08
Create `delegates' and 'delegated_contracts' tables


vishakh:
Conseil currently does not store information about bakers other than their weight in the governance process (`bakers` table). We should therefore add this information.

Using the Tezos RPC call `/chains/main/blocks/{blockid}/context/delegates/{address}` we should populate a table called `delegates`. The `delegated_contracts` field in the returned data is an array so its contents should be stored in a `delegated_contracts` table which links to the parent table `delegates`.

ivanopagano:
I expect the delegated contracts to be the same referenced by blocks, therefore they should look something like

```json
{ "manager": "pkh",
    "balance": 1000,
    "spendable": true,
    "delegate": { "setable": true, "value": "pkh" },
    "script": { "code" : <michelin code>, "storage": <micheline code> },
    "counter": 401
}
```
?

ivanopagano:
and @ba205 @vishakh 
where is the connection between a block and the delegates we're looking for?
Should I collect pkh from the block's operations?

ivanopagano:
btw, I'll ask in Riot later, yet I can't seem to find if there's a json-schema description of the involved objects

BrianGuo:
I'm curious about the decision to store `delegated_contracts` as a separate table, when `delegated_contracts` is a subset of the `accounts` table and the relationship between the delegate and the delegator could potentially be codified as a property of some `accounts` table (I'm assuming accounts is a table).

vishakh:
@BrianGuo The `accounts` table indeed already contains delegation information. We are creating `delegated_contracts` as a redundancy.

ivanopagano:
@vishakh 
should we think of renaming the bakers' table to explicitly say it's about the governance process, and use `delegate_bakers` for the baking delegates?
Will this make things more clear?

ivanopagano:
Like, `bakers_rolls` and `baking_delegates`

PR #362, opened 2019-05-14
Conseil 357-1: delegates checkpoint table

ivanopagano:
Relates to #357 

In the same way that accounts for each stored block are first indexed by id in a "checkpoint" table and then fetched and stored, we add a like-wise durable checkpoint for delegates key hashes, filled when accounts are stored.
Such data will be used to download the delegates and their contracts in a separate step, guaranteeing that any failure occurring during this step will not lose information on the delegates for the blocks already stored.
In case of failure the delegates download will be re-attempted during the next Lorre cycle, eventually cleaning the checkpoints once the operation succeeds.

This PR only store the delegates to the checkpoint.
_The actual fetching of delegates and contracts and cleanup of the checkpoints will be part of the next PR._

A new table and related constraints are added

```slq
CREATE TABLE public.delegates_checkpoint (
    delegate_pkh character varying NOT NULL,
    block_id character varying NOT NULL REFERENCES public.blocks(hash),
    block_level integer DEFAULT '-1'::integer NOT NULL
);

CREATE INDEX ix_delegates_checkpoint_block_level ON public.delegates_checkpoint USING btree (block_level);

```

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -5,7 +5,7 @@ import akka.Done
 import mouse.any._
 import com.typesafe.scalalogging.LazyLogging
 import tech.cryptonomic.conseil.tezos.{TezosTypes, FeeOperations, ShutdownComplete, TezosErrors, TezosNodeInterface, TezosNodeOperator, TezosDatabaseOperations => TezosDb}
-import tech.cryptonomic.conseil.tezos.TezosTypes.BlockAccounts
+import tech.cryptonomic.conseil.tezos.TezosTypes.{BlockTagged, Account, AccountId}
 import tech.cryptonomic.conseil.io.MainOutputs.LorreOutput
 import tech.cryptonomic.conseil.util.DatabaseUtil
 import tech.cryptonomic.conseil.config.{Custom, Everything, LorreAppConfig, Newest}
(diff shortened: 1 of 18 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -351,10 +351,44 @@ object TezosTypes {
...
+    blockHash: BlockHash,
+    blockLevel: Int,
+    content: T
+  ) {
+    val asTuple = (blockHash, blockLevel, content)
+  }
+
+  final case class Delegate(
+    balance: PositiveBigNumber,
+    frozen_balance: PositiveBigNumber,
```
dorians: This file has a mixed style of writing keys (sometimes underscore, sometimes snakeCase)

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -260,6 +261,9 @@ class TezosNodeOperator(val node: TezosRPCInterface, val network: String, batchC
     (fetchProposals, fetchBakers, fetchBallots).tupled.run(blocks.filterNot(b => isGenesis(b.data)))
   }
 
+  //move it to the node operator
+  def getDelegatesForBlock(keys: List[PublicKeyHash], blockHash: BlockHash): Future[Map[Delegate, Contract]] = ???
```
dorians: is that needed?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -238,19 +238,33 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
     * we should evaluate how to handle failed processing
     */
   private[this] def processTezosAccounts(): Future[Done] = {
+    import cats.instances.list._
```
dorians: Why not put this above object definition?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
ivanopagano: I noticed too, it's a legacy leftover from the fact that the first types where roughly matching the json-schema from tezos, while later on I introduced an exact match, because I wanted to reduce any ambiguity on the mapping.

What would you suggest? We might open a new PR, but we need to decide if we prefer an exact attribute naming match or only a rough camelCase-converted match

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: This might change, but originally cats was used only within a limited scope of the code.

I tried to limit cats values to local scopes, as long as usage is not wide-spread... maybe it's the right time to, I didn't stop to reconsider if it makes sense

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ivanopagano: Must've been a leftover of ongoing work... I'll remove it and see if anything breaks, thanks

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
dorians: I prefer camelCase if that not makes anything significantly harder.

ivanopagano: @vishakh @ba205 @piotrkosecki can I have your vote in, too?

ivanopagano: I've opened issue #368 to make the eventual changes

Issue #351, opened 2019-05-07
Stackoverflow on compilation


dorians:
Running ``sbt compile`` causes ``[error] (Compile / compileIncremental) java.lang.StackOverflowError``. 

We can make a workaround increasing the memory for a stack: ``sbt compile -J-Xss32m``. 

We should investigate the problem and figure it out why we need so much memory.

dorians:
I spent some time trying to figure out why do we need so much memory for a stack. Apparently, we are using really heavy tools and that memory is just needed to use them. 

https://github.com/pureconfig/pureconfig/issues/390
https://circe.github.io/circe/codecs/known-issues.html
https://github.com/milessabin/shapeless/issues/852

ivanopagano:
so it was just a question of time before the issue would've exploded?

dorians:
@ivanopagano Exactly

ba205:
@anonymoussprocket sbt compile works locally for me.

PR #367, opened 2019-05-21
Add missing documentation reference to jvm stack-size to use for our custom sbt commands

ivanopagano:
relates to #360 and #351 

Add to the readme documentation additional jvm stack options to safely run our custom development commands: `runConseil` and `runLorre`

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -34,7 +34,7 @@ java -Xms512m -Xmx14g -Dconfig.file=conseil.conf -cp conseil.jar tech.cryptonomi
 
 In this command, the `-Xms512m` and `-Xmx14g` can be changed based on the amount of memory available on the system, the `-Dconfig.file=conseil.conf` can be changed to point to any user created config file that contains the credentials for the respective db and Tezos node. The file can also contain any overrides for settings provided by the `reference.conf` file. Lastly, `alphanet` is the network being run by the Tezos node, it can and should be changed if running `mainnet` or `zeronet`.
 
-Start `conseil` as: 
+Start `conseil` as:
 
 ```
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -47,13 +47,13 @@ For examples of how we run these services check out the [Nautilus](https://githu
 To run `lorre` and `conseil` with `sbt`, use the following:
 
 ```
-env SBT_OPTS="-Dconfig.file=conseil.conf" && sbt "runConseil"
-env SBT_OPTS="-Dconfig.file=conseil.conf" && sbt "runLorre alphanet"
+env SBT_OPTS="-Dconfig.file=conseil.conf" && sbt -J-Xss32m "runConseil"
```
dorians: Have you checked how does it really work? We need a bigger stack only for compilation. I have a suspicion that that change will affect runtime as well and might cause memory problems on production.

ivanopagano: using `sbt` to run any operation will usually require recompilation, especially if you're not in the sbt console already, I tried those commands and they break without the jvm options

ivanopagano: I'm not convinced on the other hand that it will produce problems during execution of a pre-packaged jar.
If the stack requirement are caused by shapeless compile-time, as it seems, I don't see any reason why they should affect the runtime

dorians: I've checked it with `VisualVM` and it looks that when you run ``sbt runConseil`` sbt actually uses the same instance of JVM to compile code and to run an app. However, I've realized that this section is about running an app locally and on production we recommend running it using `java` command. That's good.

ivanopagano: I lately mentioned to @piotrkosecki the fact that no fork is actually done by sbt in my local dev env, which is weird. I guess we got something amiss, but as you say, the good part is that we advice to run with `java`

PR #369, opened 2019-05-21
Improvement for the distinct count query

piotrkosecki:
Simple trick which speeds up counting distinct values.


src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -293,7 +293,7 @@ object TezosDatabaseOperations extends LazyLogging {
     * @return       amount of distinct elements in given column
     */
   def countDistinct(table: String, column: String)(implicit ec: ExecutionContext): DBIO[Int] =
-    sql"""SELECT COUNT(DISTINCT #$column) FROM #$table""".as[Int].map(_.head)
+    sql"""SELECT COUNT(*) FROM (SELECT DISTINCT #$column FROM #$table) AS temp""".as[Int].map(_.head)
```
dorians: do we need `` AS temp`` part?

vishakh: Yes, it's a peculiarity of Postgres.

PR #371, opened 2019-05-21
370 - git good practices

dorians:
After we agree on the final version we should consider creating a separate repo for it (or maybe use Github's wiki?) and link it from any README we have.

### DIFF
--- a/CONTRIBUTION.md
+++ b/CONTRIBUTION.md
@@ -11,6 +11,9 @@ Allowed prefixes:
 * bugfix
 * docs
 * test (when adding missing tests)
+* improvement
+* refactor
+* chores (for those tasks that needs be done for infrastructure compliance, some policy change)
 * ...
 
 Commits:
@@ -26,14 +29,14 @@ Pull requests:
 Each pull request must have a name in the following format where `000` represents a ticket number which this PR refers 
 to:
 ```
-000 - description of what was changed
+000(-1) - description of what was changed
 ```
 
-Each pull request must have a nice description of what was changed. It must have 
-information about all necessary actions need to be taken during deployment like database migrations or additional 
-configuration.
+When you'are creating more than one PR for a ticket you should use postfixes like `-1`, `-2` to define order of merging.
+
+Each pull request must have a nice description of what was changed. The description should start with a list of all 
+corresponding tickets it closes or refers to. Then it must have information about all necessary actions need to be taken 
+during deployment like database migrations or additional configuration.
 
 Pull requests should be as concise as possible to make them easier to review. If your PR gets bigger and verbose, 
 consider slice it into a few smaller but fully functional ones.
-
-The description should end with a list of all corresponding tickets it closes or refers to.
--- a/CONTRIBUTION.md
+++ b/CONTRIBUTION.md
@@ -1,0 +1,39 @@
+Branching names:
+---
+Each branch must have a name in the following format with a proper prefix and number of Github's ticket it refers to:
+
+```
+prefix/000-branch-name
+``` 
+
+Allowed prefixes:
+* feature
+* bugfix
+* docs
+* test (when adding missing tests)
+* ...
+
+Commits:
+---
+Each commit (except for merge commits) must have a name in the following format:
+
+```
+000 - description of what was changed
+```
+
+Pull requests:
+---
+Each pull request must have a name in the following format where `000` represents a ticket number which this PR refers 
+to:
+```
+000 - description of what was changed
+```
+
+Each pull request must have a nice description of what was changed. It must have 
+information about all necessary actions need to be taken during deployment like database migrations or additional 
+configuration.
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTION.md
```diff
@@ -0,0 +1,39 @@
...
+```
+prefix/000-branch-name
+``` 
+
+Allowed prefixes:
+* feature
+* bugfix
+* docs
+* test (when adding missing tests)
+* ...
```
piotrkosecki: +improvement?

CONTRIBUTION.md
```diff
@@ -0,0 +1,39 @@
+Branching names:
+---
+Each branch must have a name in the following format with a proper prefix and number of Github's ticket it refers to:
+
+```
+prefix/000-branch-name
+``` 
+
+Allowed prefixes:
+* feature
```
ivanopagano: I'd shorten 
feature -> feat
bugfix -> fix

if clear enough

CONTRIBUTION.md
ivanopagano: `refactor` sounds ok to me

also I usually find that `chores` can turn out useful, for those tasks that needs be done for infrastructure compliance, some policy change, ...

CONTRIBUTION.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+Each pull request must have a nice description of what was changed. It must have 
+information about all necessary actions need to be taken during deployment like database migrations or additional 
+configuration.
+
+Pull requests should be as concise as possible to make them easier to review. If your PR gets bigger and verbose, 
+consider slice it into a few smaller but fully functional ones.
+
+The description should end with a list of all corresponding tickets it closes or refers to.
```
ivanopagano: I don't know if it's a commonly shared standard, but I'd personally put those right in the beginning

CONTRIBUTION.md
```diff
@@ -0,0 +1,39 @@
...
+* docs
+* test (when adding missing tests)
+* ...
+
+Commits:
+---
+Each commit (except for merge commits) must have a name in the following format:
+
+```
+000 - description of what was changed
```
ivanopagano: Including the ticket name in the commit description is quite unusual I must say.
There is extensive material on how to write a commit message, I'll find some reference

ivanopagano: https://chris.beams.io/posts/git-commit/ is one that seems inline with common practice

vishakh: Yes, the ticket name in the commit might be too painful, especially if we generally keep a one ticket to one PR association.

CONTRIBUTION.md
vishakh: This is helpful as the original ticket gets an automatic link to the associated PRs.

ivanopagano: I'm only suggesting to put those links right at the _beginning_, instead of the _end_ of the description

CONTRIBUTION.md
dorians: I personally find it useful when I'm debugging code. It's easier to track a reason why that line was changed that way when you can link it to its task. It makes reading git history more comfortable as well. Otherways you need to find a task based on a merge commit name which sometimes (i.e. when you use feature branch) might be tricky to track.

CONTRIBUTION.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+Pull requests:
+---
+Each pull request must have a name in the following format where `000` represents a ticket number which this PR refers 
+to:
+```
+000 - description of what was changed
```
ivanopagano: I lately found myself in need to progressively lay out a task solution through multiple PRs, to avoid an overblown single PR.
What about using a common convention for those "progressive developments" like 
`000[-0] - brief and clear description of the actual change`

So that it would be easier to figure out the sequence of steps needed to merge the whole task solution?

CONTRIBUTION.md
vishakh: Is there a way to enforce this on GitHub using any sort of hooks?

CONTRIBUTION.md
vishakh: That would be great.

CONTRIBUTION.md
ba205: That would be fine, maybe start from 1 instead of 0 though.

ivanopagano: Actually that was a lapsus, I didn't meant to start from 0 (damn array programming 😁 !)

CONTRIBUTION.md
dorians: I would leave them not shorten to be consistent and so I do not have to wonder which version to choose (fix vs. bug)

CONTRIBUTION.md
dorians: I'm not sure. There is for GIT for sure but it requires to install it manually. Anyway, I'm not sure if adding an automatic check would be useful for us.

PR #373, opened 2019-05-22
357-2 - Store delegates and delegated contract information based on checkpoint

### LINKED ISSUE #357: Create `delegates' and 'delegated_contracts' tables

ivanopagano:
closes #357 
At each cycle the delegates pkh from a block operations' are saved in a checkpoint table, transactionally.
Right after that we now read the key hashes and fetch complete delegates data and store that on a table, with the addition of copying any account data for the delegated contracts to a separate table.

The delegated_contracts data is a duplicate of the more general accounts.

We need to setup 2 new tables with associated indexes and FKs

```sql
CREATE TABLE public.delegates (
    pkh character varying PRIMARY KEY,
    block_id character varying NOT NULL,
    balance numeric,
    frozen_balance numeric,
    staking_balance numeric,
    delegated_balance numeric,
    deactivated boolean NOT NULL,
    grace_period integer NOT NULL,
    block_level integer DEFAULT '-1'::integer NOT NULL
);

CREATE TABLE public.delegated_contracts (
    account_id character varying NOT NULL,
    delegate_value character varying
);

ALTER TABLE ONLY public.delegated_contracts
    ADD CONSTRAINT contracts_delegate_pkh_fkey FOREIGN KEY (delegate_value) REFERENCES public.delegates(pkh);

ALTER TABLE ONLY public.delegated_contracts
    ADD CONSTRAINT contracts_account_id_fkey FOREIGN KEY (account_id) REFERENCES public.accounts(account_id);

ALTER TABLE ONLY public.delegates
    ADD CONSTRAINT delegates_block_id_fkey FOREIGN KEY (block_id) REFERENCES public.blocks(hash);

ALTER TABLE IF EXISTS bakers
    RENAME TO rolls;
-- we won't rename the FK constraint
```

### DIFF
--- a/doc/conseil.sql
+++ b/doc/conseil.sql
@@ -189,7 +189,31 @@ CREATE TABLE public.balance_updates (
     category character varying
 );
 
+CREATE TABLE public.delegates (
+    pkh character varying PRIMARY KEY,
+    block_id character varying NOT NULL,
+    balance numeric,
+    frozen_balance numeric,
+    staking_balance numeric,
+    delegated_balance numeric,
+    deactivated boolean NOT NULL,
+    grace_period integer NOT NULL,
+    block_level numeric DEFAULT '-1'::integer NOT NULL
+);
 
+CREATE TABLE public.delegated_contracts (
+    account_id character varying NOT NULL,
+    block_id character varying NOT NULL,
+    manager character varying NOT NULL,
+    spendable boolean NOT NULL,
+    delegate_setable boolean NOT NULL,
+    delegate_value character varying,
+    counter integer NOT NULL,
+    script character varying,
+    storage character varying,
+    balance numeric NOT NULL,
+    block_level numeric DEFAULT '-1'::integer NOT NULL
+);
 --
 -- Name: accounts_checkpoint; Type: TABLE; Schema: public; Owner: -
 --
@@ -408,6 +432,15 @@ CREATE INDEX ix_proposals_protocol ON public.proposals USING btree (protocol_has
 
 ALTER TABLE ONLY public.accounts
     ADD CONSTRAINT accounts_block_id_fkey FOREIGN KEY (block_id) REFERENCES public.blocks(hash);
+
+ALTER TABLE ONLY public.delegated_contracts
+    ADD CONSTRAINT contracts_delegate_pkh_fkey FOREIGN KEY (delegate_value) REFERENCES public.delegates(pkh);
+
+ALTER TABLE ONLY public.delegated_contracts
+    ADD CONSTRAINT contracts_block_id_fkey FOREIGN KEY (block_id) REFERENCES public.blocks(hash);
+
+ALTER TABLE ONLY public.delegates
+    ADD CONSTRAINT delegates_block_id_fkey FOREIGN KEY (block_id) REFERENCES public.blocks(hash);
 
 --
 -- Name: accounts_checkpoint checkpoint_block_id_fkey; Type: FK CONSTRAINT; Schema: public; Owner: -
--- a/doc/conseil.sql
+++ b/doc/conseil.sql
@@ -203,16 +203,7 @@ CREATE TABLE public.delegates (
 
 CREATE TABLE public.delegated_contracts (
     account_id character varying NOT NULL,
-    block_id character varying NOT NULL,
-    manager character varying NOT NULL,
-    spendable boolean NOT NULL,
-    delegate_setable boolean NOT NULL,
-    delegate_value character varying,
-    counter integer NOT NULL,
-    script character varying,
-    storage character varying,
-    balance numeric NOT NULL,
-    block_level numeric DEFAULT '-1'::integer NOT NULL
+    delegate_value character varying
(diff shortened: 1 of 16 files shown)


ivanopagano:
Currently drafted because I want to rename the current `bakers` table to `rolls`, to reduce ambiguity.

Nonetheless the PR is ready for reviews.

ivanopagano:
> Currently drafted because I want to rename the current `bakers` table to `rolls`, to reduce ambiguity.
> 
> Nonetheless the PR is ready for reviews.
Done

doc/conseil.sql
```diff
@@ -189,7 +189,31 @@ CREATE TABLE public.balance_updates (
...
+    balance numeric,
+    frozen_balance numeric,
+    staking_balance numeric,
+    delegated_balance numeric,
+    deactivated boolean NOT NULL,
+    grace_period integer NOT NULL,
+    block_level integer DEFAULT '-1'::integer NOT NULL
+);
 
+CREATE TABLE public.delegated_contracts (
```
vishakh: This would duplicate a lot of data from the `accounts` table. We are only supposed to store the IDs here for redundancy purposes.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -22,12 +21,12 @@ object TezosTypes {
     private val script = GenLens[Origination](_.script)
     private val parameters = GenLens[Transaction](_.parameters)
 
-    private val storage = GenLens[Contracts](_.storage)
-    private val code = GenLens[Contracts](_.code)
+    private val storage = GenLens[Scripted.Contracts](_.storage)
```
vishakh: Could we use this occasion to leave a doc comment for `Scripted`?

doc/conseil.sql
ivanopagano: Ouch, fine, I would assume at least a reference to the delegate, too?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
ivanopagano: When I started doing 1-on-1 conversion of tezos schema definitions to Scala types, I sticked as much as possible to the original schema structure, so that it should be much easier to figure out how the two maps to each other.
In the operations' schema this sub-object was defined as Scripted.Contracts.

The upside is that we should be able to better understand what might be broken if a new protocol changes the schema and the json decoder stops parsing correctly some piece of data.

I don't know if this is what you meant or if you were talking about something else here

doc/conseil.sql
ivanopagano: @vishakh request for clarification:
which one is the relation field (FK) to join delegated contracts with their delegate?
I'm going for delegate.value

ivanopagano: Thanks to this change I found out another flaky test, duplicate from the one I fixed earlier! 🍪

src/test/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperationsTest.scala
```diff
@@ -47,6 +47,7 @@ class TezosPlatformDiscoveryOperationsTest
   override def beforeAll(): Unit = {
     super.beforeAll()
     sut.init()
+    ()
```
dorians: why is that for?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -308,6 +307,66 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
     }
   }
 
+  /** Fetches and stores all delegates from the latest blocks stored in the database. */
```
dorians: I'm not sure why it's here in that PR. This code is already in master.

ivanopagano: Where? I don't see it...
https://github.com/Cryptonomic/Conseil/blob/master/src/main/scala/tech/cryptonomic/conseil/Lorre.scala#L292

doc/conseil.sql
vishakh: > Ouch, fine, I would assume at least a reference to the delegate, too?

Indeed.

vishakh: > @vishakh request for clarification:
> which one is the relation field (FK) to join delegated contracts with their delegate?
> I'm going for delegate.value

You mean between the `delegated_contracts` and `delegates` tables? `delegated_contracts.account_id` should join to `delegates.pkh`.

PR #376, opened 2019-05-23
354 - additional supplementary metadata

### LINKED ISSUE #354: Supplementary metadata continued
Allow definition of arbitrary metadata keys, consider the following:

```
value-map: {0: "Yay", 1: "Nay", 2: "Pass"}
```

This could be used to describe the low-cardinality enum in ballots.ballot. UIs can be built to interpret this information the way that suites them. All Conseil would need to do is pass this data along.

Additionally, it should be possible to override any metadata key. For example, while some numbers in the schema may appear to be `int`, they are functionally `decimal` with some sort of scaling. Take as an example accounts.balance, while stored as in, it is actually µtz. This value would be more meaningfully displayed as n/1000000.
```
scale: 6
```
Would indicate to the UI how to display such a value.

Overriding datatype further would help identify special strings like addresses.
```
dataType: "address"
```

dorians:
closes #354, #379

Adds support for 4 additional fields for supplementary metadata:
* `value-map`: `Map[String, String]`, example: `{"0": "Yay", "1": "Nay", "2": "Pass"}`,
* `references`: `Map[String, String]`
* `data-type`: `String`, since this value is later mapped to enum, values are limited to those supported by method [mapType](https://github.com/Cryptonomic/Conseil/blob/feature/354-additional-supplementary-metadata/src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala#L31),
* `scale`: `Int`

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -176,10 +176,11 @@ metadata-overrides {
                 display-name: "Address"
                 placeholder: "tz1..."
                 visible: true
-                data-type: "hash"
+                data-type: "AccountAddress"
               }
               block_id {
                 display-name: "Block Hash"
+                data-type: "Hash"
                 visible: true
                 reference: {
                   entity: "blocks"
@@ -188,7 +189,7 @@ metadata-overrides {
               }
               manager {
                 visible: true
-                data-type: "hash"
+                data-type: "AccountAddress"
                 reference: {
                   entity: "accounts"
                   key: "account_id"
@@ -204,7 +205,7 @@ metadata-overrides {
               delegate_value {
                 display-name: "Delegate"
                 visible: true
-                data-type: "hash"
+                data-type: "AccountAddress"
                 reference: {
                   entity: "accounts"
                   key: "account_id"
@@ -239,7 +240,7 @@ metadata-overrides {
               pkh {
                 display-name: "Address"
                 visible: true
-                data-type: "hash"
+                data-type: "AccountAddress"
                 reference: {
                   entity: "accounts"
                   key: "account_id"
@@ -250,6 +251,7 @@ metadata-overrides {
               }
               block_id {
                 display-name: "Block Hash"
+                data-type: "Hash"
                 visible: true
                 reference: {
                   entity: "blocks"
@@ -308,7 +310,7 @@ metadata-overrides {
               pkh {
                 display-name: "Address"
                 visible: true
-                data-type: "hash"
+                data-type: "AccountAddress"
                 reference: {
                   entity: "acccounts"
                   key: "account_id"
@@ -325,6 +327,7 @@ metadata-overrides {
               }
               block_id {
                 display-name: "Block Hash"
+                data-type: "Hash"
                 visible: true
                 reference: {
                   entity: "blocks"
@@ -351,6 +354,7 @@ metadata-overrides {
               }
               predecessor {
                 display-name: "Predecessor Hash"
+                data-type: "Hash"
                 visible: true
                 reference: {
                   entity: "blocks"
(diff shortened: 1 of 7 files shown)


src/main/resources/reference.conf
```diff
@@ -201,6 +206,7 @@ metadata-overrides {
               }
               balance {
                 visible: true
+                scale: 6
```
ba205: This just means that balance is microtez, and to get the actual tez, we multiply by 10^6, correct?

ivanopagano: https://github.com/Cryptonomic/Conseil/issues/354#issuecomment-494889684

I think it's explained best there

src/main/resources/reference.conf
```diff
@@ -152,9 +152,13 @@ akka {
...
 # * description: String
 #
-# Additionally, on attributes level you can provide two parameters:
+# Additionally, on attributes level you can provide following parameters:
 # * placeholder: String
 # * data-format: String
+# * value-map: Map[String, String], example: {"0": "Yay", "1": "Nay", "2": "Pass"}
+# * data-type: String, since this value is later mapped to enum, values are limited to those supported by method
+#                      tech.cryptonomic.conseil.tezos.TezosPlatformDiscoveryOperations.mapType
+# * scale: Int
```
ivanopagano: Maybe this could deserve some explanation or, even best, an example?

src/test/scala/tech/cryptonomic/conseil/routes/PlatformDiscoveryTest.scala
```diff
@@ -148,20 +148,31 @@ class PlatformDiscoveryTest extends WordSpec with Matchers with ScalatestRouteTe
...
 
       // when
       Get("/v2/metadata/tezos/mainnet/entity/attributes") ~> addHeader("apiKey", "hooman") ~> sut(overridesConfiguration) ~> check {
 
         // then
         status shouldEqual StatusCodes.OK
         contentType shouldBe ContentTypes.`application/json`
-        val result: List[Map[String, String]] = toListOfMaps[String](responseAs[String])
+        val result: List[Map[String, Any]] = toListOfMaps[Any](responseAs[String])
         result.head("name") shouldBe "attribute"
```
ivanopagano: Trivial: just realized this can get confusing for some: maybe store the head once and reuse it?
```scala
val headResult: Map[String, Any]= toListOfMaps[Any](responseAs[String]).head
headResult("name")...
...
```

src/main/resources/reference.conf
ba205: Thanks!

src/main/resources/reference.conf
dorians: done

src/main/scala/tech/cryptonomic/conseil/generic/chain/PlatformDiscoveryTypes.scala
```diff
@@ -27,13 +27,16 @@ object PlatformDiscoveryTypes {
...
+    valueMap: Option[Map[String, String]] = None,
+    scale: Option[Int] = None,
+    reference: Option[Map[String, String]] = None
   )
 
   /** Enumeration of data types */
   object DataType extends Enumeration {
     type DataType = Value
-    val Enum, Hex, Binary, Date, DateTime, String, Int, LargeInt, Decimal, Boolean = Value
+    val Enum, Hex, Binary, Date, DateTime, String, Hash, Int, LargeInt, Decimal, Boolean = Value
```
anonymoussprocket: Please add "Address" enum value

PR #378, opened 2019-05-24
377 - caching logic update

### LINKED ISSUE #377: Update caching logic
@anonymoussprocket requested following changes to the cache:
- manually initialising counts in cache
- when cache is not initialised it should not return cardinality for the attributes

piotrkosecki:
This PR resolves #377. It should be merged after PR #364 

This PR adds following changes:
- manual init for attribute counts in cache
- endpoints for initialising/status of cache `GET/POST` - `/v2/metadata/cache` (`POST` with empty body to initialise, `GET` to check the status)
- refactor of the caching - extracted it into separate class


src/test/scala/tech/cryptonomic/conseil/routes/PlatformDiscoveryTest.scala
```diff
@@ -227,5 +229,33 @@ class PlatformDiscoveryTest extends WordSpec with Matchers with ScalatestRouteTe
...
+
+    "expose endpoint for cache initialization" in {
+      // given
+      (tezosPlatformDiscoveryOperations.initAttributesCount _).when().returns(())
+
+      // when
+      Post("/v2/metadata/cache") ~> addHeader("apiKey", "hooman") ~> sut(Map.empty) ~> check {
+
+        // then
+        status shouldEqual StatusCodes.OK
```
ivanopagano: Is this enough? Should it check the mock for an initialization call? I'm not sure we have another way to test if something changed from something observable as an http-route...

src/test/scala/tech/cryptonomic/conseil/tezos/MetadataCachingTest.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  val sut: MetadataCaching[IO] = new MetadataCaching[IO](cachingStatus, attributesCache, entitiesCache, attributeValuesCache)
+
+  "Metadata caching" should {
+    "get initial state of caching status" in {
+      sut.getCachingStatus.unsafeRunSync() shouldBe NotStarted
+    }
+
+    "update cache status" in {
+      sut.updateCachingStatus(InProgress).unsafeRunSync()
```
ivanopagano: I'd suggest `sut` to be rebuilt from scratch for each test, otherwise the test cases will be entangled

src/test/scala/tech/cryptonomic/conseil/tezos/MetadataCachingTest.scala
```diff
@@ -0,0 +1,115 @@
...
+      sut.getEntities("not valid").unsafeRunSync() shouldBe None
+      sut.getEntities("example").unsafeRunSync() shouldBe Some(0L, List())
+    }
+
+    "init attribute values cache" in {
+      val attributeValuesCache: AttributeValuesCache = Map("table.column" -> (0L, RadixTree[String, String]()))
+      sut.putAllAttributeValuesIntoEmptyCache(attributeValuesCache).unsafeRunSync()
+
+      sut.getAttributeValues("not valid", "not valid either").unsafeRunSync() shouldBe None
+      val (_, result) = sut.getAttributeValues("table", "column").unsafeRunSync().get
```
ivanopagano: You can mixin `OptionValues`in the spec to get the `.value` method to extract options and a meaningful error if the option's not there

`val (_, result) = sut.getAttributesValues(...).unsafeRunSync.value`

src/test/scala/tech/cryptonomic/conseil/tezos/MetadataCachingTest.scala
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    "insert/update values in entities cache" in {
+      val emptyEntitiesCache: EntitiesCache = Map("example" -> (0L, List()))
+      val entitiesList = List(Entity("a", "b", 0))
+      val updatedEntityList = List(Entity("x", "y", 0))
+      sut.putAllEntitiesIntoEmptyCache(emptyEntitiesCache).unsafeRunSync()
+
+      // insert
+      sut.putEntities("another example", entitiesList).unsafeRunSync()
```
ivanopagano: Please, do you mind swapping `"test network"` or similar to `"another example"` to make it easier to understand what are the expectations from the call? Thanks

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -50,25 +43,25 @@ object TezosPlatformDiscoveryOperations {
...
+  cacheTTL: FiniteDuration,
+  networkName: String = "notUsed")
   (implicit executionContext: ExecutionContext) {
 
   import cats.effect._
   import cats.implicits._
 
+
   /** Method for initializing values of the cache */
   def init(): Future[Unit] = {
```
ivanopagano: I'm getting lost somewhat: who calls `init` and when?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -321,26 +303,77 @@ class TezosPlatformDiscoveryOperations(
...
   private def getUpdatedAttributes(tableName: String, columns: List[Attribute]): Future[List[Attribute]] = {
     metadataOperations.runQuery(getUpdatedAttributesQuery(tableName, columns))
   }
 
+  /** Returns current cache initialization status */
+  def getCacheStatus: Future[CachingStatus] =
+    caching.getCachingStatus.unsafeToFuture()
+
+  /** Inits attribute counts */
+  def initAttributesCount(): Unit = {
```
ivanopagano: Should we return a value so that it can be returned to the client to distinguish if the initialization was actually done or if the cache was already initialized?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -50,25 +43,25 @@ object TezosPlatformDiscoveryOperations {
...
 class TezosPlatformDiscoveryOperations(
   metadataOperations: MetadataOperations,
-  attributesCache: MVar[IO, AttributesCache],
-  entitiesCache: MVar[IO, EntitiesCache],
-  attributeValuesCache: MVar[IO, AttributeValuesCache],
+  caching: MetadataCaching[IO],
   cacheOverrides: AttributeValuesCacheConfiguration,
-  cacheTTL: FiniteDuration)
+  cacheTTL: FiniteDuration,
+  networkName: String = "notUsed")
```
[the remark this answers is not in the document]
piotrkosecki: as we should consider supporting multiple networks and I needed key for the entities cache so I could make it generic I've decided to introduce network name to this class which will be used at some point but at the moment it is just a key for EntitiesCache

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
piotrkosecki: this is the first init which is initializing schema and sizes of tables, next is a manual init which initializes distinct counts(cardinality) of columns.

src/test/scala/tech/cryptonomic/conseil/tezos/MetadataCachingTest.scala
piotrkosecki: I've mixed `with OneInstancePerTest` and it's being handled

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -50,25 +43,25 @@ object TezosPlatformDiscoveryOperations {
 /** Class providing the implementation of the metadata calls with caching */
 class TezosPlatformDiscoveryOperations(
   metadataOperations: MetadataOperations,
-  attributesCache: MVar[IO, AttributesCache],
-  entitiesCache: MVar[IO, EntitiesCache],
-  attributeValuesCache: MVar[IO, AttributeValuesCache],
+  caching: MetadataCaching[IO],
```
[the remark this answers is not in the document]
piotrkosecki: no, I'm against it, because it would mean that we have global mutable state available directly without any idea how it got there. Look at the DbUtils, you can randomly import db, and run query in any class/object in the code. I disagree with this approach.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
piotrkosecki: that's what I assumed `status` endpoint is for

src/test/scala/tech/cryptonomic/conseil/tezos/MetadataCachingTest.scala
piotrkosecki: ok, changed

src/test/scala/tech/cryptonomic/conseil/tezos/MetadataCachingTest.scala
piotrkosecki: ok, you're right

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -321,26 +303,77 @@ class TezosPlatformDiscoveryOperations(
...
-          for {
-            updatedAttributes <- IO.fromFuture(IO(getUpdatedAttributes(tableName, attributes)))
-            _ <- attributesCache.take
-            _ <- attributesCache.put(entitiesMap.updated(tableName, now -> updatedAttributes))
-          } yield updatedAttributes
-        }
-      }.sequence
-    }.unsafeToFuture()
+    caching.getCachingStatus.map { status =>
+      if (status == Finished) {
```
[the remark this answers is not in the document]
piotrkosecki: I'm not sure how would you want to do that

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -321,26 +303,77 @@ class TezosPlatformDiscoveryOperations(
...
+          entCache <- caching.getEntities(networkName)
+          attributes <- caching.getAllAttributes
+          updatedAttributes <- IO.fromFuture(IO(getAllUpdatedAttributes(entCache, attributes)))
+        } yield
+          updatedAttributes.map {
+            case (tableName, attr) =>
+              caching.putAttributes(tableName, attr)
+          }.sequence
+
+        for {
```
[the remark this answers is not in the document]
piotrkosecki: ok, changed to you first suggestion, never used the bracket, I'll need to read about it

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -321,26 +303,77 @@ class TezosPlatformDiscoveryOperations(
...
+          _ <- caching.updateCachingStatus(Finished)
+        } yield ()
+      } else {
+        IO.pure(())
+      }
+    }.unsafeRunAsyncAndForget()
+  }
+
+  /** Helper method for updating */
+  private def getAllUpdatedAttributes(entCache: Option[(LastUpdated, List[Entity])], attributes: MetadataCache[List[Attribute]]):
```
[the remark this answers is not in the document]
piotrkosecki: ok, you are right, thanks, changed according to your suggestions

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -321,26 +303,77 @@ class TezosPlatformDiscoveryOperations(
...
   }
 
   /** Runs query and attributes with updated counts */
   private def getUpdatedAttributes(tableName: String, columns: List[Attribute]): Future[List[Attribute]] = {
     metadataOperations.runQuery(getUpdatedAttributesQuery(tableName, columns))
   }
 
+  /** Returns current cache initialization status */
+  def getCacheStatus: Future[CachingStatus] =
+    caching.getCachingStatus.unsafeToFuture()
```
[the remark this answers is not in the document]
piotrkosecki: renamed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -321,26 +303,77 @@ class TezosPlatformDiscoveryOperations(
...
-    }.unsafeToFuture()
+    caching.getCachingStatus.map { status =>
+      if (status == Finished) {
+        caching.getAttributes(tableName).flatMap { attributesOpt =>
+          attributesOpt.map { case (last, attributes) =>
+            if (last + cacheTTL.toMillis > now) {
+              IO.pure(attributes)
+            } else {
+              for {
+                updatedAttributes <- IO.fromFuture(IO(getUpdatedAttributes(tableName, attributes)))
```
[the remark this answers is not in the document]
piotrkosecki: ok, I didn't know that one

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -321,26 +303,77 @@ class TezosPlatformDiscoveryOperations(
...
-            _ <- attributesCache.put(entitiesMap.updated(tableName, now -> updatedAttributes))
-          } yield updatedAttributes
-        }
-      }.sequence
-    }.unsafeToFuture()
+    caching.getCachingStatus.map { status =>
+      if (status == Finished) {
+        caching.getAttributes(tableName).flatMap { attributesOpt =>
+          attributesOpt.map { case (last, attributes) =>
+            if (last + cacheTTL.toMillis > now) {
```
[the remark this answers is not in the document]
piotrkosecki: ok, extracted

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -213,33 +175,60 @@ class TezosPlatformDiscoveryOperations(
...
+  private def getTablesCount(tables: Vector[MTable]): DBIO[Vector[Int]] = {
+    DBIOAction.sequence {
+      tables.map { table =>
+        TezosDatabaseOperations.countRows(table.name.name)
+      }
+    }
+  }
+
+  /** Returns current time in milliseconds */
+  private def now: Long = System.currentTimeMillis()
```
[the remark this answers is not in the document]
piotrkosecki: it was, but I removed it

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -213,33 +175,60 @@ class TezosPlatformDiscoveryOperations(
     */
   def getEntities: Future[List[Entity]] = {
     val result = for {
-      entities <- entitiesCache.read
-      res <- if (entities._1 + cacheTTL.toMillis > now) {
-        IO.pure(entities._2)
+      entities <- caching.getEntities(networkName)
+      res <- if (entities.get._1 + cacheTTL.toMillis > now) {
```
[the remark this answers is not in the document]
piotrkosecki: yes, forgot to refactor it after testing, changed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -157,40 +138,20 @@ class TezosPlatformDiscoveryOperations(
...
+          }
       }
-    }
+    }.map(_.toMap)
   }
 
-  /** Returns current time in milliseconds */
-  private def now: Long = System.currentTimeMillis()
+  /** Makes key out of table and column names */
+  private def makeKey(table: String, column: String): String = s"$table.$column"
```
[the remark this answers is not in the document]
piotrkosecki: ok, extracted into one place

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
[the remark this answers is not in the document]
piotrkosecki: ok, used mapN

src/main/scala/tech/cryptonomic/conseil/tezos/MetadataCaching.scala
```diff
@@ -0,0 +1,142 @@
...
+  /** Generic method for putting value into cache */
+  private def putIntoCache[A](key: String, value: A)(cache: MVar[F, MetadataCache[A]]): F[Unit] = {
+    for {
+      ca <- cache.take
+      _ <- cache.put(ca.updated(key, (now, value)))
+    } yield ()
+  }
+
+  /** Returns current time in milliseconds */
+  private def now: Long = System.currentTimeMillis()
```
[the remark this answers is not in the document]
piotrkosecki: ok, changed to `nanoTime`

src/main/scala/tech/cryptonomic/conseil/tezos/MetadataCaching.scala
```diff
@@ -0,0 +1,142 @@
...
+      _ <- cache.put(values)
+    } yield ()
+  }
+
+  /** Inserts all entities into empty cache */
+  def putAllEntitiesIntoEmptyCache(entities: EntitiesCache): F[Unit] =
+    putAllIntoEmptyCache(entities)(entitiesCache)
+
+  /** Generic method inserting all values into cache */
+  private def putAllIntoEmptyCache[A](values: MetadataCache[A])(cache: MVar[F, MetadataCache[A]]): F[Unit] =
```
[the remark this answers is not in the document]
piotrkosecki: renamed and changed

src/main/scala/tech/cryptonomic/conseil/tezos/MetadataCaching.scala
```diff
@@ -0,0 +1,142 @@
...
+
+  import cats.implicits._
+
+  /** Reads current caching status */
+  def getCachingStatus: F[CachingStatus] =
+    cachingStatus.read
+
+  /** Updates caching status */
+  def updateCachingStatus(status: CachingStatus): F[Unit] = {
+    for {
```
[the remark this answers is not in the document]
piotrkosecki: ok, changed

src/main/scala/tech/cryptonomic/conseil/tezos/MetadataCaching.scala
```diff
@@ -0,0 +1,142 @@
...
+  def getAttributes(entity: String): F[Option[(LastUpdated, List[Attribute])]] =
+    getFromCache(entity)(attributesCache)
+
+  /** Generic method for getting value from cache */
+  private def getFromCache[A](key: String)(cache: MVar[F, MetadataCache[A]]): F[Option[(LastUpdated, A)]] = {
+    cache.read.map(_.get(key))
+  }
+
+  /** Reads all attributes from cache */
+  def getAllAttributes: F[AttributesCache] =
```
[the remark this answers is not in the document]
piotrkosecki: sure, changed

src/main/scala/tech/cryptonomic/conseil/tezos/MetadataCaching.scala
```diff
@@ -0,0 +1,142 @@
...
+  /** Updates caching status */
+  def updateCachingStatus(status: CachingStatus): F[Unit] = {
+    for {
+      _ <- cachingStatus.take
+      _ <- cachingStatus.put(status)
+    } yield ()
+  }
+
+  /** Reads entities from cache */
+  def getEntities(network: String): F[Option[(LastUpdated, List[Entity])]] =
```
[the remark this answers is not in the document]
piotrkosecki: changed, thanks

src/main/scala/tech/cryptonomic/conseil/tezos/MetadataCaching.scala
```diff
@@ -0,0 +1,142 @@
...
+
+  case object InProgress extends CachingStatus
+
+  case object Finished extends CachingStatus
+
+}
+
+/** Class providing caching for metadata */
+class MetadataCaching[F[_]](
+  cachingStatus: MVar[F, CachingStatus],
```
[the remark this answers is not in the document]
piotrkosecki: ok, I haven't used it before, but it seems its similar to MVar, changed

src/main/scala/tech/cryptonomic/conseil/tezos/MetadataCaching.scala
```diff
@@ -0,0 +1,142 @@
...
+object MetadataCaching {
+  type LastUpdated = Long
+
+  type MetadataCache[A] = Map[String, (LastUpdated, A)]
+  type AttributesCache = MetadataCache[List[Attribute]]
+  type AttributeValuesCache = MetadataCache[RadixTree[String, String]]
+  type EntitiesCache = MetadataCache[List[Entity]]
+
+  /** Cache initialization statuses */
+  sealed trait CachingStatus
```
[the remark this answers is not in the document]
piotrkosecki: ok

src/main/scala/tech/cryptonomic/conseil/routes/openapi/PlatformDiscoveryEndpoints.scala
```diff
@@ -14,6 +14,29 @@ trait PlatformDiscoveryEndpoints
...
+      ),
+      response = emptyResponse(),
+      tags = List("Cache")
+    )
+
+  /** Endpoint for cache initialization status */
+  def cacheStatusEndpoint: Endpoint[String, String] =
+  endpoint(
+    request = get(
+      url = commonPath / "cache",
```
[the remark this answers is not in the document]
piotrkosecki: ok, changed to `cache/status`

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
```diff
@@ -17,6 +18,14 @@ class MetadataService(config: PlatformsConfiguration,
                       cacheConfiguration: AttributeValuesCacheConfiguration,
                       tezosPlatformDiscoveryOperations: TezosPlatformDiscoveryOperations) {
 
+  // inits attributes cache
+  def initAttributesCache(): Unit =
+    tezosPlatformDiscoveryOperations.initAttributesCount()
+
+  // fetches current caching status
+  def attributesCacheStatus: Future[CachingStatus] =
```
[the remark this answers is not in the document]
piotrkosecki: yes, renamed

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -43,19 +43,26 @@ object Conseil extends App with LazyLogging with EnableCORSDirectives with Conse
...
-      val attributesCache = MVar[IO].empty[AttributesCache].unsafeRunSync()
-      val entitiesCache = MVar[IO].empty[EntitiesCache].unsafeRunSync()
-      val attributeValuesCache = MVar[IO].empty[AttributeValuesCache].unsafeRunSync()
+      val metadataCaching = {
+        for {
+          cachingStatus <- MVar[IO].of[CachingStatus](NotStarted)
+          attributesCache <- MVar[IO].empty[AttributesCache]
+          entitiesCache <- MVar[IO].empty[EntitiesCache]
+          attributeValuesCache <- MVar[IO].empty[AttributeValuesCache]
+        } yield new MetadataCaching[IO](cachingStatus, attributesCache, entitiesCache, attributeValuesCache)
```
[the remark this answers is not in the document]
piotrkosecki: you're right, I haven't thought of it, but it's a good idea, moved

src/test/scala/tech/cryptonomic/conseil/routes/PlatformDiscoveryTest.scala
piotrkosecki: I think it's enough, it just checks if endpoint exists and responds to the correct request

src/main/scala/tech/cryptonomic/conseil/tezos/MetadataCaching.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  /** Returns current time in milliseconds */
+  def now: Long = System.nanoTime()
+
+  /** Makes key out of table and column names */
+  def makeKey(table: String, column: String): String = s"$table.$column"
+}
+
+/** Class providing caching for metadata */
+class MetadataCaching[F[_]](
```
dorians: Have you checked [ScalaCache](https://github.com/cb372/scalacache)? It supports [Caffeine](https://github.com/blemale/scaffeine) which has support for [cache refreshing](https://github.com/ben-manes/caffeine/wiki/Refresh).

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
```diff
@@ -17,6 +18,14 @@ class MetadataService(config: PlatformsConfiguration,
                       cacheConfiguration: AttributeValuesCacheConfiguration,
                       tezosPlatformDiscoveryOperations: TezosPlatformDiscoveryOperations) {
 
+  // inits attributes cache
+  def initAttributesCache(): Unit =
```
dorians: Does it support deployment on many instances right now? In order to initialize the cache, we'll need to hit each instance which could be problematic.

src/main/scala/tech/cryptonomic/conseil/tezos/MetadataCaching.scala
```diff
@@ -0,0 +1,142 @@
...
+import cats.effect.concurrent.MVar
+import com.rklaehn.radixtree.RadixTree
+import tech.cryptonomic.conseil.generic.chain.PlatformDiscoveryTypes.{Attribute, Entity}
+import tech.cryptonomic.conseil.tezos.MetadataCaching._
+
+/** Companion object providing useful types related with MetadataCaching */
+object MetadataCaching {
+  type LastUpdated = Long
+
+  type MetadataCache[A] = Map[String, (LastUpdated, A)]
```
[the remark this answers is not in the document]
piotrkosecki: ok, changed and renamed

src/main/scala/tech/cryptonomic/conseil/tezos/MetadataCaching.scala
piotrkosecki: At the point when I started implementing cache by myself the logic required custom solution.
I'll look into it and check if ScalaCache can help now.

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
piotrkosecki: why not? you just use instances IP to init cache

dorians: Yeah, but it doesn't scale well since you need to be aware of IP addresses of all machines. The idea of load balancing is to treat the whole application as if it was a single instance.

piotrkosecki: I agree, but this operation either way should be done once. That was a requirement for those changes.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
ivanopagano: It could turn out useful anyway to have it here directly?
Or not exactly the same, it could be a (boolean?) return type that actually tells you if there was a need to initialize or not, so the caller could decide based on that, without the need to make a double check before and after the init call

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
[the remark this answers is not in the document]
ivanopagano: ok, take your time. As a brief overview, it's a method that allows you to do pre and post operations which are guaranteed, around some lambda that uses the resource.
Think like: take the db connection, use it, guarantee that it's closed right after

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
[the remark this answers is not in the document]
ivanopagano: I guess moving this all status checking and cache expiration internally to the `caching` object, and it's always there that the new data would be loaded as needed, hiding everything from the PlatDiscOp service...

It might be less ideal than I'm imagining, I might be missing some detail. I'm only a bit troubled that the Operations class should explicitly drive manually the caching, based on a public status... usually the cache should hide those details from usage

src/test/scala/tech/cryptonomic/conseil/routes/PlatformDiscoveryTest.scala
ivanopagano: fine then

src/test/scala/tech/cryptonomic/conseil/tezos/MetadataCachingTest.scala
ivanopagano: awesome, didn't knew about that

src/main/scala/tech/cryptonomic/conseil/tezos/MetadataCaching.scala
```diff
@@ -10,25 +11,47 @@ import tech.cryptonomic.conseil.tezos.MetadataCaching._
...
-  type AttributesCache = MetadataCache[List[Attribute]]
-  type AttributeValuesCache = MetadataCache[RadixTree[String, String]]
-  type EntitiesCache = MetadataCache[List[Entity]]
+  /** Class representing entry in cache */
+  final case class CacheEntry[A](lastUpdated: LastUpdated, value: A)
+
+  /** Class representing key in cache */
+  final case class CacheKey(key: String)
+
+  type Cache[A] = Map[CacheKey, CacheEntry[A]]
```
ivanopagano: Maybe @dorians won't think the same, but it makes so much more sense to me 😍

src/main/scala/tech/cryptonomic/conseil/tezos/MetadataCaching.scala
```diff
@@ -10,25 +11,47 @@ import tech.cryptonomic.conseil.tezos.MetadataCaching._
...
+  def empty[F[_]](implicit concurrent: Concurrent[F]): F[MetadataCaching[F]] =
+    for {
+    cachingStatus <- Ref[F].of[CachingStatus](NotStarted)
+    attributesCache <- MVar[F].empty[AttributesCache]
+    entitiesCache <- MVar[F].empty[EntitiesCache]
+    attributeValuesCache <- MVar[F].empty[AttributeValuesCache]
+  } yield new MetadataCaching[F](cachingStatus, attributesCache, entitiesCache, attributeValuesCache)
+
+  /** Returns current time in milliseconds */
+  def now: Long = System.nanoTime()
```
ivanopagano: There are ways to handle this too in a purely functional fashion with cats clock, yet I would leave it like this for now.
Just to take note

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -49,22 +48,23 @@ class TezosPlatformDiscoveryOperations(
...
-    } yield ()
-
-    result.unsafeToFuture()
+    val entities = IO.fromFuture(IO(metadataOperations.runQuery(preCacheEntities)))
+    val attributes = IO.fromFuture(IO(metadataOperations.runQuery(preCacheAttributes)))
+    val attributeValues = IO.fromFuture(IO(metadataOperations.runQuery(preCacheAttributeValues)))
+
+    (entities, attributes, attributeValues).mapN { case (ent, attr, attrValues) =>
+      (caching.fillEntitiesCache(ent), caching.fillAttributesCache(attr), caching.fillAttributeValuesCache(attrValues))
+        .mapN {
```
ivanopagano: I didn't realize at first that the operations were bound in paired locksteps.
Maybe it would read better if the dependent operations were actually flatMapped as needed
i.e.
```scala
(entities flatMap caching.fillEntitiesCache, attributes flatMap caching.fillAttributesCache, attributeValues flatMap caching.fillAttributeValuesCache)
  .tuple3 //same as mapN (possibly there's even a more generic .tuple) but with no need to apply any extra function
  .void // keep the container (IO) and discards the values (puts Unit)
```

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -176,14 +180,16 @@ class TezosPlatformDiscoveryOperations(
...
       entities <- caching.getEntities(networkName)
-      res <- if (entities.get._1 + cacheTTL.toMillis > now) {
-        IO.pure(entities.get._2)
-      } else {
-        for {
-          updatedEntities <- IO.fromFuture(IO(metadataOperations.runQuery(preCacheEntities)))
-          _ <- caching.putAllEntities(updatedEntities)
-        } yield updatedEntities(networkName)._2
-      }
+      res <- entities.toList.map { case CacheEntry(last, ent) =>
```
ivanopagano: mechanical translation

`.map(...).sequence` is always `.traverse(...)`

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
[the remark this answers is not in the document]
ivanopagano: Maybe I wasn't too clear.
I simply meant that, since nobody outside of this class seems to use the instance passed in as a caching parameter, we might completely encapsulate the fact that a cache is built, create it with the PlatformDiscoveryOps, and use it only internally

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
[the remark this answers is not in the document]
piotrkosecki: ok, I guess you're right but i think it will be bigger change, I propose to do it in the next caching PR

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
piotrkosecki: ok, now it returns current status

src/main/scala/tech/cryptonomic/conseil/tezos/MetadataCaching.scala
piotrkosecki: ok, noted

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
piotrkosecki: ok, changed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
piotrkosecki: true, didn't notice that, changed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
ivanopagano: I actually meant not to return the "current status", but if the "status changed", that is
- true if the values weren't there before and now they are
- false if nothing changed...

or even a pair of (oldStatus, newStatus), later wrapped as some json value to be inspected

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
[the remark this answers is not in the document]
piotrkosecki: At this point I think I like cache being used explicitly. Maybe when I'll try to move more logic to cache it will be more appropriate, but we'll see.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -175,10 +175,12 @@ class TezosPlatformDiscoveryOperations(
...
           IO.pure(ent)
         } else {
-          for {
+          (for {
+            _ <- caching.putEntities(networkName, ent)
             updatedEntities <- IO.fromFuture(IO(metadataOperations.runQuery(preCacheEntities)))
             _ <- caching.putAllEntities(updatedEntities)
-          } yield updatedEntities(CacheKey(networkName)).value
+          } yield ()).unsafeRunAsyncAndForget()
+          IO.pure(ent)
```
ivanopagano: Trying to understand if I got it correctly:
is it essentially returning immediately, while the cache is updated in another thread? 

Shouldn't the code call `IO.contextShift` before running the async IO, to have it run on a different thread?

In any case the `IO.pure(ent)` is returned... hence a single `if` clause (side-effecting) is needed, with code to refresh the case, and then we should return the value anyway, something like

```scala
if (cacheExpired(last) {
  for { ... }.unsafeRunAsyncAndThrowItInTheToilet()
}
IO.pure(ent) // or, equivalently: ent.pure
```

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -249,12 +251,13 @@ class TezosPlatformDiscoveryOperations(
...
       case Some(CacheEntry(_, oldRadixTree)) =>
-        for {
+        (for {
           _ <- caching.putAttributeValues(tableName, columnName, oldRadixTree)
           attributeValues <- IO.fromFuture(IO(makeAttributesQuery(tableName, columnName, None)))
           radixTree = RadixTree(attributeValues.map(x => x.toLowerCase -> x): _*)
           _ <- caching.putAttributeValues(tableName, columnName, radixTree)
-        } yield radixTree.filterPrefix(attributeFilter).values.take(maxResultLength).toList
+        } yield ()).unsafeRunAsyncAndForget()
+        IO.pure(oldRadixTree.filterPrefix(attributeFilter).values.take(maxResultLength).toList)
```
ivanopagano: Same exact considerations I left for `getEntities`

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -290,7 +293,12 @@ class TezosPlatformDiscoveryOperations(
             if (!cacheExpired(last)) {
               IO.pure(attributes)
             } else {
-              IO.fromFuture(IO(getUpdatedAttributes(tableName, attributes))).flatTap(caching.putAttributes(tableName, _))
+              (for {
+                _ <- caching.putAttributes(tableName, attributes)
+                updatedAttributes <- IO.fromFuture(IO(getUpdatedAttributes(tableName, attributes)))
+                _ <- caching.putAttributes(tableName, updatedAttributes)
+              } yield ()).unsafeRunAsyncAndForget()
+              IO.pure(attributes)
```
ivanopagano: Same exact considerations I left for `getEntities`

PR #380, opened 2019-05-27
379 - configuration fix

dorians:


src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -36,6 +36,7 @@ object TezosPlatformDiscoveryOperations {
       case "numeric" => DataType.Decimal
       case "bool" => DataType.Boolean
       case "hash" => DataType.Hash
+      case "accountAddress" => DataType.AccountAddress
```
ivanopagano: Do we really want this to be case-sensitive in the end?
@dorians @anonymoussprocket ?

anonymoussprocket: it just following the other values, Boolean, Decimal, etc. I figure camel case is appropriate for a two-word value. I'm open to suggestion however depending on what common enum naming practices are in scala.

ivanopagano: If we want to make it consistent with typenames casing in scala, we should go CamelCase.

But I wonder if we might want to follow the so-called [Postel's Law](https://en.wikipedia.org/wiki/Robustness_principle) about being more robust to accept values

dorians: I would opt for using a different field. This method `maps type from DB to type used in query` which might be confusing.

ivanopagano: I don't get it, isn't this what we're trying to do? Map query types, with aliasing, to actual database types?
Is this used for a different sort of operation?

dorians: It's used for metadata overrides and it's not related with database types directly.

PR #383, opened 2019-05-27
382 - Remove all usage of ApiFiltering and Tezos routes

### LINKED ISSUE #382: Remove support for V1 endpoints
This is a proposal to be considered, if the time is ripe to remove somewhat older code that is needed only to have support for the API v1 query endpoints.

Downsides to stop such support mainly impact old versions of Galleon, and what else?

Pros are a significant removal of code to maintain, and a natural increase in code-coverage, as we don't spend time to actually try to test that code, I think just because it's planned for "decommissioning"

ivanopagano:
Closes #382 

This is essentially achieved through removal of all code that used the ApiFiltering trait and its implicit usage, and then removing the dependent "v1" routes for tezos path.

No database or configuration change is needed


ivanopagano:
please let's wait for @piotrkosecki to have a look, he was the main actor in moving from v1 to v2, and I'd prefer him to check if I didn't lose anything with this changes

PR #392, opened 2019-05-30
390 - allow multiple aggregated fields

### LINKED ISSUE #390: Allow multiple aggregated fields
Convert the aggregation parameter into an array. Programmatically rename the returner aggregated columns to function_field. 

```
"aggregation": [
    {
    "field": "balance",                                                                                          
    "function": "sum"
  }, {
    "field": "balance",                                                                                          
    "function": "avg"
  }]
```

The above query would return a `sum_balance` and `avg_balance` columns. The query should also allow

```
  "orderBy": [
    {
      "field": "sum_balance",
      "direction": "asc"
    }
  ]
```

piotrkosecki:
This PR closes #390. 

In this PR I changed optional aggregation field to a list, which allows aggregating on multiple fields.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
@@ -158,7 +158,7 @@ object DataTypes {
     orderBy: List[QueryOrdering] = List.empty,
     limit: Int = defaultLimitValue,
     output: OutputType = OutputType.json,
-    aggregation: Option[Aggregation] = None
+    aggregation: List[Aggregation] = List.empty
   )
 
   /** Class representing predicate used in aggregation */
@@ -205,7 +205,7 @@ object DataTypes {
     orderBy: Option[List[QueryOrdering]],
     limit: Option[Int],
     output: Option[OutputType],
-    aggregation: Option[ApiAggregation] = None
+    aggregation: List[ApiAggregation] = List.empty
   ) {
     /** Method which validates query fields */
     def validate(entity: String, tezosPlatformDiscovery: TezosPlatformDiscoveryOperations)(implicit ec: ExecutionContext):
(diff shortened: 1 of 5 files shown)


src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -205,7 +205,7 @@ object DataTypes {
     orderBy: Option[List[QueryOrdering]],
     limit: Option[Int],
     output: Option[OutputType],
-    aggregation: Option[ApiAggregation] = None
+    aggregation: Option[List[ApiAggregation]]
```
ivanopagano: why not a `: List[ApiAggregation] = List.empty`here as in `Query`?

piotrkosecki: because I would need to make a custom decoder for query as it endpoints generated cannot handle default values

src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
```diff
@@ -102,11 +102,12 @@ object DatabaseUtil {
         * @param columns     parameter containing columns which chich are being used in query
         * @return new SQLActionBuilder containing limit statement
         */
-      def addGroupBy(aggregation: Option[Aggregation], columns: List[String]): SQLActionBuilder = {
-        aggregation.fold(action) {
-          aggregates =>
-            val cols = columns.filterNot(_ == aggregates.field)
-          concatenateSqlActions(action, makeGroupBy(cols))
+      def addGroupBy(aggregation: List[Aggregation], columns: List[String]): SQLActionBuilder = {
+        val aggregationFields = aggregation.map(_.field).toSet
```
dorians: You can move that `val` to be inside `else` statement.

piotrkosecki: ok, done

PR #406, opened 2019-06-06
389 - Query predicate requirements

piotrkosecki:
Implemented possibility of adding requirements for queries to use key/datetime fields.

In configuration there will be possibility of adding 2 fields:
- `limited-query` - boolean, on entity level, which will forbid making queries if predicate does not contain any key/datetime attributes
- `sufficient-for-query` - boolean, on attribute level, which will tell us to treat given attribute as key field


src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -88,6 +88,41 @@ object DataTypes {
...
+          }.map(attributes => doesPredicateContainValidAttribute(attributes.flatten))
+      } else {
+        Future.successful(List.empty)
+      }
+    }
+  }
+
+  /** Checks if any of the attributes is valid for predicate */
+  private def doesPredicateContainValidAttribute(attributes: List[Attribute]): List[InvalidPredicateFiltering] = {
+    if(attributes.exists(attr => attr.keyType == KeyType.UniqueKey || attr.dataType == DataType.DateTime || attr.sufficientForQuery.getOrElse(false))) {
```
dorians: I would move that logic to a method in an `Attribute` class.

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
```diff
@@ -17,6 +17,10 @@ class MetadataService(config: PlatformsConfiguration,
                       cacheConfiguration: AttributeValuesCacheConfiguration,
                       tezosPlatformDiscoveryOperations: TezosPlatformDiscoveryOperations) {
 
+  // checks if attribute is valid
+  def isAttributeValid(tableName: String, columnName: String): Future[Boolean] =
```
dorians: You're mixing levels here. In method name is `attribute` but in parameters list, you refer to `tableName` and `columnName`.

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
```diff
@@ -41,12 +45,23 @@ class MetadataService(config: PlatformsConfiguration,
...
+    getAttributesHelper(path)(tezosPlatformDiscoveryOperations.getTableAttributesWithoutUpdatingCache)
+
+  // fetches attributes with given function
+  private def getAttributesHelper(path: EntityPath)(getAttributes: String => Future[Option[List[Attribute]]])
+    (implicit apiExecutionContext: ExecutionContext): Future[Option[List[Attribute]]] = {
+    for {
+      exists <- exists(path)
+      attributes <- getAttributes(path.entity)
+    } yield attributes
+      .filter { _ => exists }
```
dorians: To be more exact you can move this `filter` to a `for` section as an `if` statement.

src/test/scala/tech/cryptonomic/conseil/tezos/DataTypesTest.scala
```diff
@@ -72,22 +86,23 @@ import scala.concurrent.ExecutionContext.Implicits.global
...
+        displayName = "invalid",
         dataType = DataType.Int,
         cardinality = None,
-        keyType = KeyType.NonKey,
+        keyType = KeyType.UniqueKey,
         entity = "test"
       )
-      (tpdo.isAttributeValid _).when("test", "invalid").returns(Future.successful(false))
-      (tpdo.getTableAttributesWithoutUpdatingCache _).when("test").returns(Future.successful(Some(List(attribute))))
+      (ms.isAttributeValid _).when("testEntity", "invalid").returns(Future.successful(false))
```
dorians: Maybe not for now but I would consider replacing those heavy stubbing with some real implementation dedicated for tests when we can provide a configuration in a more clean way.

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
piotrkosecki: are you sure? I think because this is a `Future` in for-comprehension and if this `if` statement fails every time we will get `Fututre.failed` instead of `Future.successful(None)`

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
piotrkosecki: your're right, changed

src/test/scala/tech/cryptonomic/conseil/tezos/DataTypesTest.scala
```diff
@@ -213,23 +243,52 @@ import scala.concurrent.ExecutionContext.Implicits.global
         aggregation = Some(List(ApiAggregation(field = "invalid")))
       )
 
-      val result = query.validate("test", tpdo)
+      val result = query.validate(testEntityPath, ms)
 
       result.futureValue.left.get should contain theSameElementsAs List(InvalidAggregationField("invalid"), InvalidAggregationFieldForType("invalid"))
     }
 
+    "return error with field XXXXX" in {
```
ivanopagano: 🤔 ??

src/test/scala/tech/cryptonomic/conseil/tezos/DataTypesTest.scala
```diff
@@ -213,23 +243,52 @@ import scala.concurrent.ExecutionContext.Implicits.global
...
+        predicates = Some(List(ApiPredicate("InvalidAttribute", OperationType.in))),
+        orderBy = None,
+        limit = None,
+        output = None,
+        aggregation = None
+      )
+
+      val result = query.validate(testEntityPath, ms)
+
+      result.futureValue.left.get shouldBe List(InvalidPredicateFiltering("None of attributes InvalidAttribute is valid for predicate"))
```
ivanopagano: This might be arguable, but I find that hardcoding the expected error message in tests might prove painful to maintain with future code changes. WDYT?

piotrkosecki: can scalatest do the comparison somehow? or did you have something else in mind?

src/test/scala/tech/cryptonomic/conseil/tezos/DataTypesTest.scala
piotrkosecki: my mistake 😉

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -221,15 +259,17 @@ object DataTypes {
...
-      val invalidTypeAggregationField = findInvalidAggregationTypeFields(query, entity, tezosPlatformDiscovery)
+      val nonExistingFields = findNonExistingFields(query, entity, metadataService)
+      val invalidTypeAggregationField = findInvalidAggregationTypeFields(query, entity, metadataService)
+      val invalidPredicateFiltering = findInvalidPredicateFilteringFields(query, entity, metadataService)
 
       for {
         invalidNonExistingFields <- nonExistingFields
         invalidAggregationFieldForTypes <- invalidTypeAggregationField
-        updatedQuery <- replaceTimestampInPredicates(entity, query, tezosPlatformDiscovery)
+        updatedQuery <- replaceTimestampInPredicates(entity, query, metadataService)
```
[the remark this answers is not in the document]

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -88,6 +88,41 @@ object DataTypes {
...
+      }
+    }
+  }
+
+  /** Checks if any of the attributes is valid for predicate */
+  private def doesPredicateContainValidAttribute(attributes: List[Attribute]): List[InvalidPredicateFiltering] = {
+    if(attributes.exists(attr => attr.keyType == KeyType.UniqueKey || attr.dataType == DataType.DateTime || attr.sufficientForQuery.getOrElse(false))) {
+      List.empty
+    } else {
+      List(InvalidPredicateFiltering(s"None of attributes ${attributes.map(_.name).mkString(",")} is valid for predicate"))
```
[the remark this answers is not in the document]

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
piotrkosecki: ok, changed

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -88,6 +88,41 @@ object DataTypes {
...
   }
 
+  /** Helper method for finding if queries does not contain filters on key fields or datetime fields */
+  private def findInvalidPredicateFilteringFields(query: Query, path: EntityPath, metadataService: MetadataService)
+    (implicit ec: ExecutionContext): Future[List[InvalidPredicateFiltering]] = {
+    metadataService.getEntities(path.up).flatMap { entitiesOpt =>
+      val limitedQueryEntity = entitiesOpt
+        .toList
+        .flatten
+        .find(_.name == path.entity)
```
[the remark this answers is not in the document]
piotrkosecki: I don't think I've found a good way to do it using any of those method

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -88,6 +88,41 @@ object DataTypes {
...
+        .flatMap(_.limitedQuery)
+        .getOrElse(false)
+      if(limitedQueryEntity) {
```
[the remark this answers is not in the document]
piotrkosecki: I don't think I've found a good way to do it using `fold` , what did you have in mind exactly?

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: moved to Attribute class, regarding Options - @dorians correct me if I'm wrong - I think it is related to config deserialisation into case classes or it just reflect more accurate whats going on there.

src/test/scala/tech/cryptonomic/conseil/tezos/DataTypesTest.scala
ivanopagano: `... shouldBe a[List[InvalidPredicateFiltering]]`

or something similar, i.e. only checking the error type, not the message

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
```diff
@@ -50,12 +54,25 @@ class MetadataService(config: PlatformsConfiguration,
...
+  // fetches table attributes without updating cache
+  def getTableAttributesWithoutUpdatingCache(path: EntityPath)(implicit apiExecutionContext: ExecutionContext): Future[Option[List[Attribute]]] =
+    getAttributesHelper(path)(tezosPlatformDiscoveryOperations.getTableAttributesWithoutUpdatingCache)
+
+  // fetches attributes with given function
+  private def getAttributesHelper(path: EntityPath)(getAttributes: String => Future[Option[List[Attribute]]])
+    (implicit apiExecutionContext: ExecutionContext): Future[Option[List[Attribute]]] = {
+    exists(path).flatMap {
+      case true =>
+      getAttributes(path.entity).map(
```
ivanopagano: Cosmetic: wrong indentation in these cases

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -221,20 +251,16 @@ object DataTypes {
...
-      } yield {
-        invalidNonExistingFields ::: invalidAggregationFieldForTypes match {
-          case Nil => Right(updatedQuery)
-          case wrongFields => Left(wrongFields)
-        }
+      (findNonExistingFields(query, entity, metadataService),
+      findInvalidAggregationTypeFields(query, entity, metadataService),
+      findInvalidPredicateFilteringFields(query, entity, metadataService),
+      replaceTimestampInPredicates(entity, query, metadataService)).mapN {
+        case (invalidNonExistingFields, invalidAggregationFieldForTypes, invalidPredicateFilteringFields, updatedQuery) =>
```
ivanopagano: Triviality: not sure, but you might not need to use a `case` to deconstruct the input, the input should be multliple args, and not a tuple... could be wrong though

src/test/scala/tech/cryptonomic/conseil/tezos/DataTypesTest.scala
piotrkosecki: I've got this kind of warning about type erasure:
`Type parameter should not be specified because it will be erased at runtime, please use _ instead.  Note that in future version of ScalaTest this will give a compiler error.`, I won't change it for now.

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: yes, you're right, changed

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
piotrkosecki: reformatted

src/test/scala/tech/cryptonomic/conseil/tezos/DataTypesTest.scala
ivanopagano: yes, you can solve this by something equivalent like
```scala
val errors = result.futureValue.left.get
errors should have size 1
errors.head shouldBe a[InvalidPredicateFiltering]
```

piotrkosecki: yes, done it

PR #407, opened 2019-06-10
400 - Use relevant blockhash when loading accounts and delegates details

### LINKED ISSUE #400: Use the relevant block reference to download lorre's accounts and delegates
When contacting the tezos node and loading accounts and delegates data, Lorre is currently ignoring the block that was actually referencing those entities, and simply creates the query endpoint location based on the latest head found on a previous call to tezos.

This means that, as we load batches of data from the chain, we are inconsistently saving the latest version of each account and delegate, while the highest-level block saved is still far lower in the chain, thus presenting an inconsistent picture of the whole system.

Lorre should use the correct hash reference to load those entities

ivanopagano:
Fixes #400 

The change loads accounts data and delegates data, using an endpoint based on the actual blockhash that referred the specific account, as to load the data as it was when the block was included in the chain.
This should improve consistency of the stored data, considering that now Lorre may consumes nodes' data only partially, in case of any unexpected issue.
After the change we should always expect conseil to provide a consistent picture of the chain.

As a side-effect, this change correctly implement the option to let Lorre proceed with the next cycle even in the case of a failure during block fetching.
This behaviour is now correctly activated by setting the environment variable
`LORRE_FAILURE_IGNORE` to `true` of `yes`

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -280,8 +280,6 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
       checkpoints =>
         logger.debug("I loaded all stored account references and will proceed to fetch updated information from the chain")
         val (pages, total) = tezosNodeOperator.getAccountsForBlocks(checkpoints)
-        // //custom progress tracking for accounts
-        // val logProgress = logProcessingProgress(entityName = "account", totalToProcess = total, processStartNanos = System.nanoTime()) _
 
         /* Process in a strictly sequential way, to avoid opening an unbounded number of requests on the http-client.
         * The size and partition of results is driven by the NodeOperator itself, were each page contains a
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -16,7 +16,6 @@ import scala.annotation.tailrec
 import scala.concurrent.{Await, Future}
 import scala.concurrent.duration.Duration
 import scala.util.{Failure, Success, Try}
-import cats.kernel.Monoid
 
 /**
   * Entry point for synchronizing data between the Tezos blockchain and the Conseil database.
@@ -93,14 +92,14 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
      * Otherwise, any error will make Lorre proceed as expected by default (stop or wait for next cycle)
      */
     val attemptedProcessing =
-      if (ignoreProcessFailures.forall(ignore => ignore == "false" || ignore == "no"))
-        processing
-      else
+      if (ignoreProcessFailures.forall(ignore => ignore == "true" || ignore == "yes"))
         processing.recover {
           //swallow the error and proceed with the default behaviour
-          case f@(AccountsProcessingFailed(_, _) | BlocksProcessingFailed(_, _)) =>
+          case f@(AccountsProcessingFailed(_, _) | BlocksProcessingFailed(_, _) | DelegatesProcessingFailed(_, _)) =>
             logger.error("Failed processing but will keep on going next cycle", f)
         }
+      else
+        processing
 
     Await.result(attemptedProcessing, atMost = Duration.Inf)
 
@@ -187,6 +186,10 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
         }
       })
     } transform {
+      case Failure(accountFailure @ AccountsProcessingFailed(_, _)) =>
+        Failure(accountFailure)
+      case Failure(delegateFailure @ DelegatesProcessingFailed(_, _)) =>
+        Failure(delegateFailure)
       case Failure(e) =>
         val error = "Could not fetch blocks from client"
         logger.error(error, e)
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -92,7 +92,7 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
      * Otherwise, any error will make Lorre proceed as expected by default (stop or wait for next cycle)
      */
     val attemptedProcessing =
-      if (ignoreProcessFailures.forall(ignore => ignore == "true" || ignore == "yes"))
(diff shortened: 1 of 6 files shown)


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -262,28 +273,27 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
       }
 
     val saveAccounts = db.run(TezosDb.getLatestAccountsFromCheckpoint) map {
       checkpoints =>
         logger.debug("I loaded all stored account references and will proceed to fetch updated information from the chain")
         val (pages, total) = tezosNodeOperator.getAccountsForBlocks(checkpoints)
-        //custom progress tracking for accounts
-        val logProgress = logProcessingProgress(entityName = "account", totalToProcess = total, processStartNanos = System.nanoTime()) _
+        // //custom progress tracking for accounts
+        // val logProgress = logProcessingProgress(entityName = "account", totalToProcess = total, processStartNanos = System.nanoTime()) _
```
piotrkosecki: why is this commented?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -158,6 +183,11 @@ class TezosNodeOperator(val node: TezosRPCInterface, val network: String, batchC
   def getAccountsForBlocks(accountsBlocksIndex: Map[AccountId, BlockReference]): PaginatedAccountResults = {
     import TezosTypes.Syntax._
 
+    val reverseIndex =
+      accountsBlocksIndex.groupBy{case (id, blockRef) => blockRef._1}
```
piotrkosecki: can you pattern match also the `blockRef`? `_.1` is not very descriptive

piotrkosecki:
left some minor comments, otherwise LGTM

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -266,6 +299,11 @@ class TezosNodeOperator(val node: TezosRPCInterface, val network: String, batchC
   def getDelegatesForBlocks(keysBlocksIndex: Map[PublicKeyHash, BlockReference]): PaginatedDelegateResults = {
     import TezosTypes.Syntax._
 
+    val reverseIndex =
+      keysBlocksIndex.groupBy { case (pkh, blockRef) => blockRef._1 }
+       .mapValues(_.keySet)
```
dorians: This method is deprecated in 2.13. We should avoid it.

dorians:
I would get rid of phrases like `_._1` but in general looks good.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: Good one, I forgot to remove it

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
ivanopagano: There are 7 places where this is used in the codebase.
Should we do it right now, or maybe define a custom issue with his own very simple PR?
Should we wait for 2.13 to become relevant to Conseil?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -268,10 +268,12 @@ object TezosDatabaseOperations extends LazyLogging {
       copyAccountsToDelegateContracts(
         delegates.flatMap(_.content.values.flatMap(_.delegated_contracts)).toSet
       )
-    for {
+
+    (for {
       updated <- delegatesUpdateAction.map(_.sum)
       _ <- contractsUpdateAction
-    } yield updated
+    } yield updated).transactionally
```
ba205: What's the reasoning for this being transactional now, as opposed to before?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -76,6 +76,32 @@ class TezosNodeOperator(val node: TezosRPCInterface, val network: String, batchC
   type BallotBlock = (Block, List[Voting.Ballot])
   type BakerBlock = (Block, List[Voting.BakerRolls])
 
+  /**
+    * Generic fetch for paginated data relative to a specific block.
```
ba205: In this case, the Key will usually be BlockHash, and the Entity would be any data linked to that hash, like that block's accounts/delegates, correct? Why are we generalizing over Key?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -76,6 +76,32 @@ class TezosNodeOperator(val node: TezosRPCInterface, val network: String, batchC
...
+    * Will return back the entities indexed by their unique key
+    * from those passed as inputs, through the use of the
+    * `entityLoad` function, customized per each entity type.
+    * e.g. we use it to load accounts and delegates without
+    * duplicating any logic.
+    */
+  def getPaginatedEntitiesForBlock[Key, Entity](
+    entityLoad: (List[Key], BlockHash) => Future[Map[Key, Entity]]
```
ba205: Rename to 'getEntityForBlock', since that's the shape of the name used for this function in general.

ivanopagano: I'm not sure I follow, this method concern is specifically to load lazily evaluated "pages" of some type, that's the added behaviour over the passed-in loader `entityLoad`.
And that the reason for the unusual return type: `Iterator[Future[...]]` which loads async pages of results on demand

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: It was a bug, I effectively observed locally that the contracts's reference copy on db would fail because the delegate it was referencing (as a FK) wasn't yet transactionally visible to the second operation.

Issue #225, opened 2019-01-09
Separate metadata definition from general app config


anonymoussprocket:

anonymoussprocket:
Splitting off metadata overrides also makes sense. They should be listed as an array of files to include. I would expect to see something like tezos.mainnet.metadata, tezos.alphanet.metadata, etc.

PR #410, opened 2019-06-11
225 - separation of metadata config

dorians:
refers to #225 

Extract configuration of metadata files into separate config files.


dorians:
> Could you also add instructions on how to override these files from the command line?

It's the same as it used to be. The structure of the config is exactly the same, only the localization in files has changed. You can override it as is described in a README using:
```
-J-Dconfig.resource=your-conf-dev.conf
```

PR #413, opened 2019-06-12
412 - removed `IS TRUE` from query when `inverse = false`

### LINKED ISSUE #412: Remove unnecessary IS TRUE from query
Query currently adds `IS TRUE`  to every predicate with `inverse = false`, but it significantly slows down the query. Needs to be removed.

piotrkosecki:
fixes #412 

When `inverse` is set to `false`, `IS TRUE` won't be added. This will improve the query performance.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
@@ -194,7 +194,11 @@ object DatabaseUtil {
         case OperationType.endsWith => sql"LIKE '%#${vals.head}'"
         case OperationType.isnull => sql"ISNULL"
       }
-      concatenateSqlActions(op, sql" IS #${!inverse}")
+      if(inverse) {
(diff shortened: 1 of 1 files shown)


ivanopagano:
Isn't there a way to instead wrap the clause with a `NOT ($clause)` only for `inverse = true`?

piotrkosecki:
@dorians we have test cases for that, functionality does not change, only performance

piotrkosecki:
@ivanopagano from what I've tested `NOT` does not work in all cases, so the code would be less generic

ivanopagano:
> @ivanopagano from what I've tested `NOT` does not work in all cases, so the code would be less generic

Do we have some test to verify the generated sql? (if we don't, I guess we should, to make our lives simpler...)

piotrkosecki:
@ivanopagano we will have a possibility of creating those tests when my PR #414 will be merged - we could compare SQL outputs. At the moment I don't think we have any way to check generated queries.

PR #417, opened 2019-06-13
416 - More metadata: displayPriority, displayOrder

### LINKED ISSUE #416: More metadata: displayPriority, displayOrder
We need to add two more metadata properties at attribute level: `displayPriority` and `displayOrder`.

Priority will specify relative weight, like title vs subtitle. It would be 0, 1, 2 (high, medium, low). The idea is that we may drop items marked as "low" depending on display context.

Order is a numeric value which the UI can use to sort the attributes for display.

dorians:
closes #416 

We need to add two more metadata properties at attribute level: displayPriority and displayOrder.

Priority will specify relative weight, like title vs subtitle. It would be 0, 1, 2 (high, medium, low). The idea is that we may drop items marked as "low" depending on display context.

Order is a numeric value which the UI can use to sort the attributes for display.

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -163,6 +163,8 @@ akka {
 # * data-type: String, since this value is later mapped to enum, values are limited to those supported by method
 #                      tech.cryptonomic.conseil.tezos.TezosPlatformDiscoveryOperations.mapType
 # * scale: Int, indicate to the UI how to display a value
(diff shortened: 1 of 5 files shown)


ba205:
@dorians shouldn't the new fields be used in PlatformDiscoveryTest, too? AttributeConfiguration is used in that file, too. Also, is AttributeConfiguration used for anything besides tests?

PR #419, opened 2019-06-14
418 - running pre-caching for the attributes cardinality during startup

### LINKED ISSUE #418: Run pre-caching for the attributes cardinality during startup
Currently it's being triggered via endpoint.

piotrkosecki:
resolves #418

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Conseil.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Conseil.scala
@@ -53,6 +53,11 @@ object Conseil extends App with LazyLogging with EnableCORSDirectives with Conse
         case Success(_) => logger.info("Pre-caching successful!")
       }
 
+      tezosPlatformDiscoveryOperations.initAttributesCache.onComplete {
+        case Failure(exception) => logger.error("Pre-caching attributes failed", exception)
+        case Success(_) => logger.info("Pre-caching attributes successful!")
+      }
+
       lazy val metadataService = new MetadataService(platforms, transformation, cacheOverrides, tezosPlatformDiscoveryOperations)
       lazy val platformDiscovery = PlatformDiscovery(metadataService)(tezosDispatcher)
       lazy val data = Data(platforms, tezosPlatformDiscoveryOperations, server)(tezosDispatcher)
--- a/src/main/scala/tech/cryptonomic/conseil/Conseil.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Conseil.scala
@@ -46,7 +46,7 @@ object Conseil extends App with LazyLogging with EnableCORSDirectives with Conse
       lazy val cacheOverrides = new AttributeValuesCacheConfiguration(metadataOverrides)
 
       lazy val tezosPlatformDiscoveryOperations =
-        TezosPlatformDiscoveryOperations(ApiOperations, metadataCaching, cacheOverrides, server.cacheTTL)(executionContext)
+        TezosPlatformDiscoveryOperations(ApiOperations, metadataCaching, cacheOverrides, server.cacheTTL)(executionContext, contextShift)
 
       tezosPlatformDiscoveryOperations.init().onComplete {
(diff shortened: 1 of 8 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -295,6 +297,7 @@ class TezosPlatformDiscoveryOperations(
             } else {
               (for {
                 _ <- caching.putAttributes(tableName, attributes)
+                _ <- IO.contextShift(executionContext).shift
```
ba205: Should switching to IO be put in this pr, or did it directly help with the caching task?

piotrkosecki: that was a change requested by @ivanopagano in previous caching PR, but it was merged before I could push it, so I thought I'll add it here

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -177,6 +177,7 @@ class TezosPlatformDiscoveryOperations(
         } else {
           (for {
             _ <- caching.putEntities(networkName, ent)
+            _ <- IO.contextShift(executionContext).shift
```
ivanopagano: If it helps, you can reduce repetition by making an implicit ContextShift[IO] available at the class beginning, and simply call `IO.shift` henceforth

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -360,40 +363,29 @@ class TezosPlatformDiscoveryOperations(
...
+     }
+
+    result.flatMap(_.toList.sequence) >> caching.updateCachingStatus(Finished)
+  }.unsafeToFuture()
 
   /** Helper method for updating */
   private def getAllUpdatedAttributes(entities: List[Entity], attributes: Cache[List[Attribute]]): IO[Map[String, List[Attribute]]] = {
     val queries = attributes
-      .filterKeys(entities.map(_.name).toSet)
+      .filterKeys {
```
ivanopagano: this was quite subtle

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -53,6 +53,11 @@ object Conseil extends App with LazyLogging with EnableCORSDirectives with Conse
         case Success(_) => logger.info("Pre-caching successful!")
       }
 
+      tezosPlatformDiscoveryOperations.initAttributesCache.onComplete {
```
ivanopagano: Did we consider doing retries? Maybe in a separate PR?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
piotrkosecki: yes 😄

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
piotrkosecki: I think it should be don in other PR for this and for preCaching

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
piotrkosecki: ok, done

dorians:
Looks good, my only concern is what happens when we run `Lorre`? Should we rerun `Conseil` after that to have cardinality up to date?

PR #425, opened 2019-06-18
Fixes for Lorre sync issues

vishakh:
This PR makes two changes to make Lorre more compliant with the recent behaviour of Tezos nodes:

- Lorre makes an initial connection check to Tezos by querying for the block head. While querying for the head, it also tried to fetch associated data related to governance, etc. However, since governance only kicked in, for example, on block ~210,000 on the Tezos `mainnet` the initial connection check keeps failing until the block data complies with the expected format. To get around this, new functions have been added to query only the block head and not any associated data.

- The newest version of the Tezos node returns error messages for certain `delegates` queries (see example below) and the corresponding data is not populated into our tables. Due to this, a foreign key constraint is violated when `delegated_contracts` are written (see screenshot below). Until we learn more about this Tezos issue and its resolution, the specific foreign key constraint has been removed.

Example of failed delegates query:

```json
[
  {
    "kind": "temporary",
    "id": "proto.001-PtCJ7pwo.tez.subtraction_underflow",
    "amounts": [
      "0",
      "13763256500"
    ]
  }
]
```

Example of foreign key issue:

image

### DIFF
--- a/doc/conseil.sql
+++ b/doc/conseil.sql
@@ -424,8 +424,9 @@ CREATE INDEX ix_proposals_protocol ON public.proposals USING btree (protocol_has
 ALTER TABLE ONLY public.accounts
     ADD CONSTRAINT accounts_block_id_fkey FOREIGN KEY (block_id) REFERENCES public.blocks(hash);
 
-ALTER TABLE ONLY public.delegated_contracts
-    ADD CONSTRAINT contracts_delegate_pkh_fkey FOREIGN KEY (delegate_value) REFERENCES public.delegates(pkh);
+-- Disabling this constraint due to Tezos node issues preventing some delegates from being fetched.
+--ALTER TABLE ONLY public.delegated_contracts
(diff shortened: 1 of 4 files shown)


ivanopagano:
As a side-note related to failures downloading referenced delegates, I was already pondering about parsing expected errors from tezos (i.e. those complying with the schema) and apply some policy (like retries, or pending status?) to those queries before giving up the request completely

Thoughts? @vishakh @anonymoussprocket @bhaskarkishore @ba205 ?

vishakh:
> As a side-note related to failures downloading referenced delegates, I was already pondering about parsing expected errors from tezos (i.e. those complying with the schema) and apply some policy (like retries, or pending status?) to those queries before giving up the request completely
> 
> Thoughts? @vishakh @anonymoussprocket @bhaskarkishore @ba205 ?

I think the idea is sound but we first need to get a clear indication which failures are expected and which ones are not.

vishakh:
> The code looks good.
> 
> Personal taste: I'd use a different name like
> `getBareBlock`, `getRawBlock`
> 
> I can't say why, but `getBlockOnly` doesn't convey to me the expected meaning. As I said, simply personal feelings.

Done

PR #428, opened 2019-06-19
427 - Introducing fix for query aggregation when single field is used in query body

### LINKED ISSUE #427: Query fails when single field is used and it's being aggregated
Example body for query in accounts:
```
{
    "fields": ["account_id"],
    "predicates": [
    	{ "field": "account_id", "set": ["tz1"], "operation": "startsWith", "inverse": false },
    	{ "field": "balance", "set": [1000], "operation": "lt", "inverse": false },
    	{ "field": "delegate_value", "set": [], "operation": "isnull", "inverse": false }
    	],
    "orderBy": [],
    "aggregation": [{"field": "account_id", "function": "count"}],
    "limit": 1
}
```

piotrkosecki:
resolves #427

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
@@ -103,11 +103,12 @@ object DatabaseUtil {
         * @return new SQLActionBuilder containing limit statement
         */
       def addGroupBy(aggregation: List[Aggregation], columns: List[String]): SQLActionBuilder = {
-        if(aggregation.isEmpty) {
+        val aggregationFields = aggregation.map(_.field).toSet
+        val existingAggregationFields = columns.toSet.diff(aggregationFields).toList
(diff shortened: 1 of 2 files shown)


src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
```diff
@@ -103,11 +103,12 @@ object DatabaseUtil {
         * @return new SQLActionBuilder containing limit statement
         */
       def addGroupBy(aggregation: List[Aggregation], columns: List[String]): SQLActionBuilder = {
-        if(aggregation.isEmpty) {
+        val aggregationFields = aggregation.map(_.field).toSet
+        val existingAggregationFields = columns.toSet.diff(aggregationFields).toList
```
ivanopagano: I'm not sure if the name express what is in this set

piotrkosecki: renamed

PR #433, opened 2019-06-26
More verbose logging for Lorre

vishakh:
It is currently difficult to operate Lorre as it is hard to follow what it is specifically doing at any point. Especially when there is a failure, no diagnosis can be made unless an informed developer runs Lorre locally and tries to recreate the failure conditions with full debug logging.

As Lorre follows a pretty well defined pipeline of actions, it would be immensely valuable to verbosely print information about all interactions it is undertaking with the database and the blockchain node. This way, it becomes much easier to diagnose issues and choke points without even having deep knowledge of the system. We should, in fact, err in favour of verbose logging.

This PR adds several logging statements to make it clearer when Lorre is interacting with the databases or starting a new step of its pipeline. Some debug statements have been changed into info statements. 

This is only a beginning as much more granular logging can be added.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
@@ -69,7 +69,8 @@ trait BlocksDataFetchers {
     private def makeUrl = (offset: Offset) => s"blocks/${hashRef.value}~${String.valueOf(offset)}"
 
     //fetch a future stream of values
-    override val fetchData =
+    override val fetchData = {
+      logger.info("Fetching blocks")
       Kleisli(
         offsets =>
           node.runBatchedGetQuery(network, offsets, makeUrl, fetchConcurrency).onError {
@@ -85,6 +86,7 @@ trait BlocksDataFetchers {
                 .pure[Future]
           }
       )
+    }
 
     // decode with `JsonDecoders`
     override val decodeData = Kleisli { json =>
@@ -115,7 +117,8 @@ trait BlocksDataFetchers {
 
     private val makeUrl = (hash: BlockHash) => s"blocks/${hash.value}/operations"
 
-    override val fetchData =
+    override val fetchData = {
+      logger.info("Fetching operations")
       Kleisli(
         hashes =>
           node.runBatchedGetQuery(network, hashes, makeUrl, fetchConcurrency).onError {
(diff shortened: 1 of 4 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
```diff
@@ -69,7 +69,8 @@ trait BlocksDataFetchers {
     private def makeUrl = (offset: Offset) => s"blocks/${hashRef.value}~${String.valueOf(offset)}"
 
     //fetch a future stream of values
-    override val fetchData =
+    override val fetchData = {
+      logger.info("Fetching blocks")
```
ivanopagano: The logging here should be in the Kleisli constructor's internal function.
i.e. after the `offsets =>` are passed.
Calling `fetchData`is simply returning the kleisli wrapper around the real function. The kleisli returned is yet to be "run" at this point.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
```diff
@@ -69,7 +69,8 @@ trait BlocksDataFetchers {
...
       Kleisli(
         offsets =>
           node.runBatchedGetQuery(network, offsets, makeUrl, fetchConcurrency).onError {
```
ivanopagano: Here we want
```scala
logger.info("Fetching blocks for offsets ${offsets.min} to ${offsets.max}")
```
Or similar (I didn't check the syntax)

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
```diff
@@ -115,7 +117,8 @@ trait BlocksDataFetchers {
 
     private val makeUrl = (hash: BlockHash) => s"blocks/${hash.value}/operations"
 
-    override val fetchData =
+    override val fetchData = {
+      logger.info("Fetching operations")
```
ivanopagano: same considerations as for the blocks fetcher logging with kleislis

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
vishakh: Would you mind giving me the whole snippet for `fetchData()`? I am not familiar enough with the syntax to confidently make the change.

PR #439, opened 2019-06-27
Refactor: 434 fix checkpoint design to scale

### LINKED ISSUE #434: The checkpoints tables do not scale
As the number of accounts grows, it's already showing up in staging and local (resource-limited) contexts that the operation to fetch all data in the checkpoint can easily time-out, stopping the whole process.
The next run can only add to the problem, as new entries are added to the existing ones for the new cycle.

We need to rethink how accounts are fetched, or how the checkpoint is used, to make it scale with the incoming data load.

ivanopagano:
Fixes #434 
In order to reduce the impact of the checkpoint's growth when reading the data to update, this PR simply loads the whole table and filters duplicate keys algorithmically.

The impact on code execution is remarkably low as measurements shows a max time in the order of tens of milliseconds in the worst case (for tens of thousands of rows to filter out).
On the other hand the query to load the data is now simplified to the bone, which should reduce the risk of timeouts. Sorting should be optimized via indexing of the block_level column.

Additional improvements can be made, by doing a recursive filtering via lazy collection over the sorted rows, yet I can't see any real added value at this point.

The issue is not solved completely, as still I couldn't figure out why the checkpoint is not properly emptied after each cycle (unless concurrency is at work here), which seems logical looking at the code. As the table accumulates cruft we lose performance over the cleanup.

The redesign PR #375 could introduce a totally different way to process accounts and delegates in-place as the stream of blocks is being processed, removing the need to have a checkpoint table at all.

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -70,13 +70,13 @@ batchedFetches {
 
   #Used to specify the max-time allowed for each block-page to finish processing
   #Currently takes into account the time to process all corresponding accounts for each page
-  blockPageProcessingTimeout = 15 minutes
+  blockPageProcessingTimeout = 1 hour
 
   #Used to specify the max-time allowed for each accounts-page to finish processing
-  accountPageProcessingTimeout = 5 minutes
+  accountPageProcessingTimeout = 15 minutes
 
   #Used to specify the max-time allowed for each delegate-page to finish processing
-  delegatePageProcessingTimeout = 5 minutes
+  delegatePageProcessingTimeout = 15 minutes
 }
 
 # Settings for blockchains Conseil can connect to. Nested by blockchain name and network.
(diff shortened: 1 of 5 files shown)


src/main/resources/reference.conf
```diff
@@ -70,13 +70,13 @@ batchedFetches {
 
   #Used to specify the max-time allowed for each block-page to finish processing
   #Currently takes into account the time to process all corresponding accounts for each page
-  blockPageProcessingTimeout = 15 minutes
+  blockPageProcessingTimeout = 1 hour
```
ba205: Isn't this a little dangerous? Are we assured consistently logging is going to happen during this hour?

vishakh: Turns out we do need this after all. Sometimes the checkpoint table can get really clogged up. Took 48 minutes to break the logjam on my local machine.

PR #444, opened 2019-07-01
434: Use the in-memory values for account ids to fetch and store accounts

### LINKED ISSUE #434: The checkpoints tables do not scale

ivanopagano:
fixes #434 

The current proposal uses akka-streams to keep the tcp connection pool under control.

It does so both at the blocks processing level, and mostly at the accounts and delegates processing level, in the latter case by counting the number of entities to fetch before doing it.

Additionally, to try and speed up the whole process, the account ids to load are now taken from memory, using the results of blocks just loaded.
The checkpoint table is only used to guarantee that those accounts are not lost even in the case of failure or sudden stop of the application.
The checkpoint is cleaned after the in-memory ids have finished processing.

I only could locally test the proposal, but I didn't experience issues with the pool before, so I can't guarantee yet that this solution is finally solving the problems experienced in staging

Break-down:

 * Instead of reading checkpoints for accounts every cycle from db, use
   the already available in-memory values from blocks processing
 * Use the checkpoint table only as a fallback if the accounts fetching
   should fail, or the system is stopped, to retry the fetching at the
   first chance
 * In case the in-memory processing succeeds, clean the checkpoint
 * Remove the blocking solution to process the data pages, and make use
   of akka-streams to keep the pool consumption of outgoing client
   requests at bay
 * refactor the checkpoint loading to make the sorted filtering as an
   inner function

### Changes to DB
none

### Changes to API
none

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -2,6 +2,8 @@ package tech.cryptonomic.conseil
 
 import akka.actor.ActorSystem
 import akka.Done
+import akka.stream.scaladsl.Source
+import akka.stream.ActorMaterializer
 import mouse.any._
 import com.typesafe.scalalogging.LazyLogging
 import tech.cryptonomic.conseil.tezos.{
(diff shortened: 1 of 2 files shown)


ivanopagano:
We should evaluate on the staging env if this change improves latency, so that we can resume the original timeouts in the configuration

PR #451, opened 2019-07-04
Add build files for DockerHub

vishakh:
Pulling in with much gratitude changes made by @tulpenhaendler to make Conseil more friendly for Docker deployments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,60 @@ Query API for the Tezos blockchain
 
 ## Running Conseil
 
+
+### Docker
+
+To build this project git clone it and run:
+
+```
+docker build -t conseil .
+```
+
+The Container can be configured using the following environment variables:
+
+Database Config:
+
+
+* DB_Host - default: db
+* DB_User - default: user
+* DB_Password - default: password
+* DB_Database: default: conseil
+
+
+
+Tezos Node Config:
+
+
+* XTZ_Scheme - http or https, default : http
+* XTZ_Host - default: node
+* XTZ_Prefix - prefix for api calls, default: ""
+* XTZ_Port - default 8732
+* XTZ_Network - default mainnet
+
+
+Conseil Config:
+
+
+* API_PORT - Conseil API port, default: 80
+
+Or, you can use your own config file, in which case specify the environment variable `CONFIG` with the path to your file
+
+To run Lorre ( the indexer ) run:
+
+```
+docker run conseil lorre
+```
+
+and Conseil ( the API ):
+
+```
+docker run conseil conseil
+```
+
+
+## Running Conseil - classic
+
+
 Conseil has two entry points:
 - `src/main/tech/cryptonomic/Conseil/Conseil.scala` runs a server exposing a RESTful API over blockchain data, stored in its own database.
 - `src/main/tech/cryptonomic/Conseil/Lorre.scala` is a background process that loops and synchronizes (Tezos) blockchain data to database.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -6,6 +6,60 @@ Query API for the Tezos blockchain
...
+* XTZ_Host - default: node
+* XTZ_Prefix - prefix for api calls, default: ""
+* XTZ_Port - default 8732
+* XTZ_Network - default mainnet
+
+
+Conseil Config:
+
+
+* API_PORT - Conseil API port, default: 80
```
leenpaws: Cloudflare uses 443, should this be changed? or is port 80 replacing port 1337 on the container, would require a change to nginx.

README.md
```diff
@@ -6,6 +6,60 @@ Query API for the Tezos blockchain
...
+```
+
+The Container can be configured using the following environment variables:
+
+Database Config:
+
+
+* DB_Host - default: db
+* DB_User - default: user
+* DB_Password - default: password
```
leenpaws: Presuming the default credentials are going to be replaced by the credentials in the conseil.conf file?

README.md
```diff
@@ -6,6 +6,60 @@ Query API for the Tezos blockchain
...
+
+
+Tezos Node Config:
+
+
+* XTZ_Scheme - http or https, default : http
+* XTZ_Host - default: node
+* XTZ_Prefix - prefix for api calls, default: ""
+* XTZ_Port - default 8732
+* XTZ_Network - default mainnet
```
leenpaws: Currently only dev nodes are using containers, shouldn't the defaults be tailored towards dev environments which use alphanet?

README.md
vishakh: See the entrypoint script below. It replaces these with env vars.

README.md
vishakh: Most people using Conseil will use `mainnet`.

README.md
vishakh: No, this is for a generic deployment.

docker/template.conf
```diff
@@ -0,0 +1,105 @@
+include "metadata.conf"
+
+conseil: {
+  hostname: "0.0.0.0"
+  port: {{APIPORT}}
```
leenpaws: is this 80 or 1337?

vishakh: The default value as set above in `entrypoint.sh` is 80. If you set the env var `API_PORT` to some other number it will use that instead.

ivanopagano: Same goes for all the other env-vars usage

docker/template.conf
```diff
@@ -0,0 +1,105 @@
...
+
+conseildb = {
+  dataSourceClass = "org.postgresql.ds.PGSimpleDataSource"
+  properties = {
+    databaseName = "{{DBDB}}"
+    user = "{{DBUSER}}"
+    password = "{{DBPW}}"
+    url = "jdbc:postgresql://{{DBHOST}}:{{DBPORT}}/{{DBDB}}"
+  }
+  numThreads = 64
```
ivanopagano: Maybe on a later PR:
these values are custom-tailored based on specific environmental resources.
The number is quite high, and there's an heuristic to compute it based on cores and disk access speed.

It might be worthwhile to inject these via envvars too (even using a single var, as they should match anyway), enabling custom tailoring based on measured usage of db resources and jvm threads

ivanopagano: Or we could simply leave this to the use of the custom configuration file override...

PR #459, opened 2019-07-05
Bugfix/457 sequence checkpoint cleanup

### LINKED ISSUE #457: Cleanup of the checkpoint tables is not correctly serialized
The operation to clean the stored accounts/delegates after saving the whole entity in the db, was possibly executed asynchronously, use monadic operations like flatMap/flatTap and similar to ensure sequential semantic.

### LINKED ISSUE #458: Use in-memory delegate reference to save them
Apply the same operational semantic as for accounts, to use the loaded delegate hashes immediately for data fetching and persistence.
Use the checkpoint table only as  a backup for unexpected failure of the system, simply to avoid data-loss

ivanopagano:
fix #457 
closes #458 
Increase performance of delegates storage using in-memory data, if possible.
Correct potential issues with incorrect interleaving of database operations.


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -300,183 +305,227 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+    )(implicit mat: ActorMaterializer): Future[Done] = {
+      logger.info("Processing latest Tezos data for account delegates...")
 
-    def logWriteFailure: PartialFunction[Try[_], Unit] = {
-      case Failure(e) =>
-        logger.error("Could not write accounts to the database")
+      def keepMostRecent(associations: List[(PublicKeyHash, BlockReference)]): Map[PublicKeyHash, BlockReference] =
+      associations.foldLeft(Map.empty[PublicKeyHash, BlockReference]) { (collected, entry) =>
+        val key = entry._1
+        if (collected.contains(key)) collected else collected + (key -> entry._2)
```
dorians: why not `(collected, (key, value))`? instead of `(collected, entry)`?

PR #461, opened 2019-07-08
450 - metadata: int

### LINKED ISSUE #450: Metadata: int
Allow overriding of attribute data type to `Int`. This would be applicable to things like block level, account counter, block priority, block protocol index, block cycle, block period index, period position, etc.

dorians:
closes #450 

Allows using "int" as data type in metadata overrides.


ivanopagano:
Please could you point the PR to the appropriate issue number?

PR #462, opened 2019-07-08
442 - Metadata update for boolean values

### LINKED ISSUE #442: Metadata update for boolean values
Please generate 'true' and 'false' for distinct values of attributes labeled as boolean instead of the current 'T' and 'F'.

dorians:
closes #442 

Boolean values are now represented as `true` and `false` instead of `t` and `f`.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
@@ -471,7 +471,7 @@ object TezosDatabaseOperations extends LazyLogging {
     * @return       distinct elements in given column as a list
     */
   def selectDistinct(table: String, column: String)(implicit ec: ExecutionContext): DBIO[List[String]] =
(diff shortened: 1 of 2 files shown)


src/main/resources/metadata.conf
```diff
@@ -15,6 +15,8 @@
 # * display-priority: relative weight, like title vs subtitle. Allowed values: [0, 1, 2]
 # * display-order: numeric value which the UI can use to sort the attributes for display
 # * sufficient-for-query: Boolean, which is for checking if predicate on this field is sufficient to make a query
+# * currency-symbol: used in context of account balances, example: "ꜩ"
+# * currency-symbol-code: used in context of account balances, example: 42793
```
ivanopagano: Is it possible to add an explanation of what the code should represent, I have no clue, is it a unicode code-point?

dorians:
> Isn't there any part of the codebase that should handle those datatypes to make computations or querying of any sort?

Those parameters are because we need them on frontend and they're related with metadata overrides. The only logic is that we can override data from the database on the fly with that parameters.

src/main/resources/metadata.conf
dorians: @anonymoussprocket, could you comment on that? I only know that it's a value which is needed for front-end.

PR #463, opened 2019-07-08
460 - amend proposals decoding to correctly store proposals in the database

### LINKED ISSUE #460: voting protocols ids are not stored
We currently have signs of missing protocols data from the database.

Lorre is not correctly storing such data, as the parsing fails.

To reproduce just run
`sbt "runLorre -v -d 10000 -h BLDKDUoTtLhPhdHFqhGWWPJB1RLYUzBwHXwcQ3DUp6jWsxJNYvi prodnet"`

where your local `prodnet` config must refer to a production node

The error in parsing logs this output

```
I fetched voting proposal protocols json from tezos node that I'm unable to decode: [["PtdRxBHvc91c2ea2evV6wkoqnzW7TadTg9aqS9jAn2GbcPGtumD",63]]
circe.DecodingFailure$$anon$2: String: DownArray
```

ivanopagano:
fixes #460 

Read protocols data as actually tuples of the proposal hash with associated number of supporters.
Update the codec definitions to read correctly the values and store the additional info into the db.

## Changes to db-schema
We need to introduce a new column to the `proposals` table

```sql
ALTER TABLE public.proposals ADD COLUMN supporters integer;
```

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
@@ -52,6 +52,9 @@ object TezosTypes {
   /** convenience alias to simplify declarations of block hash+level tuples */
   type BlockReference = (BlockHash, Int)
 
+  /** use to remove ambiguities about the meaning in voting proposals usage */
+  type ProposalSupporters = Int
+
   final case class PublicKey(value: String) extends AnyVal
 
   final case class PublicKeyHash(value: String) extends AnyVal
@@ -452,7 +455,7 @@ object TezosTypes {
   object Voting {
 
     final case class Vote(value: String) extends AnyVal
-    final case class Proposal(protocols: List[ProtocolId], block: Block)
+    final case class Proposal(protocols: List[(ProtocolId, ProposalSupporters)], block: Block)
     final case class BakerRolls(pkh: PublicKeyHash, rolls: Int)
     final case class Ballot(pkh: PublicKeyHash, ballot: Vote)
   }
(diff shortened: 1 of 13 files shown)


src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
```diff
@@ -18,7 +18,7 @@ import scala.language.postfixOps
 import scala.util.Random
 
 class TezosDatabaseOperationsTest
-  extends WordSpec
+    extends WordSpec
```
piotrkosecki: I think we have different settings here...

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -452,7 +455,7 @@ object TezosTypes {
   object Voting {
 
     final case class Vote(value: String) extends AnyVal
-    final case class Proposal(protocols: List[ProtocolId], block: Block)
+    final case class Proposal(protocols: List[(ProtocolId, ProposalSupporters)], block: Block)
```
piotrkosecki: maybe make a type alias for the pair?

ivanopagano: I'm not sure, it would spell something like `ProposedProtocol`, `ProtocolProposal`?

I see room for confusion because it would be used to define a `Proposal`... and then what would be the proper attribute name for those instead of `protocols`?

How easy would be to mix the two things when trying to read and reason about client code...?

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
ivanopagano: I'm quite astonished by how this class was totally reformatted, I simply changed a single attribute in the `proposals` table...

Yet the formatting can't be different, the definition's in the git repo's `.scalafmt.conf` file.
Unless scalafmt is bugged itself (I heard that it might be not consistent between runs in some case, yet I tried to re-run it and it consistently gave me this format)

PR #474, opened 2019-07-12
238 - Make conversion for db using chimney lib, when convenient

### LINKED ISSUE #238: Consider Chimney to reduce boilerplate in DatabaseConversions
I think this library could help us avoid some of boilerplate code in this object: https://github.com/scalalandio/chimney

What do you think?

_Originally posted by @piotrkosecki in https://github.com/Cryptonomic/Conseil/pull/232_

ivanopagano:
Closes #238 

Make use of the (already imported) library chimney for value object transformation, simplifying a bit the syntax.

In some cases the macro expansion seems to diverge (e.g. for any of the operation's subtypes).
For these and for cases where the added value was negligible, the old conversion code was kept.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
@@ -4,9 +4,12 @@ import tech.cryptonomic.conseil.tezos.TezosTypes._
 import tech.cryptonomic.conseil.tezos.FeeOperations._
 import tech.cryptonomic.conseil.util.Conversion
 import cats.{Id, Show}
+import cats.syntax.option._
+import mouse.any._
 import java.sql.Timestamp
 import monocle.Getter
 import io.scalaland.chimney.dsl._
+import io.scalaland.chimney.Transformer
 
 object DatabaseConversions {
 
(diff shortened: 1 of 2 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -33,91 +36,85 @@ object DatabaseConversions {
...
+
+        implicit val rowTransformer: Transformer[(AccountId, Account), Tables.AccountsRow] =
+          (entry: (AccountId, Account)) => {
+            val (id, acc) = entry
+            acc
+              .into[Tables.AccountsRow]
+              .withFieldConst(_.accountId, id.id)
+              .withFieldConst(_.blockId, hash.value)
+              .withFieldConst(_.blockLevel, BigDecimal(level))
+              .withFieldComputed(_.delegateSetable, _.delegate.setable)
```
piotrkosecki: why `withFieldComputed`?

ivanopagano: It's reading inside an inner component, I don't believe chimney's able to figure that out...
In fact this exercise led me to think that multiple input patching could be really a huge addition to the library.
That is, you give me a, b, c and tell me how to use 'em all to build a d

ba205:
As of now, we've added lines into the codebase instead of subtracting them, and the old code seems more readable to me and to the average developer. I'm not super convinced this should be merged.

ivanopagano:
> As of now, we've added lines into the codebase instead of subtracting them, and the old code seems more readable to me and to the average developer. I'm not super convinced this should be merged.

I totally see the point, therefore my proposal is to clean up most of the changes, where those are not useful to simplify the code, keep the conversion where the gain is evident, keep the additional tests introduced in this PR

vishakh:
> > As of now, we've added lines into the codebase instead of subtracting them, and the old code seems more readable to me and to the average developer. I'm not super convinced this should be merged.
> 
> I totally see the point, therefore my proposal is to clean up most of the changes, where those are not useful to simplify the code, keep the conversion where the gain is evident, keep the additional tests introduced in this PR

This sounds quite reasonable. After further reflection (lol), Chimney does seem pretty useful as long as it's used judiciously.

I would suggest returning to this PR soon.

@ba205

vishakh:
> > > As of now, we've added lines into the codebase instead of subtracting them, and the old code seems more readable to me and to the average developer. I'm not super convinced this should be merged.
> > 
> > 
> > I totally see the point, therefore my proposal is to clean up most of the changes, where those are not useful to simplify the code, keep the conversion where the gain is evident, keep the additional tests introduced in this PR
> 
> This sounds quite reasonable. After further reflection (lol), Chimney does seem pretty useful as long as it's used judiciously.
> 
> I would suggest returning to this PR soon.
> 
> @ba205

Upon further, further reflection no usage of Chimney in this PR is egregious. We should update the PR sometime soon so we can merge it!

vishakh:
Closing for now. Will reopen in the future after big refactors are done.

PR #475, opened 2019-07-12
333 - Switch to the commonly used (year month)-based format for release tags

### LINKED ISSUE #333: Use a more lenient tagging policy for releases
Allow any format as in 
`$msg-release-$seqn`
to be correctly parsed for project versioning.

This way manual release tags can be made with some meaningful content, e.g.
`april-2019-release-4`
while the automated tagging from sbt (eventually called in the CI env) could be
`ci-release-4` or even `2019-04-release-4`

ivanopagano:
Closes #333
Make the automatic task for release (`sbt gitTag`) use year and month to format the tag itself, as we've done previously with manual releases.

This means that to tag from the local code (in sync with `master`) it should be enough to run
```bash
sbt gitTag
git push origin --tags
```

This will produce a tag like `2019-july-release-10` on a properly aligned branch (that is, with no uncommitted code)

The sbt command will output the actually produced tag so you can check if doing it locally, and eventually remove the tag if something wicked happened

### DIFF
--- a/doc/conseil.sql
+++ b/doc/conseil.sql
@@ -156,9 +156,9 @@ CREATE TABLE public.blocks (
     meta_cycle_position integer,
     meta_voting_period integer,
     meta_voting_period_position integer,
-    expected_commitment boolean
+    expected_commitment boolean,
+    priority integer
 );
-
 
 --
 -- TOC entry 188 (class 1259 OID 7467507)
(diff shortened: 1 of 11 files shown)


ivanopagano:
Something went wrong when trying to rebase from a cleaner master, I'm gonna close and reopen a new PR (#476) , sorry for the inconvenience.

PR #479, opened 2019-07-16
456 - capture internal transactions

### LINKED ISSUE #456: Capture internal transactions
Tezos operations can have internal operations, i.e. when a call to a smart contract programatically triggers a call to a second contract. Here is an example:

```json
{
            "protocol": "PsYLVpVvgbLhAhoqAkMFUo6gudkJ9weNXhUYCiLDzcUpFpkk8Wt",
            "chain_id": "NetXdQprcVkpaWU",
            "hash": "opMiTubQRZEt7x12rNnxEkTsaGwJLxSHxfKcyxfN9hpX5kED7Mu",
            "branch": "BMb1bp8isA9pRJNAgQkPEP24uh1Eb53Xnp1TzwG4Zkpodf6sY1c",
            "contents": [
                {
                    "kind": "transaction",
                    "source": "tz1MS1g7tETWfiPXtXx6Jx1XUrYJzzFY4QYN",
                    "fee": "0",
                    "counter": "120917",
                    "gas_limit": "2215",
                    "storage_limit": "0",
                    "amount": "0",
                    "destination": "KT1XYHyoewY5CMDdcYB5BjN7dQbWreV5cWgH",
                    "parameters": {
                        "string": "tz1MS1g7tETWfiPXtXx6Jx1XUrYJzzFY4QYN"
                    },
                    "metadata": {
                        "balance_updates": [],
                        "operation_result": {
                            "status": "applied",
                            "storage": {
                                "prim": "Unit"
                            },
                            "consumed_gas": "2008",
                            "storage_size": "69"
                        },
                        "internal_operation_results": [
                            {
                                "kind": "transaction",
                                "source": "KT1XYHyoewY5CMDdcYB5BjN7dQbWreV5cWgH",
                                "nonce": 0,
                                "amount": "1000",
                                "destination": "tz1MS1g7tETWfiPXtXx6Jx1XUrYJzzFY4QYN",
                                "parameters": {
                                    "prim": "Unit"
                                },
                                "result": {
                                    "status": "applied",
                                    "balance_updates": [
                                        {
                                            "kind": "contract",
                                            "contract": "KT1XYHyoewY5CMDdcYB5BjN7dQbWreV5cWgH",
                                            "change": "-1000"
                                        },
                                        {
                                            "kind": "contract",
                                            "contract": "tz1MS1g7tETWfiPXtXx6Jx1XUrYJzzFY4QYN",
                                            "change": "1000"
                                        }
                                    ],
                                    "consumed_gas": "107"
                                }
                            }
                        ]
                    }
                }
            ],
            "signature": "sigbpeU9JBdJha1ZcedRDDkjFCyuvyMMtwjWPpfi2Poa2cuZmkwc2ziqgjzBx6BiVhydvuP3QN6rkx97CxtsbPHtE6dbpb1d"
        }
```

These internal operations are currently being ignored by Lorre. Instead, they should be picked up and their information should be added to the `operations` table with a flag column called `internal` set to `true`.

piotrkosecki:
Resolves #456 
Added capturing internal transactions and other internal transactions and inserting them to the operations table with `internal` flag.
Schema was described here:
https://tezos-prod.cryptonomic-infra.tech/describe/chains/main/blocks/head/operations

### Required database changes:
In `operations` table:
+     internal boolean NOT NULL

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
@@ -130,7 +130,7 @@ object DatabaseConversions {
   }
 
   private val convertEndorsement: PartialFunction[(Block, OperationHash, Operation), Tables.OperationsRow] = {
-    case (block, groupHash, Endorsement(level, metadata)) =>
+    case (block, groupHash, e @ Endorsement(level, metadata)) =>
       Tables.OperationsRow(
         operationId = 0,
         operationGroupHash = groupHash.value,
@@ -140,7 +140,8 @@ object DatabaseConversions {
         slots = Some(metadata.slots).map(concatenateToString),
         blockHash = block.data.hash.value,
         blockLevel = block.data.header.level,
-        timestamp = toSql(block.data.header.timestamp)
+        timestamp = toSql(block.data.header.timestamp),
+        internal = e.internal
       )
   }
 
@@ -154,12 +155,13 @@ object DatabaseConversions {
         nonce = Some(nonce.value),
         blockHash = block.data.hash.value,
         blockLevel = block.data.header.level,
-        timestamp = toSql(block.data.header.timestamp)
+        timestamp = toSql(block.data.header.timestamp),
+        internal = false
       )
   }
 
   private val convertActivateAccount: PartialFunction[(Block, OperationHash, Operation), Tables.OperationsRow] = {
-    case (block, groupHash, ActivateAccount(pkh, secret, metadata)) =>
+    case (block, groupHash, a @ ActivateAccount(pkh, secret, metadata)) =>
       Tables.OperationsRow(
         operationId = 0,
         operationGroupHash = groupHash.value,
@@ -168,12 +170,13 @@ object DatabaseConversions {
         secret = Some(secret.value),
         blockHash = block.data.hash.value,
         blockLevel = block.data.header.level,
-        timestamp = toSql(block.data.header.timestamp)
+        timestamp = toSql(block.data.header.timestamp),
+        internal = a.internal
       )
   }
 
   private val convertReveal: PartialFunction[(Block, OperationHash, Operation), Tables.OperationsRow] = {
-    case (block, groupHash, Reveal(counter, fee, gas_limit, storage_limit, pk, source, metadata)) =>
+    case (block, groupHash, r @ Reveal(counter, fee, gas_limit, storage_limit, pk, source, metadata)) =>
       Tables.OperationsRow(
         operationId = 0,
         operationGroupHash = groupHash.value,
@@ -188,7 +191,8 @@ object DatabaseConversions {
         consumedGas = metadata.operation_result.consumed_gas.flatMap(extractBigDecimal),
         blockHash = block.data.hash.value,
         blockLevel = block.data.header.level,
-        timestamp = toSql(block.data.header.timestamp)
+        timestamp = toSql(block.data.header.timestamp),
+        internal = r.internal
       )
   }
 
@@ -196,7 +200,7 @@ object DatabaseConversions {
     case (
         block,
         groupHash,
-        Transaction(counter, amount, fee, gas_limit, storage_limit, source, destination, parameters, metadata)
+        t @ Transaction(counter, amount, fee, gas_limit, storage_limit, source, destination, parameters, metadata)
         ) =>
       Tables.OperationsRow(
         operationId = 0,
@@ -216,7 +220,8 @@ object DatabaseConversions {
         paidStorageSizeDiff = metadata.operation_result.paid_storage_size_diff.flatMap(extractBigDecimal),
         blockHash = block.data.hash.value,
         blockLevel = block.data.header.level,
-        timestamp = toSql(block.data.header.timestamp)
+        timestamp = toSql(block.data.header.timestamp),
+        internal = t.internal
       )
   }
 
@@ -224,7 +229,7 @@ object DatabaseConversions {
     case (
         block,
         groupHash,
-        Origination(
+        o @ Origination(
           counter,
           fee,
           source,
@@ -261,12 +266,13 @@ object DatabaseConversions {
         paidStorageSizeDiff = metadata.operation_result.paid_storage_size_diff.flatMap(extractBigDecimal),
         blockHash = block.data.hash.value,
         blockLevel = block.data.header.level,
-        timestamp = toSql(block.data.header.timestamp)
+        timestamp = toSql(block.data.header.timestamp),
+        internal = o.internal
       )
   }
 
   private val convertDelegation: PartialFunction[(Block, OperationHash, Operation), Tables.OperationsRow] = {
-    case (block, groupHash, Delegation(counter, source, fee, gas_limit, storage_limit, delegate, metadata)) =>
+    case (block, groupHash, d @ Delegation(counter, source, fee, gas_limit, storage_limit, delegate, metadata)) =>
       Tables.OperationsRow(
         operationId = 0,
         operationGroupHash = groupHash.value,
@@ -281,7 +287,8 @@ object DatabaseConversions {
         consumedGas = metadata.operation_result.consumed_gas.flatMap(extractBigDecimal),
         blockHash = block.data.hash.value,
         blockLevel = block.data.header.level,
-        timestamp = toSql(block.data.header.timestamp)
+        timestamp = toSql(block.data.header.timestamp),
+        internal = d.internal
       )
   }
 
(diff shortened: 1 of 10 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -246,9 +247,60 @@ object TezosTypes {
...
       operation_result: RESULT,
-      balance_updates: List[OperationMetadata.BalanceUpdate]
+      balance_updates: List[OperationMetadata.BalanceUpdate],
+      internal_operation_results: Option[List[InternalOperationResults.InternalOperationResult]] = None
   )
 
+// Internal operations result definitions
+  object InternalOperationResults {
+
+    sealed trait InternalOperationResult {
```
ivanopagano: Extending Product with Serializable should generally be considered good practice

src/main/scala/tech/cryptonomic/conseil/tezos/JsonDecoders.scala
```diff
@@ -105,9 +106,9 @@ object JsonDecoders {
 
     /* Collects definitions to decode delegates and their contracts */
     object Delegates {
-      //reusing much of the values used in operations
       import Numbers._
       import Scripts._
+      //reusing much of the values used in operations
```
ivanopagano: Why moving the comment? It was referring to the import clauses 😄

src/main/scala/tech/cryptonomic/conseil/tezos/JsonDecoders.scala
```diff
@@ -224,6 +227,26 @@ object JsonDecoders {
       implicit val transactionMetadataDecoder: Decoder[ResultMetadata[OperationResult.Transaction]] = deriveDecoder
       implicit val originationMetadataDecoder: Decoder[ResultMetadata[OperationResult.Origination]] = deriveDecoder
       implicit val delegationMetadataDecoder: Decoder[ResultMetadata[OperationResult.Delegation]] = deriveDecoder
+      implicit val internalOpResultsDecoder: Decoder[InternalOperationResults.InternalOperationResult] =
```
ivanopagano: Just wondering if it might work

Considering that
- there's a custom derivation config using "kind" as discriminator field,
- we have a separate object/module scope than the "external" operations, to avoid name clashes

Do you think that renaming the internal operation results exactly as their kind could allow for a single derivation, i.e.

```scala
implicit val internalOperationResultDecoder: Decoder[InternalOperationResult.InternalOperationResult] =
  deriveDecoder
```
Letting circe handle the ADT hierarchy?

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -372,6 +368,74 @@ object DatabaseConversions {
...
+              .withFieldConst(_.storage_limit, InvalidPositiveDecimal("Discarded"))
+              .withFieldConst(
+                _.metadata,
+                ResultMetadata[OperationResult.Delegation](delegation.result, List.empty, None)
+              )
+              .transform
+        }
+    }
+
+  def appendToFile(str: String) =
```
ivanopagano: Leftover?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -162,7 +162,9 @@ object TezosTypes {
   }
 
   /** root of the operation hiearchy */
-  sealed trait Operation extends Product with Serializable
+  sealed trait Operation extends Product with Serializable {
+    val internal: Boolean = false
```
ivanopagano: I'm almost sure this isn't needed?

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -306,22 +313,11 @@ object DatabaseConversions {
         kind = kind,
         blockHash = block.data.hash.value,
         blockLevel = block.data.header.level,
-        timestamp = toSql(block.data.header.timestamp)
+        timestamp = toSql(block.data.header.timestamp),
+        internal = op.internal
       )
   }
 
-  implicit val blockToOperationsRow = new Conversion[List, Block, Tables.OperationsRow] {
```
ivanopagano: Wasn't this actually used? Very good catch, then!

src/test/scala/tech/cryptonomic/conseil/tezos/JsonDecodersTest.scala
```diff
@@ -341,7 +341,18 @@ class JsonDecodersTest extends WordSpec with Matchers with EitherValues with Opt
 
       }
 
-      "decode an transaction operation from json" in new OperationsJsonData {
+    "decode an transaction with internal transaction result operation from json" in new OperationsJsonData {
```
ivanopagano: For me it's fine to remove the other test entirely: I don't think we're supposed to handle differently transactions, based on the presence or absence of internal results.

Less code to maintain 😁

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
piotrkosecki: yes! I thought I got rid of all debugging 😄

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
piotrkosecki: It wasn't used and at first I spent some time trying to figure out why it's not working 😄

src/main/scala/tech/cryptonomic/conseil/tezos/JsonDecoders.scala
piotrkosecki: something must've gone wrong with code reformatting/merge 😄

src/main/scala/tech/cryptonomic/conseil/tezos/JsonDecoders.scala
piotrkosecki: unfortunately I couldn't make this circe derivation work, thats why I used this custom decoder

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
piotrkosecki: it's needed, I use it in the Operation -> OperationRow transformations

src/test/scala/tech/cryptonomic/conseil/tezos/JsonDecodersTest.scala
piotrkosecki: I wanted to test if it's correctly deserialized, I think it's needed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -246,9 +247,60 @@ object TezosTypes {
...
-      balance_updates: List[OperationMetadata.BalanceUpdate]
+      balance_updates: List[OperationMetadata.BalanceUpdate],
+      internal_operation_results: Option[List[InternalOperationResults.InternalOperationResult]] = None
   )
 
+// Internal operations result definitions
+  object InternalOperationResults {
+
+    sealed trait InternalOperationResult {
+      val source: ContractId
```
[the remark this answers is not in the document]
piotrkosecki: I removed unused fields, changed to def

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
piotrkosecki: ok, added

src/main/scala/tech/cryptonomic/conseil/tezos/JsonDecoders.scala
piotrkosecki: nvm, apparently I've done something wrong before, now it works, thanks!

src/test/scala/tech/cryptonomic/conseil/tezos/JsonDecodersTest.scala
ivanopagano: You got me wrong, I'd keep the one you added now. I'd get rid of the previous one for transaction, this one looks more complete

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
ivanopagano: If I look at the conversion from `block` where you actually convert the `internalRows` in
[DatabaseConversions](https://github.com/Cryptonomic/Conseil/pull/479/files#diff-36344a07cac3951e698e283f1b1bf238R462), you correctly take care to 
- convert each internal to a regular operation
- convert the result to a db operation row
- copy the result to _set internal to true_

Thus you can actually set internal to the row representation only, with no need to propagate the value in the domain model of Operations and InternalOperations

src/test/scala/tech/cryptonomic/conseil/tezos/JsonDecodersTest.scala
piotrkosecki: ok, removed, but coverage failed: -0,02% 😄

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
piotrkosecki: you're right, I did not get it at first, sure, changed

PR #482, opened 2019-07-17
Conseil 452 add ballots to operations

### LINKED ISSUE #484: Expose ballot voting data
Some ballot voting data is not exposed in ConseilJS.

It's particularly these RPC routes that would be of interest:
- /chains/main/blocks/head/votes/ballots
- /chains/main/blocks/head/votes/listings

It would be convenient with a function in ConseilJS that returned the data from the /votes/ballots route and also attached the number of unused votes. The data from votes/listings could be merged into getBallots(), so it returned {pkh, ballot, rolls, block_id, block_level} instead.

ba205:
Adding ballots to operations table.

Closes #484.


src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -233,7 +233,12 @@ object TezosTypes {
   final case object DoubleEndorsementEvidence extends Operation
   final case object DoubleBakingEvidence extends Operation
   final case object Proposals extends Operation
-  final case object Ballot extends Operation
+  final case class Ballot(ballot: Vote) extends Operation
+
+  sealed trait Vote extends Product with Serializable
```
ivanopagano: I might try to suggest reusing an existing data, which is defined as `Voting.Vote`
It's simply a String wrapper, but
1. there's already a decoder defined for it, therefore we just need to `import Votes._` inside the `JsonDecoders.Circe.Operations` object definition and should be all set up
2. the existing decoder takes care to only correctly decode expected values (look at the definition in the JsonDecoders file
3. the votes are eventually stored as Strings in the table conversion, so we won't even need to pattern match, simply extract the string from the Votes value
4. it'll avoids different overlapping definition of a essentially the same concept (if we go for the enumeration encoding, I'd be keen to remove the existing one for a single way anyway)

I leave it to you to choose what you find best

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -233,7 +233,12 @@ object TezosTypes {
   final case object DoubleEndorsementEvidence extends Operation
   final case object DoubleBakingEvidence extends Operation
   final case object Proposals extends Operation
-  final case object Ballot extends Operation
+  final case class Ballot(ballot: Vote) extends Operation
```
ivanopagano: You might need to look into decodings and conversions tests both to add tests for the new operation detail

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
ba205: I think this isn't high priority enough to go through the effort, I'll use your Vote. Thank you!

doc/conseil.sql
```diff
@@ -2,10 +2,8 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 9.5.17
--- Dumped by pg_dump version 11.4 (Ubuntu 11.4-1.pgdg18.04+1)
-
--- Started on 2019-06-27 13:32:57 EDT
+-- Dumped from database version 11.4
```
vishakh: We run Postgres 9 in production. Could you dump this file for that version so we don't have compatibility issues?

doc/conseil.sql
```diff
@@ -268,12 +254,12 @@ CREATE TABLE public.operations (
     originated_contracts character varying,
     block_hash character varying NOT NULL,
     block_level integer NOT NULL,
-    "timestamp" timestamp without time zone NOT NULL
+    "timestamp" timestamp without time zone NOT NULL,
+    ballot character varying
```
vishakh: Please swap the order of these columns.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -233,7 +233,7 @@ object TezosTypes {
   final case object DoubleEndorsementEvidence extends Operation
   final case object DoubleBakingEvidence extends Operation
   final case object Proposals extends Operation
-  final case object Ballot extends Operation
+  final case class Ballot(ballot: Voting.Vote) extends Operation
```
vishakh: How come this is a case class and not case object like the others?

src/test/scala/tech/cryptonomic/conseil/tezos/DatabaseConversionsTest.scala
```diff
@@ -819,13 +819,14 @@ class DatabaseConversionsTest
 
       "convert a Ballot operation to a database row" in {
 
-        val converted = (block, groupHash, Ballot: Operation).convertTo[Tables.OperationsRow]
+        val converted = (block, groupHash, sampleBallot: Operation).convertTo[Tables.OperationsRow]
```
vishakh: The test above refers to a type from `TezosTypes` while this this test refers to a local object. Why is this?

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTestFixtures.scala
```diff
@@ -721,9 +722,14 @@ trait TezosDataGeneration extends RandomGenerationKit {
...
 
+    val sampleBallot =
+      Ballot(
+        ballot = Vote("yay")
+      )
+
     val sampleOperations =
       sampleEndorsement :: sampleNonceRevelation :: sampleAccountActivation :: sampleReveal :: sampleTransaction :: sampleOrigination :: sampleDelegation ::
-          DoubleEndorsementEvidence :: DoubleBakingEvidence :: Proposals :: Ballot :: Nil
+          DoubleEndorsementEvidence :: DoubleBakingEvidence :: Proposals :: sampleBallot :: Nil
```
vishakh: Same observation here..

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
ba205: because it has a parameter, like the earlier operations.

src/test/scala/tech/cryptonomic/conseil/tezos/DatabaseConversionsTest.scala
ba205: because ballot now has a parameter, like sampleDelegation/other operations that are case objects instead of case classes in TezosTypes

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTestFixtures.scala
ba205: It's because of the change from case class to case object.

doc/conseil.sql
ba205: I deleted the only difference between this version and postgres 9 as a temp fix. I'll need to uninstall/reinstall Postgres for the future.

doc/conseil.sql
```diff
@@ -391,112 +362,104 @@ ALTER TABLE ONLY public.operations
 
 
 --
--- TOC entry 2064 (class 1259 OID 7467561)
 -- Name: fki_block; Type: INDEX; Schema: public; Owner: -
 --
 
 CREATE INDEX fki_block ON public.operation_groups USING btree (block_id);
 
-CREATE INDEX ix_operation_groups_block_level ON public.operation_groups USING btree (block_level);
```
ba205: Note, this simply moved down in the file for some reason. I did not, in fact, remove an index.

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -30,6 +30,16 @@ object DatabaseConversions {
     case _ => None
   }
 
+  def extractBallot(ballot: Voting.Vote): Option[String] = ballot match {
```
ivanopagano: This function seems redundant to me, if the `Ballot` operation always contains a `Voting.Vote` there's no need to pattern match, but we can read the vote value and wrap it in a `Some`.
Follow-up in-lined in the later comment for `convertBallot`

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -298,13 +309,29 @@ object DatabaseConversions {
...
+        operationId = 0,
+        operationGroupHash = groupHash.value,
+        kind = "ballot",
+        blockHash = block.data.hash.value,
+        blockLevel = block.data.header.level,
+        timestamp = toSql(block.data.header.timestamp),
+        ballot = extractBallot(ballot),
+        internal = false,
+        proposal = proposal,
+        source = extractContractId(source)
```
ivanopagano: I suggest: `source = source.map(_.id)`

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -298,13 +309,29 @@ object DatabaseConversions {
...
+  private val convertBallot: PartialFunction[(Block, OperationHash, Operation), Tables.OperationsRow] = {
+    case (block, groupHash, Ballot(ballot, proposal, source)) =>
+      Tables.OperationsRow(
+        operationId = 0,
+        operationGroupHash = groupHash.value,
+        kind = "ballot",
+        blockHash = block.data.hash.value,
+        blockLevel = block.data.header.level,
+        timestamp = toSql(block.data.header.timestamp),
+        ballot = extractBallot(ballot),
```
ivanopagano: I suggest `ballot = Some(ballot.value)`

src/test/scala/tech/cryptonomic/conseil/tezos/DatabaseConversionsTest.scala
```diff
@@ -819,14 +819,16 @@ class DatabaseConversionsTest
...
 
         converted.operationId shouldBe 0
         converted.operationGroupHash shouldBe groupHash.value
         converted.blockHash shouldBe block.data.hash.value
         converted.blockLevel shouldBe block.data.header.level
         converted.timestamp shouldBe Timestamp.from(block.data.header.timestamp.toInstant)
-        converted.kind shouldBe "ballot"
+        converted.ballot shouldBe Some("yay")
```
ivanopagano: The newly-added fields in the operations (i.e. ballot, source), should now be added to the list of empty fields checked in all the other operations' cases.

I'm referring to the checks that go
```
forAll(
  ... a long chained list of converted.fieldname
) {
  _ shouldBe 'empty
}
```

PR #487, opened 2019-07-23
469 - query aggregarion predicate does not work correctly

### LINKED ISSUE #469: Query aggregation predicate does not use correct field
Example query:
```
{
	"aggregation": [{
		"field": "level",
		"function": "count",
		"predicate": {
			"field": "count_level",
			"inverse": false,
			"operation": "gt",
			"set": [10]
		}
	}],
	"fields": ["baker", "level"],
	"limit": 50,
	"orderBy": [{
		"direction": "desc",
		"field": "count_level"
	}],
	"output": "json",
	"predicates": [{
		"field": "timestamp",
		"inverse": false,
		"operation": "between",
		"set": [1554076800000, 1556668799000]
	}]
}
```
results with this kind of SQL query:
```
SELECT COUNT(level) as count_level,baker FROM blocks WHERE true  AND level > '10' AND timestamp BETWEEN '2019-04-01 00:00:00.0' AND '2019-04-30 23:59:59.0' GROUP BY baker ORDER BY count_level desc LIMIT 50
```

piotrkosecki:
resolves #469 - the issue was that I was not using `HAVING` for the aggregation predicate.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
@@ -510,8 +510,9 @@ object TezosDatabaseOperations extends LazyLogging {
   )(implicit ec: ExecutionContext): DBIO[List[QueryResponse]] = {
 
     val q = makeQuery(table, columns, aggregation)
-      .addPredicates(aggregation.flatMap(_.getPredicate) ::: predicates)
+      .addPredicates(predicates)
       .addGroupBy(aggregation, columns)
+      .addHaving(aggregation)
       .addOrdering(ordering)
       .addLimit(limit)
     if (outputType == OutputType.sql) {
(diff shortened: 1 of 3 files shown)


vishakh:
Could you provide a brief overview of the change in the SQL generation behaviour?

piotrkosecki:
@vishakh for the example described in the #469  now the SQL looks like this:
```
SELECT COUNT(level) as count_level,baker FROM blocks WHERE true  AND level > '10' AND timestamp BETWEEN '2019-04-01 02:00:00.0' AND '2019-05-01 01:59:59.0' GROUP BY baker ORDER BY count_level desc LIMIT 50
```
After my change will look like this:
```
SELECT COUNT(level) as count_level,baker FROM blocks WHERE true  AND timestamp BETWEEN '2019-04-01 02:00:00.0' AND '2019-05-01 01:59:59.0' GROUP BY baker HAVING true AND COUNT(level) > '10' ORDER BY count_level desc LIMIT 50
```

PR #491, opened 2019-07-26
478 - integration with nautilus cloud

### LINKED ISSUE #478: Integration with Nautilus Cloud
Conseil currently picks up valid API keys from its own config file. It should optionally also check for API keys from a given Nautilus Cloud instance:
- The config file should have an optional entry for a Nautilus Cloud provider with any other settings that are required.
- Conseil should somehow cache valid API keys so it doesn't have to hit Nautilus Cloud too often.

piotrkosecki:
Resolves #478 

Added basic API keys updating from nautilus-cloud

### Configuration changes
```
nautilus-cloud {
  host = "http://localhost"
  port = 1234
  path = "apiKeys/<envName>"
  key = "exampleApiKey"
  delay = 10 seconds
  interval = 30 seconds
}
```


src/main/scala/tech/cryptonomic/conseil/config/Shared.scala
```diff
@@ -47,3 +47,12 @@ object Natural {
 }
 
 final case class VerboseOutput(on: Boolean) extends AnyVal
+
+final case class NautilusCloudConfiguration(
```
bhaskarkishore: Please add a doc string.

src/main/scala/tech/cryptonomic/conseil/config/Security.scala
```diff
@@ -1,8 +1,38 @@
...
 
-object Security {
+import scala.concurrent.{ExecutionContext, Future}
+
+object Security extends ErrorAccumulatingCirceSupport {
+
+  /** Keys cache */
+  private val nautilusCloudKeys: Ref[IO, Set[String]] = Ref.of[IO, Set[String]](Set.empty).unsafeRunSync()
+
+  def updateKeys(
```
bhaskarkishore: Please add a doc string.

Additionally, it would be nice to get some logging from this function (or where ever appropriate) incase the key fetch from Nautilus cloud fails.

piotrkosecki: added

src/main/scala/tech/cryptonomic/conseil/config/Shared.scala
piotrkosecki: added

src/main/scala/tech/cryptonomic/conseil/config/Security.scala
```diff
@@ -1,8 +1,46 @@
...
+      ncc: NautilusCloudConfiguration
+  )(implicit executionContext: ExecutionContext, system: ActorSystem, mat: Materializer): Unit = {
+    val update = for {
+      apiKeys <- Http()
+        .singleRequest(
+          HttpRequest(uri = makeUri(ncc))
+            .withHeaders(RawHeader("X-Api-Key", ncc.key))
+        )
+        .flatMap(Unmarshal(_).to[Set[String]])
+      _ <- nautilusCloudKeys.update(_ => apiKeys).unsafeToFuture()
```
ivanopagano: It's possible to simply `nautilusCloudKeys.set(apiKeys).[...]`

src/main/scala/tech/cryptonomic/conseil/config/Security.scala
```diff
@@ -1,8 +1,46 @@
...
+    } yield logger.info(s"Managed to update api keys with $apiKeys")
+    update onComplete {
```
ivanopagano: I usually call `.andThen` to log failures, as it accepts a PartialFunction, so I can avoid writing an empty success case.

In this situation tho, I'd only log success if the outcome is thus, therefore:

```scala
val update = for {
  ...
} yield ()
update onComplete {
  case Success(_) => logger.info("Managed to update api keys with {}", apiKeys)
  case Failure(exception) => logger.error("Error during API keys update", exception)
}
```

ivanopagano: On a different note, my personal stance when writing log message, is that we should stop using messages that sounds like they're spit out from the first Terminator version or some '70s TV show.
Even Hal in Space Odissey conveys information like a human... I try to be follow its lead for a better user experience. 😁

src/main/scala/tech/cryptonomic/conseil/config/Security.scala
```diff
@@ -12,10 +50,12 @@ object Security {
...
       * Determines whether a given API key is valid.
-      * @param candidateApiKey  The given API key
-      * @return                 True is valid, false otherwise.
+      *
+      * @param candidateApiKey The given API key
+      * @return True is valid, false otherwise.
       */
-    def validateApiKey(candidateApiKey: String): Boolean = keys(candidateApiKey)
+    def validateApiKey(candidateApiKey: String): Future[Boolean] =
+      nautilusCloudKeys.get.map(ncKeys => (ncKeys ++ keys)(candidateApiKey)).unsafeToFuture()
```
ivanopagano: For un-named sets like in this case, would you say that using `contains` explicitly is easier to understand?

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -29,18 +30,23 @@ object Conseil
...
-
+    case Right((server, platforms, securityApi, verbose, metadataOverrides, nautilusCloud)) =>
       implicit val system: ActorSystem = ActorSystem("conseil-system")
       implicit val materializer: ActorMaterializer = ActorMaterializer()
       implicit val executionContext: ExecutionContextExecutor = system.dispatcher
+      nautilusCloud.map { ncc =>
+        system.scheduler.schedule(ncc.delay, ncc.interval)(Security.updateKeys(ncc))
+      }
 
+      val validateApiKey: Directive[Tuple1[String]] = headerValueByName("apikey").tflatMap[Tuple1[String]] {
```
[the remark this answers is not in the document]
piotrkosecki: we had it, but it was removed, I think after we started using `endpoints` library, I found this directive here - should I move it somewhere?

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -29,18 +30,23 @@ object Conseil
...
+        case Tuple1(apiKey) =>
+          onComplete(securityApi.validateApiKey(apiKey)).flatMap {
+            case Success(true) => provide(apiKey)
+            case _ => complete((Unauthorized, "Incorrect API key"))
+          }
+        case _ =>
+          complete((Unauthorized, "Incorrect API key"))
```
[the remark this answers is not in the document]
piotrkosecki: ok, changed

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -29,18 +30,23 @@ object Conseil
...
-          provide(apiKey)
-        case _ =>
-          complete((Unauthorized, "Incorrect API key"))
-      }
-
+    case Right((server, platforms, securityApi, verbose, metadataOverrides, nautilusCloud)) =>
       implicit val system: ActorSystem = ActorSystem("conseil-system")
       implicit val materializer: ActorMaterializer = ActorMaterializer()
       implicit val executionContext: ExecutionContextExecutor = system.dispatcher
+      nautilusCloud.map { ncc =>
```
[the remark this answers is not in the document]
piotrkosecki: sure, good catch

src/main/scala/tech/cryptonomic/conseil/config/Security.scala
piotrkosecki: ok, done

src/main/scala/tech/cryptonomic/conseil/config/Security.scala
bhaskarkishore: Wait a minute, both the terminator and HAL are homicidal. Now if we are using that as critical criteria, may i humbly put forth that we use Bender from Futurama as an exemplar. :)

src/main/scala/tech/cryptonomic/conseil/config/Security.scala
```diff
@@ -1,8 +1,46 @@
...
+  )(implicit executionContext: ExecutionContext, system: ActorSystem, mat: Materializer): Unit = {
+    val update = for {
+      apiKeys <- Http()
+        .singleRequest(
+          HttpRequest(uri = makeUri(ncc))
+            .withHeaders(RawHeader("X-Api-Key", ncc.key))
+        )
+        .flatMap(Unmarshal(_).to[Set[String]])
+      _ <- nautilusCloudKeys.update(_ => apiKeys).unsafeToFuture()
+    } yield logger.info(s"Managed to update api keys with $apiKeys")
```
bhaskarkishore: I may be wrong, but this line seems to output all the fetched keys into the log. That may not be ideal, especially when we may have many many keys.

src/main/scala/tech/cryptonomic/conseil/config/Security.scala
piotrkosecki: ok, changed. 

Regarding log messages. I like terminator way of logging. It is concise and brings only useful information. Did you have in mind something like this?: `Hello dear user. We've encountered some problems during API keys update request. But do not worry, here you can find what went wrong:` 😄 (I've watched Space Odyssey long time ago)

src/main/scala/tech/cryptonomic/conseil/config/Security.scala
piotrkosecki: ok, changed

src/main/scala/tech/cryptonomic/conseil/config/Security.scala
piotrkosecki: ok, I found it useful to check if the keys were loaded correctly. What do you think about changing it to `debug` level? Or do you prefer just removing them from the log?

bhaskarkishore: If would prefer keys to not be in the log under normal operations, just a one liner saying that the fetch was successful.

If you feel it useful, you can add a secondary debug statement for printing the keys out.

piotrkosecki: ok, it was useful during development, now I removed it

src/main/scala/tech/cryptonomic/conseil/config/Security.scala
```diff
@@ -1,8 +1,47 @@
...
+  private val nautilusCloudKeys: Ref[IO, Set[String]] = Ref.of[IO, Set[String]](Set.empty).unsafeRunSync()
+
+  /** Updates API keys from Nautilus-Cloud endpoint */
+  def updateKeys(
+      ncc: NautilusCloudConfiguration
+  )(implicit executionContext: ExecutionContext, system: ActorSystem, mat: Materializer): Unit = {
+    val update = for {
+      apiKeys <- Http()
+        .singleRequest(
+          HttpRequest(uri = makeUri(ncc))
```
bhaskarkishore: Can this handle HTTPS requests?

piotrkosecki: yes, the same method is used in `TezosNodeOperator` which usually uses HTTPS for requests

src/main/scala/tech/cryptonomic/conseil/config/Security.scala
piotrkosecki: @bhaskarkishore Bender? `Here you go, your request failed. I'll make my own request, with blackjack and hookers. In fact, forget the request` 🤣

ivanopagano: I was aiming more at something like the Downton Abbey's butler...

"Sir, I beg your pardon. It's quite embarrassing yet the updates you requested were temporarily unavailable, may I take care to provide you with a suitable alternative right away or has milord changed his mind?"

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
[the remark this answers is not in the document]
ivanopagano: nope, then, thanks

PR #496, opened 2019-08-02
187 - forks testing scenarios

### LINKED ISSUE #187: Writing mock node tests and mock database queries to test forking algorithm
https://github.com/Cryptonomic/Conseil/pull/161 addresses #48, and the core algorithm is implemented, and five different chains with simulated Lorre iterations have been simulated using mock tezos nodes.

To complete testing of the core forking algorithm, we want to do the following:

- [ ] In test/.../MockTezosNodes.scala, implement an addition to the NormalChain, OneBranchFork, OneBranchForkAlternating, TwoBranchFork, and TwoBranchForkAlternating objects (maybe turn them into classes) that allows them to have a list of MockTezosNodes, and switch between them.
- [ ] In test/.../TezosNodeOperatorTest.scala, write unit tests for getBlocksNotInDatabase for NormalChain, OneBranchFork, OneBranchForkAlternating, TwoBranchFork, and TwoBranchForkAlternating mock nodes. This may include writing unit tests for functions that getBlocksNotInDatabase calls, as well as editing the mock node definitions to include operations (probably an empty list would be fine)
- [ ] Write Mock database queries using the InMemoryDatabase class to test Lorre.scala's processTezosBlocks Function, when given NormalChain, OneBranchFork, OneBranchForkAlternating, TwoBranchFork, and TwoBranchForkAlternating as mock nodes. This is crucial. If the database is as expected, we can be reasonably convinced that Lorre can handle forks.

ivanopagano:
closes #187  and #48 

The PR includes the code to record invalidated blocks after a fork changed history of the near chain from the previous Lorre's cycle.

Included is a test suite to verify the most significant and fundamental forking scenarios.

In addition, the read side of the application (Conseil app) will now take the invalidation into account when returning results, therefore invalidated blocks and other entities stored via an invalidated block will not show in practically any external rest query.

The generation of predicated queries has been reviewed to accommodate the need for custom query definitions, based on the specific table considered.
Hence the QueryBuilder has been introduced as a trait too, that can be dynamically invoked based on table queried, returning a specific instance that handles the invalidation appropriately.

Extra details on the generated queries can be printed on the logs by enabling the "debbugging" level in the logback.xml file for the right category

```xml
    <logger name="tech.cryptonomic.conseil.tezos.TezosDatabaseOperations$" level="DEBUG"/>
```

## Database schema changes introduced

A new table is added to keep trace of invalidation
```sql
CREATE TABLE public.invalidated_blocks (
    hash character varying PRIMARY KEY,
    level integer NOT NULL,
    is_invalidated boolean NOT NULL
);
```

### ApiOperations

- fetchBlock/fetchLatestBlock ✓
- fetchAccount ✓
- fetchOperationGroup (see TezosDatabaseOperations#operationsForGroup) ✓
- queryWithPredicates (see TezosDatabaseOperations#selectWithPredicates) ✓

### TezosDatabaseOperations

Operations
- operationsForGroup ✓
- operationsByGroupHash 𐄂 (solved at the caller)

Meta
- countRows (ignoring)
- countDistinct (ignoring)
- selectDistinct (ignoring)
- selectDistinctLike (ignoring)
- selectWithPredicates ✓

Fees
- calculateAverageFees ✓

As can be seen, the metadata counts and attributes' queries are not currently modified, considering the relatively low impact that a few additional blocks' data might have on a long-running instance of tezos, and the mostly statistical use we make of this data.

## the Checkpoint's problem
If we stop to consider that checkpointing keeps data between Lorre cycles, in case something didn't complete correctly,
we would realise that between calls to store the account ids and retrieve the accounts' data, or store delegates' key-hashes and retrieving delegates,
a fork might happen that actually invalidates the block-level and block-id references, as those blocks might've not been included in the new fork.

Thus we should invalidate those references too, and avoid to load such data?

On a second thought, even if we could load those values, they will reference invalidated blocks, which would not create any real issue, as for any other data
related to a now-invalid fork.

But realistically we should expect the values to be missing, thus actually failing the fetch request, possibly self-referentially failing the operation in the first place.


coverage.sbt
```diff
@@ -1,6 +1,10 @@
 coverageExcludedPackages := Seq(
   "<empty>",
   ".*\\.tezos.Tables",
+  ".*\\directives.EnableCORSDirectives",
```
ba205: Why does this pr specifically require updates to our coverage report?

src/main/scala/tech/cryptonomic/conseil/tezos/michelson/renderer/MichelsonRenderer.scala
```diff
@@ -52,7 +52,7 @@ object MichelsonRenderer {
       self
         .map(_.render())
         .mkString(" ;\n")
-        .lines
+        .linesIterator
```
ba205: pull master into this branch, this was fixed by @piotrkosecki in a different pr

coverage.sbt
ivanopagano: It must be an old change introduced by the initial versions.

Possibly I needed to start filtering out code that we didn't really expect to be testing, to get a realistic coverage %

src/main/scala/tech/cryptonomic/conseil/tezos/michelson/renderer/MichelsonRenderer.scala
ivanopagano: I will, sure, I fixed it here before

src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
```diff
@@ -5,18 +5,44 @@ import slick.jdbc.{GetResult, PositionedParameters, SQLActionBuilder}
...
   lazy val db = Database.forConfig("conseildb")
 
+  trait QueryBuilder[+T <: Table[_]] {
+    def makeQuery(
+        table: String,
+        columns: List[String],
+        aggregation: List[Aggregation]
+    ): SQLActionBuilder
+
+    /* use this to know how the blocks invalidation table is named*/
```
piotrkosecki: this and two other below: lost a space before closing comment

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -508,14 +647,29 @@ object TezosDatabaseOperations extends LazyLogging {
       outputType: OutputType,
       limit: Int
   )(implicit ec: ExecutionContext): DBIO[List[QueryResponse]] = {
+    /* this will get into scope the conversion needed from TableRef to a full TableQuery
+     * so that we can ask for it using only the table name string
+     */
+    import tech.cryptonomic.conseil.tezos.DatabaseQueries._
+
+    val allPredicates = aggregation.flatMap(_.getPredicate) ::: predicates
```
piotrkosecki: this will need to change as aggregation predicates need to use `HAVING` clause

piotrkosecki:
I focused mostly on query part, and it looks ok, beside the fact that it will need to change a bit with newest changes from master(for example adding `HAVING` clause). Also it may need additional changes if the PR #505 will be merged first.

src/main/scala/tech/cryptonomic/conseil/util/DatabaseUtil.scala
ivanopagano: Might be the formatter? I don't think it's manual intervention from me

ivanopagano:
I plan to discuss with @anonymoussprocket to split this monster PR into digestible smaller units, to integrate over time into `master`, taking care that no regression is introduced at each step

vishakh:
Closing for now as the PR has become quite stale. We will return to this once the big ticket items for Conseil are done.

PR #498, opened 2019-08-02
documentation update

anonymoussprocket:


docs/README.md
```diff
@@ -2,7 +2,7 @@
...
 
-Conseil is a fundamental part of the [Nautilus](https://github.com/Cryptonomic/Nautilus) infrastructure for high-performance chain analytics. It is the backbone of Cryptonomic's product offerings. The [Arronax]() block explorer and reporting tool, and the [Tezori](https://github.com/Cryptonomic/Tezori)/[Galleon](https://galleon-wallet.tech) wallet are both made possible by Conseil. The aforememtion products are additionally using [ConseilJS](https://github.com/Cryptonomic/ConseilJS) &ndash; a Typescript wrapper with a Tezos node interface.
+Conseil is a fundamental part of the [Nautilus](https://github.com/Cryptonomic/Nautilus) infrastructure for high-performance chain analytics. It is the backbone of Cryptonomic's product offerings. The [Arronax](https://arronax-beta.cryptonomic.tech) block explorer and reporting tool, and the [Tezori](https://github.com/Cryptonomic/Tezori)/[Galleon](https://galleon-wallet.tech) wallet are both made possible by Conseil. The aforememtion products are additionally using [ConseilJS](https://github.com/Cryptonomic/ConseilJS) &ndash; a Typescript wrapper with a Tezos node interface.
```
ivanopagano: typographic maniac: the one in the end should be an `&mdash;` to be correct.

docs/README.md
```diff
@@ -59,11 +59,40 @@ Both `conseil` and `lorre` write to `syslog`. The logs are verbose enough to det
 
 ### Configuration
 
-`conseil` and `lorre` use [HOCON](https://github.com/lightbend/config/blob/master/HOCON.md) format to store configuration. Read more at [Typesafe Config] github page(https://github.com/lightbend/config)
+`conseil` and `lorre` use [HOCON](https://github.com/lightbend/config/blob/master/HOCON.md) format to store configuration. Read more at [Typesafe Config](https://github.com/lightbend/config) github page.
```
ivanopagano: Maybe we should take the change and rename that to Lightbend Config.
The company changed name years ago

docs/README.md
```diff
@@ -59,11 +59,40 @@ Both `conseil` and `lorre` write to `syslog`. The logs are verbose enough to det
...
 
-### Datasource configuration
+### Metadata overrides
+
+Metadata enables dynamic querying of data sources provided by Conseil and allows for chain-agnostic UIs, like Arronax, to be built while still providing a high level of specialization. Some of the metadata override keys are helpful for query construction, but a large number of them exist for rendering.
+
+In addition to the metadata derived from the schema, it's possible to extend and override it. This information is stored in configuration files inside `/src/main/resources/metadata/` which are referenced from `/src/main/resources/metadata.conf` that is itself included from `src/main/resources/reference.conf`. Metadata hierarchy is platform/network/entity/attribute, different options are available at each level.
+
+Note the naming convention in this file. Hyphenated keys become camel-cased when presented via the metadata service in JSON format. For example `display-name` in the config file appears as `displayName` in the JSON response.
```
ivanopagano: Would it help to write it as "camelCased" to help out the reader?
Like this it creates a cognitive dissonance reading one convention expressed (by chance) using another

anonymoussprocket: wouldn't that be grammatically incorrect?

docs/README.md
anonymoussprocket: changed

docs/README.md
```diff
@@ -59,11 +59,40 @@ Both `conseil` and `lorre` write to `syslog`. The logs are verbose enough to det
...
+- `visible`: Item visibility, default is `false`. This means when a new entity is added to the Conseil database, it must be explicitly enabled.
+
+#### Entity metadata
+
+- `display-name-plural`: Arronax uses this to render entity tab titles here it will show "Blocks" instead of "Block".
+
+#### Attribute metadata
+
+- `currency-symbol`: Associated with Currency type fields, contains currency symbol or code, for example, 'XTZ', or 'ꜩ'
+- `currency-symbol-code`: Same as `currency-symbol`, but stores the Unicode index of the currency symbol, for 'ꜩ' it would be `42793`
```
ivanopagano: Isn't the "Unicode index" technically named `code-point`? Using that removes any possible ambiguity

docs/README.md
```diff
@@ -59,11 +59,40 @@ Both `conseil` and `lorre` write to `syslog`. The logs are verbose enough to det
...
+- `data-format`: Presently used to customize date format. String contents of this item have no impact on how the data is sent or queried, it would be up to the implementing UI to interpret this content.
```
ivanopagano: Should this entry actually show what kind of standard notation is used to define the formats?
Possibly improve for another doc iteration

anonymoussprocket: another iteration, the conseil docs need a deeper update

docs/README.md
```diff
@@ -59,11 +59,40 @@ Both `conseil` and `lorre` write to `syslog`. The logs are verbose enough to det
...
+- `display-order`: Hints to a UI as to the order in which the fields should appear by default.
```
ivanopagano: how is the order denoted? is an number where lower means higher priotity?

docs/README.md
```diff
@@ -59,11 +59,40 @@ Both `conseil` and `lorre` write to `syslog`. The logs are verbose enough to det
...
+- `display-priority`: Hints to the UI the relative priority of the data, allowing lower priority information that could be removed due to space constraints while keeping more relevant things in view.
```
ivanopagano: likewise as for order: is this a number and lower means more priority?

docs/README.md
```diff
@@ -59,11 +59,40 @@ Both `conseil` and `lorre` write to `syslog`. The logs are verbose enough to det
...
+- `data-type`: One of: DateTime, String, Int, Decimal, Boolean, Hash, AccountAddress, Currency.
+- `placeholder`: Intended to be used for sample values that might appear as placeholder text in search boxes.
+- `reference`
```
ivanopagano: missing entry description... btw I wonder how safe is to publish this... is it meant to later on expose joins in the rest api?

docs/README.md
ivanopagano: yes, it would be a "poetic licence", but all in all it's not needed, considering the examples follow trail

PR #500, opened 2019-08-02
499 - fix for the aggregation validation

### LINKED ISSUE #499: Aggregation does not work for currency type fields
With the update adding another type of the field we forgot to add `Currency` to the data types validation, for the types which could be used with sum/max/min/avg aggregations.

piotrkosecki:
fixes #499 

Removed unused ExecutionContext and used scalafmt formatting

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/michelson/renderer/MichelsonRenderer.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/michelson/renderer/MichelsonRenderer.scala
@@ -52,7 +52,7 @@ object MichelsonRenderer {
       self
         .map(_.render())
         .mkString(" ;\n")
-        .lines
+        .linesIterator
         .mkString("\n" + indent)
   }
 }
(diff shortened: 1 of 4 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/michelson/renderer/MichelsonRenderer.scala
```diff
@@ -52,7 +52,7 @@ object MichelsonRenderer {
       self
         .map(_.render())
         .mkString(" ;\n")
-        .lines
+        .linesIterator
```
ba205: Really interesting history here, linesIterator was deprecated because it had the same functionality at lines, then undeprecated because lines conflicted with other Scala functions. How does this impact aggregation?

piotrkosecki: it does not impact aggregation directly, but because of that travis did not want to build the project and run tests

PR #510, opened 2019-08-09
509 - fixing missing cardinality issue

### LINKED ISSUE #509: Missing cardinality information from the metadata endpoints
Metadata endpoints for entities and attributes were missing count/cardinality fields.

piotrkosecki:
resolves #509 

The issue was that metadata was stored in the values in MetadataService and they were never updated - update of the values in the cache is being done in background and but because of that it was omitted.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
@@ -61,8 +61,30 @@ class MetadataService(
   // fetches entities
   def getEntities(path: NetworkPath): Option[List[Entity]] = entities.get(path)
 
+  // gets current entities
+  def getCurrentEntities(path: NetworkPath): Future[Option[List[Entity]]] = {
+    platformDiscoveryOperations.getEntities.map { allEntities =>
+      if(exists(path))
+        Some(transformation.overrideEntities(path, allEntities, shouldLog = false))
+      else
+        None
+    }
+  }
+
   // fetches table attributes
   def getTableAttributes(path: EntityPath): Option[List[Attribute]] = attributes.get(path)
+
+  // fetches current attributes
+  def getCurrentTableAttributes(path: EntityPath): Future[Option[List[Attribute]]] = {
+    platformDiscoveryOperations.getTableAttributes(path.entity).map { maybeAttributes =>
+      maybeAttributes.flatMap { attributes =>
+        if(exists(path))
+          Some(transformation.overrideAttributes(path, attributes, shouldLog = false))
+        else
+          None
+      }
+    }
+  }
 
   // fetches table attributes without updating cache
   def getTableAttributesWithoutUpdatingCache(path: EntityPath): Future[Option[List[Attribute]]] =
--- a/src/test/scala/tech/cryptonomic/conseil/metadata/MetadataServiceTest.scala
+++ b/src/test/scala/tech/cryptonomic/conseil/metadata/MetadataServiceTest.scala
@@ -215,6 +215,39 @@ class MetadataServiceTest
         result shouldBe Some(List(Entity("entity", "entity-name", 0)))
       }
 
+    "fetch the list of supported entities with updated values" in {
+      // given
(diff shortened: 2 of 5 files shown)


src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
```diff
@@ -61,9 +61,31 @@ class MetadataService(
   // fetches entities
   def getEntities(path: NetworkPath): Option[List[Entity]] = entities.get(path)
 
+  // gets current entities
+  def getCurrentEntities(path: NetworkPath): Future[Option[List[Entity]]] = {
+    platformDiscoveryOperations.getEntities.map { allEntities =>
```
ivanopagano: I think we can remove a redundant future mapping if the path check is done right out
```scala
  if (exists(path))
    platformDiscoveryOperations.getEntities.map { allEntities =>
        Some(transformation.overrideEntities(path, allEntities, shouldLog = false))
    }
  else Future.successful(None)
```

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
```diff
@@ -61,9 +61,31 @@ class MetadataService(
...
+      else
+        None
+    }
+  }
+
   // fetches table attributes
   def getTableAttributes(path: EntityPath): Option[List[Attribute]] = attributes.get(path)
 
+  // fetches current attributes
+  def getCurrentTableAttributes(path: EntityPath): Future[Option[List[Attribute]]] = {
```
ivanopagano: Similarly we can check the path right out here, too. This also simplifies the map+flatMap nestings (I usually find out they highlight that some refactoring is possible).
Here we can push a little farther with an OptionT and single mapping
```scala
    if(exists(path)) {
      OptionT(platformDiscoveryOperations.getTableAttributes(path.entity)).map { attributes =>
        transformation.overrideAttributes(path, attributes, shouldLog = false)
      }.value
    } else Future.successful(None)
```
but it's not needed

src/test/scala/tech/cryptonomic/conseil/metadata/MetadataServiceTest.scala
```diff
@@ -215,6 +215,39 @@ class MetadataServiceTest
...
       }
 
+    "fetch the list of supported entities with updated values" in {
+      // given
+      platformDiscoveryOperations.addEntity(Entity("entity", "entity-name", 0))
+
+      val overwrittenConfiguration = Map(
+        "tezos" ->
+          PlatformConfiguration(
+            None,
```
ivanopagano: Named parameters might help understanding what's really going on here?

src/test/scala/tech/cryptonomic/conseil/metadata/MetadataServiceTest.scala
```diff
@@ -380,6 +413,39 @@ class MetadataServiceTest
...
       }
 
+    "filter out a hidden entity (by default) with updated entities" in {
+      // given
+      platformDiscoveryOperations.addEntity(Entity("entity", "entity-name", 0))
+
+      val overwrittenConfiguration = Map(
+        "tezos" ->
+          PlatformConfiguration(
+            None,
```
ivanopagano: Same as above, named parameters might help understanding?

src/test/scala/tech/cryptonomic/conseil/metadata/MetadataServiceTest.scala
```diff
@@ -473,6 +539,49 @@ class MetadataServiceTest
...
 
+    "fetch the list of supported attributes with updated values" in {
+      // given
+      platformDiscoveryOperations.addEntity(Entity("entity", "entity-name", 0))
+      platformDiscoveryOperations.addAttribute(Attribute("attribute", "attribute-name", Int, None, NonKey, "entity"))
+
+      val overwrittenConfiguration = Map(
+        "tezos" ->
+          PlatformConfiguration(
+            None,
```
ivanopagano: Same as before

PR #513, opened 2019-08-12
Added cycle to operations table via block data.

ba205:
Added cycle to operations table via block data, updated tests.


src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -40,6 +40,15 @@ object DatabaseConversions {
     case _ => None
   }
 
+  //Note, cycle 0 starts at the level 2 block
+  def extractCycle(block: Block) : Option[Int] = {
```
ivanopagano: Alternative using the existing partial function `discardGenesis` would be

```scala
def extractCycle(block: Block): Option[Int] =
  discardGenesis.lift(block.data.metadata) //this returns an Option[BlockHeaderMetadata]
    .map(_.level.cycle) //this is Option[Int]
```

Which might require no additional function but can be inlined at call site too, if willing

ivanopagano:
I see that PlatformDiscovery tests are missing the new column and still suffer from missing ballot values, seemingly from PR #482 ?

Please verify these so that Travis checks green before merging

PR #514, opened 2019-08-13
449 - make constants instructions rather that expressions

### LINKED ISSUE #449: Micheline/Michelson converter
Conseil is not processing literal parameters correctly resulting in things of this nature:

```
Unparsable code: {"int":"1"}
Unparsable code: [{"string":"2019-07-02T16:17:38Z"},{"string":"2019-07-02T16:17:08Z"}]
Unparsable code: {"string":"hello"}
```

dorians:
closes #449 

This PR changes the way constant types are parsed (as an instruction, not type).

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/michelson/dto/MichelsonInstruction.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/michelson/dto/MichelsonInstruction.scala
@@ -20,7 +20,9 @@ package tech.cryptonomic.conseil.tezos.michelson.dto
  *   | MichelsonSingleInstruction (with embedded MichelsonInstructionSequence as above)
  *   MichelsonInstructionSequence (with three instructions separated with ";": "DIP { ... }", "SWAP" and empty instruction)
  * */
-sealed trait MichelsonInstruction extends MichelsonElement
+sealed trait MichelsonInstruction extends MichelsonElement {
+  lazy val normalized: MichelsonInstruction = this
+}
 
 /* Class representing a simple Michelson instruction which can contains following expressions */
 case class MichelsonSingleInstruction(
@@ -32,8 +34,17 @@ case class MichelsonSingleInstruction(
 /* Class representing a sequence of Michelson instructions */
 case class MichelsonInstructionSequence(instructions: List[MichelsonInstruction] = List.empty)
     extends MichelsonInstruction {
-  lazy val normalized: MichelsonInstruction = if (instructions.isEmpty) MichelsonEmptyInstruction else this
+  override lazy val normalized: MichelsonInstruction = if (instructions.isEmpty) MichelsonEmptyInstruction else this
 }
+
+/* Class representing an int constant */
+case class MichelsonIntConstant(int: Long) extends MichelsonInstruction
+
+/* Class representing a string constant */
+case class MichelsonStringConstant(string: String) extends MichelsonInstruction
+
+/* Class representing a bytes constant */
+case class MichelsonBytesConstant(bytes: String) extends MichelsonInstruction
 
 /* Class representing an empty Michelson instruction */
 case object MichelsonEmptyInstruction extends MichelsonInstruction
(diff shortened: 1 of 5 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/michelson/dto/MichelsonInstruction.scala
```diff
@@ -32,8 +34,17 @@ case class MichelsonSingleInstruction(
 /* Class representing a sequence of Michelson instructions */
 case class MichelsonInstructionSequence(instructions: List[MichelsonInstruction] = List.empty)
     extends MichelsonInstruction {
-  lazy val normalized: MichelsonInstruction = if (instructions.isEmpty) MichelsonEmptyInstruction else this
+  override lazy val normalized: MichelsonInstruction = if (instructions.isEmpty) MichelsonEmptyInstruction else this
 }
 
+/* Class representing an int constant */
+case class MichelsonIntConstant(int: Long) extends MichelsonInstruction
```
anonymoussprocket: "Literal" instead of "Constant" would be better here.

src/main/scala/tech/cryptonomic/conseil/tezos/michelson/parser/JsonParser.scala
```diff
@@ -120,25 +98,51 @@ object JsonParser {
...
-        }
+        embeddedElements = args.getOrElse(List.empty).map(toMichelsonElement)
       )
   }
 
   case class JsonInstructionSequence(instructions: List[JsonInstruction]) extends JsonInstruction {
     override def toMichelsonInstruction = MichelsonInstructionSequence(instructions.map(_.toMichelsonInstruction))
   }
 
+  /*
```
anonymoussprocket: There are other types, `address`, `nat`, boolean, to name a few, do they need to be implemented separately?

src/main/scala/tech/cryptonomic/conseil/tezos/michelson/dto/MichelsonInstruction.scala
dorians: Probably yes, but in the docs, they use word "Constant".

src/main/scala/tech/cryptonomic/conseil/tezos/michelson/parser/JsonParser.scala
dorians: I'm not sure how the types are represented because it's not documented so I created those structures only for `int`, `string` and `bytes`. They were the only constructions for which parsing failed.

src/main/scala/tech/cryptonomic/conseil/tezos/michelson/dto/MichelsonInstruction.scala
```diff
@@ -32,8 +34,17 @@ case class MichelsonSingleInstruction(
 /* Class representing a sequence of Michelson instructions */
 case class MichelsonInstructionSequence(instructions: List[MichelsonInstruction] = List.empty)
```
ivanopagano: Might not be the best idea to spend time on this... but, since the parser will never call this with an empty list arg, why not changing the constructor to only accept non empty lists? varargs can be used to do that.
Would this avoid the need for a normalized form wholesale?

src/main/scala/tech/cryptonomic/conseil/tezos/michelson/parser/JsonParser.scala
```diff
@@ -38,6 +39,17 @@ object JsonParser {
     def toMichelsonExpression: MichelsonExpression
   }
 
+  type EmbeddedElement = Either[Either[JsonExpression, JsonInstruction], List[JsonInstruction]]
```
ivanopagano: Minor: Is there a way to use an ADT for this instead of "nested disjunctions"? Personally I find them hard to parse mentally and to reason with.

src/main/scala/tech/cryptonomic/conseil/tezos/michelson/parser/JsonParser.scala
```diff
@@ -120,25 +98,51 @@ object JsonParser {
...
+  }
+
+  /*
+   * Wrapper for bytes constant
+   *
+   * {"bytes": "0500"}
+   *
+   * */
+  case class JsonBytesConstant(bytes: String) extends JsonInstruction {
+    override def toMichelsonInstruction = MichelsonBytesConstant(bytes)
```
ivanopagano: Just to clarify, does this value encode byte-strings? That is, multiple byte-encoded values?

PR #518, opened 2019-08-23
Link balance_updates to operations

### LINKED ISSUE #471: Where possible the balance_updates table should include operation IDs
There is no user-friendly way to link `balance_updates` records to specific operations. Where possible, the table should include a value for an `operation_group_hash` column. 

This should apply to records whose source is either `operation` or `operation_result`.

ba205:
resolves #471


anonymoussprocket:
@ba205, the build is failing, please review

ivanopagano:
@vishakh @ba205 @anonymoussprocket please consider my note [here](https://github.com/Cryptonomic/Conseil/issues/471#issuecomment-525271046)

doc/conseil.sql
```diff
@@ -80,7 +80,8 @@ CREATE TABLE public.balance_updates (
     change numeric NOT NULL,
     level numeric,
     delegate character varying,
-    category character varying
+    category character varying,
+    operation_group_hash character varying
```
ivanopagano: Is this actually needed? see [this note](https://github.com/Cryptonomic/Conseil/issues/471#issuecomment-525271046)

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
```diff
@@ -225,7 +225,7 @@ class TezosDatabaseOperationsTest
             val generatedUpdateRows =
             operation
               .convertToA[List, Tables.BalanceUpdatesRow]
-              .map(_.copy(sourceId = Some(opRow.operationId)))
+              .map(_.copy(sourceId = Some(opRow.operationId), operationGroupHash = Some(opRow.operationGroupHash)))
```
ivanopagano: Needed?

doc/conseil.sql
ba205: So the issue is I wanted a quick fix, and as far as I saw in the table in dev, source_hash only works for sources that are blocks. When I tried to do a similar approach to place operation hashes in source_hash, I ended up somehow not capturing block sources, so I chose the easier way out this time.

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
ba205: See above. This is the quick fix that works, in running Lorre and in tests

ba205:
> I need a feedback regarding the necessity to reference the operation group from balance updates.
> 
> Additionally: is it correct that we don't test the platform discovery attributes for balance updates? Do we expose that table to the rest client?

@piotrkosecki @anonymoussprocket Do you guys know how balance_updates is exposed?

ba205:
@anonymoussprocket @piotrkosecki @ivanopagano I will make a new pr, as I meant to merge this branch into another, not directly into master. I'll keep this open until questions have been resolved however.

doc/conseil.sql
ivanopagano: I'm not sure if I'm missing anything, but I was trying to point out that, for balance updates referring to "operation" or "operation_result" in the `balance_updates.source` column, the referred operation can be found by reading the `operation_id` written in the `balance_updates.source_id` column.

I see no need of the operation group hash, you can simply get the operations row and read the operation group hash from that...

ba205: True, but we're leaning towards a more normalized data model with as few joins as possible, iirc. @anonymoussprocket thoughts?

ba205: I'm pretty sure this has to do with ease of use for Arronax specifically. Correct me if I'm wrong.

ba205:
> @anonymoussprocket @piotrkosecki @ivanopagano I will make a new pr, as I meant to merge this branch into another, not directly into master. I'll keep this open until questions have been resolved however.

Update, I didn't realize the other branch was already merged and deleted, this branch is fine

piotrkosecki:
> > I need a feedback regarding the necessity to reference the operation group from balance updates.
> > Additionally: is it correct that we don't test the platform discovery attributes for balance updates? Do we expose that table to the rest client?
> 
> @piotrkosecki @anonymoussprocket Do you guys know how balance_updates is exposed?

it should be exposed in the rest API in config files: https://github.com/Cryptonomic/Conseil/blob/master/src/main/resources/metadata/tezos.alphanet.conf
https://github.com/Cryptonomic/Conseil/blob/master/src/main/resources/metadata/tezos.mainnet.conf
if it's not mentioned in those config files it won't be visible through API

PR #522, opened 2019-08-27
bugfix: Restore the foreign key from operations to operation-groups

ivanopagano:
Previous PR #504 previously removed the key to handle the fact that op-groups need a composite key to account for forks in the chain.
The slick definitions needed a fresh re-generation which broke the client code.
This PR fix the situation re-introducint the FK referencing the now-composite key on groups.

Addtionally:
 * reformat the codebase
 * regenerate the slick definitions using the updated FK
 * fix a broken test on platform discovery
 * remove logging on tests: you can be enabled it again as needed, by changing the default
   level on the `logback-test.xml`


dorians:
Looks good but please try to do reformat in separate PR :)

Issue #454, opened 2019-07-05
Add block level and block hash indexes to rolls table


vishakh:
The `rolls` table has no indexes defined on it. As new records are appended to the table per block, the data has become impossible to query. We should therefore add indexes on the columns `block_id` and `block_level` for the table.

ivanopagano:
Was this closed as part of #536 ?

@ba205 @anonymoussprocket

vishakh:
RIP rolls table.

PR #524, opened 2019-08-28
Add indexes for rolls on both block-id and block-level columns

ivanopagano:
resolves: #454 

Add indexing on Rolls


ba205:
Edited https://github.com/Cryptonomic/Conseil/pull/531 to include this

PR #527, opened 2019-08-28
497 - improve fees dataset

### LINKED ISSUE #497: Improve fees dataset
Store level and cycle in the `fees` table.

piotrkosecki:
resolves #497 

### Database changes:
In table `fees` two new columns were added:
```
    cycle integer,
    level integer
```

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
@@ -39,13 +39,7 @@ object DatabaseConversions {
 
   implicit val averageFeesToFeeRow = new Conversion[Id, AverageFees, Tables.FeesRow] {
     override def convert(from: AverageFees) =
-      Tables.FeesRow(
-        low = from.low,
-        medium = from.medium,
-        high = from.high,
-        timestamp = from.timestamp,
-        kind = from.kind
-      )
+      from.into[Tables.FeesRow].transform
   }
 
   implicit val blockAccountsToAccountRows =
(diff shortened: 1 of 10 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -39,13 +39,7 @@ object DatabaseConversions {
...
   implicit val averageFeesToFeeRow = new Conversion[Id, AverageFees, Tables.FeesRow] {
     override def convert(from: AverageFees) =
-      Tables.FeesRow(
-        low = from.low,
-        medium = from.medium,
-        high = from.high,
-        timestamp = from.timestamp,
-        kind = from.kind
-      )
+      from.into[Tables.FeesRow].transform
```
ivanopagano: ❤️

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
```diff
@@ -953,7 +953,9 @@ class TezosDatabaseOperationsTest
           medium = mu,
           high = mu + sigma,
           timestamp = latest,
-          kind = ops.head.kind
+          kind = ops.head.kind,
+          cycle = None,
+          level = Some(0)
```
ivanopagano: Are we testing here that actual latest value is the one used in the average result?

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
dorians: good one!

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
piotrkosecki: no, we're not, those values were generated, I'm not why they are not `Nones`, as they are optional in the table

ivanopagano:
Now master has the sql schema synced with the slick models, you can merge it to get a consistent story of the db

PR #528, opened 2019-08-28
507 - Parametrize Tezos chains

### LINKED ISSUE #507: Parametrize Tezos chains
The Tezos network undergoes protocol upgrades as part of its governance period. Once the network votes in favour of a protocol amendment proposal, the corresponding binary is run in the background by the Tezos network on a test chain. 

We already run queries of this form to query the active (main) chain:

`https://tezos-prod.cryptonomic-infra.tech:443/chains/main/blocks/head`

If we want to instead query the test chain, we modify the query slightly:

`https://tezos-prod.cryptonomic-infra.tech:443/chains/test/blocks/head`

In order to properly test our end to end stack against imminent protocol environments, we would like to run a separate `staging` environment where all data is fetched from the `test` chain and all transactions are sent to the `test` chain as well.

For Conseil, this means the chain name must be parametrized everywhere. The configuration file should accept a new item called `platforms.{platform}.{network}.node.chain`, for example `platforms.tezos.mainnet-test.node.chain: test`. This value should be used by Lorre for all its queries to the Tezos node.

dorians:
closes #507

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/io/MainOutputs.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/io/MainOutputs.scala
@@ -165,8 +165,8 @@ object MainOutputs {
 
   /* custom display of each configuration type */
   private val showPlatformConfiguration: PartialFunction[PlatformConfiguration, String] = {
-    case TezosConfiguration(_, TezosNodeConfiguration(host, port, protocol, prefix)) =>
-      s"node $protocol://$host:$port/$prefix"
+    case TezosConfiguration(_, TezosNodeConfiguration(host, port, protocol, prefix, chainEnv)) =>
+      s"node $protocol://$host:$port/$prefix/$chainEnv"
     case _ =>
       "a non-descript platform configuration"
   }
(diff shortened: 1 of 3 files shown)


src/main/scala/tech/cryptonomic/conseil/io/MainOutputs.scala
```diff
@@ -165,8 +165,8 @@ object MainOutputs {
 
   /* custom display of each configuration type */
   private val showPlatformConfiguration: PartialFunction[PlatformConfiguration, String] = {
-    case TezosConfiguration(_, TezosNodeConfiguration(host, port, protocol, prefix)) =>
-      s"node $protocol://$host:$port/$prefix"
+    case TezosConfiguration(_, TezosNodeConfiguration(host, port, protocol, prefix, chainEnv)) =>
+      s"node $protocol://$host:$port/$prefix/$chainEnv"
```
ivanopagano: shouldn't we put the `/chains/` path fragment here, too, for consistency?

dorians: It's only for debugging purpose so I would leave it as short as possible.

PR #529, opened 2019-08-28
Proposal Metadata Added

### LINKED ISSUE #453: Capture information for proposals transactions
A typical proposals operation looks like this when returned by the Tezos RPC API:

```json
{
            "protocol": "PsddFKi32cMJ2qPjf43Qv5GDWLDPZb3T3bF6fLKiF5HtvHNU7aP",
            "chain_id": "NetXdQprcVkpaWU",
            "hash": "opS9cGJ51jCcLwkofdnQbwLnKNoqBzuQMfoMXPviS99kszQHiRQ",
            "branch": "BMchFBNHU3BA8AtAVCwyRA9SpWNSrbtzA8K6WNUNpVxxNcApHqh",
            "contents": [
                {
                    "kind": "proposals",
                    "source": "tz1MecudVJnFZN5FSrriu8ULz2d6dDTR7KaM",
                    "period": 10,
                    "proposals": [
                        "Psd1ynUBhMZAeajwcZJAeq5NrxorM6UCU4GJqxZ7Bx2e9vUWB6z"
                    ],
                    "metadata": {}
                }
            ],
            "signature": "sigoEwr6zeS28sufRgBznqAjf2jnwPxYw3vG3L5xVSYdJgJoACxVU9n4BrUjU8QJNdB9w7c4gxLYmaFzVJgdRD4gB4HroviY"
        }
```

We are currently not capturing any of these fields. Lorre should start picking them up.

ba205:
Closes #453. Merge plan is to move this into balance updates branch, and merge that to endorsement branch, and then to master.


src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -331,12 +332,29 @@ object DatabaseConversions {
...
+    case (block, groupHash, Proposals(source, period, proposals)) =>
+      Tables.OperationsRow(
+        operationId = 0,
+        operationGroupHash = groupHash.value,
+        kind = "proposals",
+        blockHash = block.data.hash.value,
+        blockLevel = block.data.header.level,
+        timestamp = toSql(block.data.header.timestamp),
+        internal = false,
+        proposal = proposals.map(_.toString()),
```
ivanopagano: wouldn't it make more sense to call this field and the related db-column as `proposals`?

src/test/scala/tech/cryptonomic/conseil/tezos/DatabaseConversionsTest.scala
```diff
@@ -776,14 +776,17 @@ class DatabaseConversionsTest
...
 
         converted.operationId shouldBe 0
         converted.operationGroupHash shouldBe groupHash.value
         converted.blockHash shouldBe block.data.hash.value
         converted.blockLevel shouldBe block.data.header.level
         converted.timestamp shouldBe Timestamp.from(block.data.header.timestamp.toInstant)
         converted.kind shouldBe "proposals"
+        converted.source shouldBe Some(ContractId("tz1VceyYUpq1gk5dtp6jXQRtCtY8hm5DKt72"))
+        //converted.period shouldBe Some(10)
+        converted.proposal shouldBe Some("[\"Psd1ynUBhMZAeajwcZJAeq5NrxorM6UCU4GJqxZ7Bx2e9vUWB6z\"]")
```
ivanopagano: I would suggest using triple-quotes for raw strings (so we won't need to escape quotes)
`"""["hashcode"]"""`

src/test/scala/tech/cryptonomic/conseil/tezos/DatabaseConversionsTest.scala
```diff
@@ -776,14 +776,17 @@ class DatabaseConversionsTest
...
+        val converted = (block, groupHash, sampleProposals: Operation).convertTo[Tables.OperationsRow]
 
         converted.operationId shouldBe 0
         converted.operationGroupHash shouldBe groupHash.value
         converted.blockHash shouldBe block.data.hash.value
         converted.blockLevel shouldBe block.data.header.level
         converted.timestamp shouldBe Timestamp.from(block.data.header.timestamp.toInstant)
         converted.kind shouldBe "proposals"
+        converted.source shouldBe Some(ContractId("tz1VceyYUpq1gk5dtp6jXQRtCtY8hm5DKt72"))
+        //converted.period shouldBe Some(10)
```
ivanopagano: what about this?

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
ivanopagano: I would also suggest using our custom-made `concatenateToString` or `toCommaSeparated` methods, instead of the default `list.toString`.
They're in scope here.

This will give us more control and visibility on the actual output format, instead of relying on the scala libraries defaults.

src/test/scala/tech/cryptonomic/conseil/tezos/DatabaseConversionsTest.scala
```diff
@@ -776,14 +776,17 @@ class DatabaseConversionsTest
...
+        converted.proposal shouldBe Some("[\"Psd1ynUBhMZAeajwcZJAeq5NrxorM6UCU4GJqxZ7Bx2e9vUWB6z\"]")
 
         forAll(
```
ivanopagano: As can be seen here, every test on operation conversions additionally takes care that every field in the row that is unused by a specific operation kind, is verified to be empty.

This means that newly introduced fields (like `period`) will need to be added to this "empty field list" in the test, for every operation type that doesn't actually use the field

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
ba205: proposal was already a table in the db and used by ballots as well. Does it make sense to have a "proposal" and "proposals" field? and alrighty on the concatenateToString idea

src/test/scala/tech/cryptonomic/conseil/tezos/DatabaseConversionsTest.scala
ba205: Good idea!

src/test/scala/tech/cryptonomic/conseil/tezos/DatabaseConversionsTest.scala
ba205: Yeah, I think uncommenting it out would solve the problem *sweatdrop my bad.

ivanopagano: Did you reply to the wrong question?

PR #537, opened 2019-09-02
538: Test/create regression basic suite

### LINKED ISSUE #538: Integration tests
Create a suite of integration tests to run separately from the build process for regression testing ahead of any deployments.

ivanopagano:
Closes #538 

This PR builds a basic tool for testing the results of running Lorre and then Conseil against a fixed range of blocks, to verify no regression is introduced by updates in the code.

The implementation adds a new `it` (integration tests) module (in addition to main and test) where this code can be eventually converted into a full integration-test suite to automate in the CI environment.

The current implementation only tests the current endpoints
- info
- head block data

### Usage
The tests need be run manually in this version.
- You have to setup a running postgres database instance with the _correct schema_ and _no data_ in it
- You have to launch the sbt console with the command: `sbt it:console`
- Then you have to import and run the endpoint checking class with the following commands
```bash
> import tech.cryptonomic.conseil.client.DataEndpointsClientProbe
> DataEndpointsClientProbe.runRegressionSuite(Some("prodnet")).unsafeRunSync
# Don't pass any parameter if you already loaded the data with Lorre on a previous check run
```

This will load the first 10k blocks in Lorre based on "prodnet" data (this step will be skipped if you call `runRegressionSuite()` with no arguments.
When the blocks are loaded a conseil instance will be run, and the test will check the endpoints, taking care to clean up the process after that.
A basic report will be printed to the console with the result of the regression test.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,7 +5,6 @@ lazy val conseil = (project in file("."))
   .configs(IntegrationTest)
   .settings(
     Defaults.itSettings
-    // other settings here
   )
 
 val akkaVersion = "2.5.21"
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,13 @@
 name := "Conseil"
 scalaVersion := "2.12.8"
 
+lazy val conseil = (project in file("."))
+  .configs(IntegrationTest)
+  .settings(
+    Defaults.itSettings
+    // other settings here
+  )
+
 val akkaVersion = "2.5.21"
 val akkaHttpVersion = "10.1.8"
 val akkaHttpJsonVersion = "1.25.2"
@@ -9,6 +16,7 @@ val catsVersion = "1.6.0"
 val monocleVersion = "1.5.1-cats"
 val endpointsVersion = "0.9.0"
 val circeVersion = "0.11.1"
+val http4sVersion = "0.20.10"
 
 scapegoatVersion in ThisBuild := "1.3.8"
 parallelExecution in Test := false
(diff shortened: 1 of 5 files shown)


dorians:
How about adding it to travis build?

src/it/scala/tech/cryptonomic/conseil/client/EndpointClient.scala
```diff
@@ -0,0 +1,169 @@
...
+        _ <- timer.sleep(30.seconds)
+      } yield proc
+
+    def usingConseil[A](syncNetwork: Option[String] = None)(testBlock: => IO[A]) =
+      startConseil(syncNetwork).bracket(
+        use = _ => testBlock
+      )(
+        release = _ =>
+          for {
+            pid <- IO(grepConseilPID())
```
dorians: why wrapping each line in `IO`?

src/it/scala/tech/cryptonomic/conseil/client/EndpointClient.scala
```diff
@@ -0,0 +1,169 @@
...
+    private def stopPid(pid: String) = Seq("kill", "-2", pid).!
+
+    private def syncData(network: String) = IO((runLorre(network).!).ensuring(_ == 0, "lorre failed to load correctly"))
+
+    private def startConseil(syncNetwork: Option[String]): IO[Process] =
+      for {
+        _ <- if (syncNetwork.nonEmpty) syncData(syncNetwork.get) else IO(0)
+        proc <- IO(runConseil.run())
+        _ <- IO(println("waiting for conseil to start"))
+        _ <- timer.sleep(30.seconds)
```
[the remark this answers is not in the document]
ivanopagano: I can't think of anything right out of my mind, unless I capture the process' output and match on some given line, which doesn't look great.

Please consider that the purpose is to catch plain errors for quite simple cases, and it's not supposed to be run constantly during the daily development cycle.

src/it/scala/tech/cryptonomic/conseil/client/EndpointClient.scala
ivanopagano: For consistency of execution? Otherwise side-effects might come out intertwined incorrectly. It's essentially sequencing explicitly the operations.

build.sbt
```diff
@@ -1,6 +1,13 @@
 name := "Conseil"
 scalaVersion := "2.12.8"
 
+lazy val conseil = (project in file("."))
+  .configs(IntegrationTest)
+  .settings(
+    Defaults.itSettings
+    // other settings here
```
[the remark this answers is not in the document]
ivanopagano: Thanks

ivanopagano:
> How about adding it to travis build?

it's still not automated. It needs the appropriate hooks into sbt to be launched from the shell, without running the sbt console manually.

I'd leave this as a later step.

src/it/scala/tech/cryptonomic/conseil/client/EndpointClient.scala
```diff
@@ -0,0 +1,169 @@
...
+import org.http4s.client.dsl.io._
+import org.http4s.Uri
+import io.circe.{Json, parser}
+import cats.effect.IO
+import cats.syntax.apply._
+
+/** Currently can be used to test any conseil instance that loaded blocks levels 1 to 1000
+  * against predefined expectations on the responses
+  */
+object DataEndpointsClientProbe {
```
ba205: Why not name this EndpointClient?

ivanopagano: The "Data" prefix is not necessarily correct, because other endpoints apart from `v2/data` are used for the test.

EndpointClient is too generic, it doesn't state what's his duty. It's too easy to take this for a client that gives you back data from the endpoint, like non-testing code.

EndpointsVerificationClient? EndpointsTester?

Open to suggestions

PR #544, opened 2019-09-11
License update

anonymoussprocket:
taken from https://www.apache.org/licenses/LICENSE-2.0.txt at the time of the checkin


ivanopagano:
Just to make a remark, I'm sorry to admit that we didn't check licence compatibilities issues when using OSS dependencies for the project, maybe you want me to collect the list of libraries and licences involved?

PR #548, opened 2019-09-11
Cleaning/520 move ballot proposal operation data

### LINKED ISSUE #520: Move ballot, proposal operation data
`ballot`s and `proposal`s are operation types, they should be recoded in the operations table.

Once this is done, existing ballot and proposal tables should be removed.

dorians:
closes #520 

What's inside:
* generating schema (schema was out of date)
* removing ballots and proposals tables
* cleaning code a little bit (cleaning compiler warnings)


doc/conseil.sql
```diff
@@ -565,14 +522,6 @@ ALTER TABLE ONLY public.operations
 ALTER TABLE ONLY public.operations
     ADD CONSTRAINT fk_opgroups FOREIGN KEY (operation_group_hash, block_hash) REFERENCES public.operation_groups(hash, block_id);
 
---
--- Name: proposals proposal_block_id_fkey; Type: FK CONSTRAINT; Schema: public; Owner: -
---
-
-ALTER TABLE ONLY public.proposals
-    ADD CONSTRAINT proposal_block_id_fkey FOREIGN KEY (block_id) REFERENCES public.blocks(hash);
```
ivanopagano: There was no FK from ballots to blocks?

src/test/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperationsTest.scala
```diff
@@ -187,6 +187,9 @@ class TezosPlatformDiscoveryOperationsTest
         sut.getTableAttributes("operations").futureValue shouldBe
           Some(
             List(
+              Attribute("branch", "Branch", DataType.String, None, KeyType.NonKey, "operations"),
```
ivanopagano: Possibly another PR

What do you think about using an order-independent verification method like doing
```scala
sut.getTableAttributes(...).futureValue.value should contain theSameElementsAs List(...)
```
, so that random shuffling of the attributes output based on slick's definition will not break the test anymore?

doc/conseil.sql
dorians: Nope

PR #549, opened 2019-09-11
545: P005 protocol fields update

### LINKED ISSUE #545: P005 message field changes
- [ ] `source` should be changed to PublicKeyHash type, as tz accounts are not contracts and currently source is of type ContractId
- [ ] mark the removed `origination` fields as nullable: `manager_pubkey`, `spendable`, `delegable`

ivanopagano:
Closes #545 

Take actions to make conseil data compatible with the changes in protocol, regarding fields.

I was able to complete sync of our current zeronet node in ~3.5 hrs with no error relating to the changes introduced


ivanopagano:
Opening as Draft as I want to locally try to test it against the zeronet node for compatibility bugs

ba205:
@ivanopagano I see that you added "slot" to operations for the endorsement change. We have "slots" already, is it better to keep them separate? For comparison, we use "proposals" currently to reference something in ballot and proposals operations

ivanopagano:
> @ivanopagano I see that you added "slot" to operations for the endorsement change. We have "slots" already, is it better to keep them separate? For comparison, we use "proposals" currently to reference something in ballot and proposals operations

I only followed what was mentioned in the protocol update, since I couldn't tell if the "slot" field would be the same, if "slots" could be removed, or what.
I choose the most backward compatible and safest choice.

PR #554, opened 2019-09-20
550 - save baking and endorsing rights

### LINKED ISSUE #550: Save baking and endorsement rights
Each block in the Tezos blockchain is baked by a single baker and endorsed by multiple endorsers. The rights to bake and endorse are known in advance and can be compared to the actual bakers and endorsers of the block to generate metrics about network health. We would therefore like to save the baking rights and endorsement rights of blocks along with the rest of the blocks' data.
 
The route `/chains/main/blocks/head/helpers/baking_rights` gives the baking rights for the current block.
 
Here is an example of the returned values:
 
```json
[
    {
        "level": 603946,
        "delegate": "tz3bTdwZinP8U1JmSweNzVKhmwafqWmFWRfk",
        "priority": 0,
        "estimated_time": "2019-09-11T23:51:26Z"
    },
    {
        "level": 603946,
        "delegate": "tz1dwu9aYb7CRNq4Y2zAjipdjFuSVKhHS8vA",
        "priority": 1,
        "estimated_time": "2019-09-11T23:52:41Z"
    }
]
```
 
This data should be saved in a table called `baking_rights` with `(level, delegate)` as the primary key and an index on `level`. Each record from the JSON string corresponds to one record in the database table.
 
The route `/chains/main/blocks/head/helpers/endorsing_rights` gives the endorsement rights for the current block.
 
Here is an example of the returned values:
 
```json
[
    {
        "level": 603950,
        "delegate": "tz3bvNMQ95vfAYtG8193ymshqjSvmxiCUuR5",
        "slots": [
            26,
            24
        ],
        "estimated_time": "2019-09-11T23:55:26Z"
    },
    {
        "level": 603950,
        "delegate": "tz3bTdwZinP8U1JmSweNzVKhmwafqWmFWRfk",
        "slots": [
            25
        ],
        "estimated_time": "2019-09-11T23:55:26Z"
    }
]
```
 
This data should be saved in a table called `endorsing_rights` with `(level, delegate, slot)` as the primary key and an index on `level`. Each record from the JSON string can correspond to multiple records in the database table since the slots array should be flattened into multiple entries.

Both tables should also capture the block hash.

piotrkosecki:
resolves #550 

## database changes:
```
CREATE TABLE public.baking_rights (
    level integer NOT NULL,
    delegate character varying NOT NULL,
    priority integer NOT NULL,
    estimated_time timestamp without time zone NOT NULL,
    PRIMARY KEY(level, delegate)
);

CREATE INDEX baking_rights_level_idx ON public.baking_rights (level);

CREATE TABLE public.endorsing_rights (
    level integer NOT NULL,
    delegate character varying NOT NULL,
    slots character varying NOT NULL,
    estimated_time timestamp without time zone NOT NULL,
    PRIMARY KEY(level, delegate, slots)
);


CREATE INDEX endorsing_rights_level_idx ON public.endorsing_rights (level);

--
```


doc/conseil.sql
```diff
@@ -308,6 +308,29 @@ CREATE TABLE public.rolls (
...
+    PRIMARY KEY(level, delegate)
+);
+
+CREATE INDEX baking_rights_level_idx ON public.baking_rights (level);
+
+CREATE TABLE public.endorsing_rights (
+    block_hash character varying NOT NULL,
+    level integer NOT NULL,
+    delegate character varying NOT NULL,
+    slots character varying NOT NULL,
```
ivanopagano: Isn't this supposed to be a single "slot", from the ticket description?
Each slot his own row

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -207,6 +209,17 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
 
     }
 
+    def processBakingAndEndorsingRights(fetchingResults: tezosNodeOperator.BlockFetchingResults): Future[List[Unit]] = {
```
ivanopagano: I can't see a reason to have a `List[Unit]` vs. `Unit`, maybe it would be a better return type?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -221,8 +234,12 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
         Source
           .fromIterator(() => pages)
           .mapAsync[tezosNodeOperator.BlockFetchingResults](1)(identity)
-          .mapAsync(1) {
-            processBlocksPage(_) flatTap (_ => processTezosAccountsCheckpoint >> processTezosDelegatesCheckpoint)
+          .mapAsync(1) { fetchingResults =>
+             processBlocksPage(fetchingResults).flatMap { result =>
```
ivanopagano: If we don't need the "intermediate results" of the flatMap, _flatTap_ is the good choice.

Here what needs be reported is the `processBlocksPage` result, hence we can flatTap over backing and endorsing rights, as well as checkpoints, removing the need to keep track of the intermediate `result` to map it back at the end (flatTap is exactly doing that for us).

ivanopagano: What if the backing/endorse rights fails to correctly process?

We actually have a similar issue with voting data (which is currently being partially removed, luckily).

The point is mainly about transactional guarantees of consistency for the whole processing. Once blocks (and tightly-related data) has been atomically stored, any failure in subsequent processing of corollary data might fail, and Lorre will not pick it up at next retried run.
That's the whole reason for checkpointing accounts/delegates ids, so that any failure will not lose that information anyway, which will be picked up by those two instructions right here at line 240.

I'm only pointing this out here by chance, this is a more general issue of consistency that will continue to threaten the lorre process as we add more and more data, food for future thought.
I'm hereby saying that we need to (re-)think the way data is stored, to manage a safe way to not risk losing any data due to complex elaboration. One approach is to do multi-stage processing, saving data as soon and as raw as possible, and then parse and process it in an immediate later stage. Then when can clean the raw data only if it's been confirmed as processed.

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -613,4 +615,27 @@ object DatabaseConversions {
...
+        .withFieldRenamed(_.estimated_time, _.estimatedTime)
+        .transform
+    }
+  }
+
+  implicit val endorsingRightsToRows = new Conversion[Id, (BlockHash, EndorsingRights), Tables.EndorsingRightsRow] {
+    override def convert(from: (BlockHash, EndorsingRights)): tezos.Tables.EndorsingRightsRow = {
+      val (blockHash, endorsingRights) = from
+      endorsingRights.into[Tables.EndorsingRightsRow]
+        .withFieldConst(_.slots, endorsingRights.slots.mkString(","))
```
ivanopagano: Please re-use existing private methods to cleanly and consistently summarize collections to a string.
There's a method to make a csv, but consistently spacing values.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -114,6 +114,27 @@ class TezosNodeOperator(val node: TezosRPCInterface, val network: String, batchC
...
 
+
+  /**
+    * Fetches baking rights for given block
+    * @param blockHash  Hash of given block
+    * @return           Baking rights
+    */
+  def getBakingRightsForBlock(blockHash: BlockHash): Future[List[BakingRights]] =
+    node
+    .runAsyncGetQuery(network, s"blocks/${blockHash.value}/helpers/baking_rights")
```
ivanopagano: formatting

doc/conseil.sql
piotrkosecki: you are right, I misunderstood it, I'm going to change that, good catch

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: ok, used flatTap, also, there is a lot of places it could fail and we could not retrieve data elsewhere. For example I was testing my changes in the different branch few days ago and I stopped Lorre in some unfortunate moment that did not allow to continue fetching so the easiest solution was to remove DB and start again.

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
piotrkosecki: removed with the other change

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
piotrkosecki: reformatted

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: yes, it is, changed

doc/conseil.sql
piotrkosecki: changed

src/main/scala/tech/cryptonomic/conseil/tezos/Tables.scala
```diff
@@ -21,6 +21,7 @@ trait Tables {
     AccountsCheckpoint.schema,
     BakingRights.schema,
     BalanceUpdates.schema,
+    Ballots.schema,
```
ivanopagano: Are we sure this is wanted and not some unexpected merge error with master?
I think one of the rollup changes included the removal of these tables (ballots and proposals)

#548 for reference

Please check the sql schema and try to regenerate slick's definitions to make sure we're not re-introducing things that shouldn't be there.

piotrkosecki: clumsy me, I forgot to recreate container before generating table, good catch

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: In general this shouldn't happen. The system should be able to stop in-between blocks' processing and never lose data, apart from:
- voting information
- rights to endorsements/baking (as of this PR)

piotrkosecki: If I encounter it again I'll msg you. If I remember correctly the problem was with violating primary key in some table - Lorre crashed.

PR #560, opened 2019-09-24
540 - DB namespaces

### LINKED ISSUE #540: DB namespaces
Create a namespace for the Tezos tables that Lorre & Conseil use.

piotrkosecki:
resolves #540 

## Schema changes
It requires schema changes, as title of this PR states, we need to migrate schema from `public` to `tezos`


src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -74,7 +74,7 @@ class TezosPlatformDiscoveryOperations(
   /** Pre-caching attributes without cardinality */
   private def preCacheAttributes: DBIO[AttributesCache] = {
     val result = for {
-      tables <- MTable.getTables(Some(""), Some("public"), Some(""), Some(Seq("TABLE")))
+      tables <- MTable.getTables(Some(""), Some(""), Some(""), Some(Seq("TABLE")))
```
ivanopagano: I would expect this to be "public" -> "tezos", how comes it's now empty?

src/test/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperationsTest.scala
```diff
@@ -184,8 +184,7 @@ class TezosPlatformDiscoveryOperationsTest
       }
 
       "return list of attributes of operations" in {
-        sut.getTableAttributes("operations").futureValue shouldBe
-          Some(
+        sut.getTableAttributes("operations").futureValue.get should contain allElementsOf
```
ivanopagano: I think `theSameElementsAs` would better fit the test.

`allElementsOf` means that the attributes might actually be _more than those listed_ in the expectations

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
piotrkosecki: you are right, changed

src/test/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperationsTest.scala
piotrkosecki: changed in all test cases and reformatted

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -182,7 +182,7 @@ class TezosPlatformDiscoveryOperations(
   /** Method querying slick metadata tables for entities */
   private def preCacheEntities: DBIO[EntitiesCache] = {
     val result = for {
-      tables <- MTable.getTables(Some(""), Some("public"), Some(""), Some(Seq("TABLE")))
+      tables <- MTable.getTables(Some(""), Some("tezos"), Some(""), Some(Seq("TABLE")))
```
bhaskarkishore: I am getting the feeling that, these should be parameterized via configuration.

ivanopagano: Indeed, but maybe we can put those changes with a PR that actually uses the parameterized database namespace everywhere.
Doing it now, would be an isolated change between many that will need be done, anyway.
WDYT?

bhaskarkishore: Fair enough

doc/conseil.sql
```diff
@@ -16,17 +16,21 @@ SET xmloption = content;
 SET client_min_messages = warning;
 SET row_security = off;
 
+CREATE SCHEMA IF NOT EXISTS tezos;
+ALTER ROLE conseiluser SET search_path TO tezos,public;
+ALTER DATABASE "conseil-local" SET search_path TO tezos,public;
```
ba205: Personal preference, I think "conseil" over "conseil-local" would suffice.

vishakh:
This PR need a pull from `master` in order to be merged.

doc/conseil.sql
piotrkosecki: `conseil-local` is a name of the db in the `docker-compose`, I could change it in the both places to `conseil`, WDYT?

PR #562, opened 2019-09-24
Protocol 005 updates

anonymoussprocket:


doc/conseil.sql
```diff
@@ -43,9 +43,9 @@ SET default_with_oids = false;
 CREATE TABLE public.accounts (
     account_id character varying NOT NULL,
     block_id character varying NOT NULL,
-    manager character varying NOT NULL,
-    spendable boolean NOT NULL,
-    delegate_setable boolean NOT NULL,
+    manager character varying,
```
vishakh: Were these all made in response to changes in the Tezos RPC schema?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -190,7 +190,7 @@ object TezosTypes {
       gas_limit: PositiveBigNumber,
       storage_limit: PositiveBigNumber,
       public_key: PublicKey,
-      source: ContractId,
+      source: PublicKeyHash,
```
vishakh: Please remind me again why we have two distinct types called `ContractId` and `PublicKeyHash`?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -264,38 +265,39 @@ object TezosTypes {
...
     ) extends InternalOperationResult
 
+    /* some fields are only kept for backward-compatibility, as noted*/
     case class Origination(
         kind: String,
-        source: ContractId,
+        source: PublicKeyHash,
         nonce: Int,
-        manager_pubkey: PublicKeyHash,
+        manager_pubkey: Option[PublicKeyHash], // retro-compat from protocol 5+
```
vishakh: Can you explain why we need to maintain retro compatibility?

doc/conseil.sql
vishakh: Due to my subsequent PR, I understand this bit better now.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
vishakh: I now understand you were trying to keep Lorre consistent with data from nodes running older protocols. Maybe this is still possible. I will investigate.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
ivanopagano: Semantics reason only. There's no real difference, only that at the time I wanted to differentiate the usage.

PR #566, opened 2019-09-27
553 - new configuration option to allow blank API keys

### LINKED ISSUE #553: New configuration option to allow blank API keys
Based on feedback in #552, add an option like `security.apiKeys.allowBlank` which would take a boolean value.

If it is set to false and `security.apiKeys.keys` is missing or empty, Conseil should write an error the log at startup and then shut down.

If it is set to true, regardless of what is in `security.apiKeys.keys`, Conseil should accept incoming HTTP requests without an `apiKey` header. If a header is present it should be validated against the list in `security.apiKeys.keys` according to the same rules that apply today.

dorians:
closes #553

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/config/Security.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/config/Security.scala
@@ -10,6 +10,7 @@ import cats.effect.IO
 import cats.effect.concurrent.Ref
 import com.typesafe.scalalogging.LazyLogging
 import de.heikoseeberger.akkahttpcirce.ErrorAccumulatingCirceSupport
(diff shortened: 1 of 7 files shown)


src/main/scala/tech/cryptonomic/conseil/config/Security.scala
```diff
@@ -43,20 +44,43 @@ object Security extends ErrorAccumulatingCirceSupport with LazyLogging {
   private def makeUri(ncc: NautilusCloudConfiguration): String =
     s"${ncc.host}:${ncc.port}/${ncc.path}"
 
+  final case class InvalidSecurityConfiguration(message: String) extends RuntimeException(message)
```
ivanopagano: Possibly: IllegalStateException?

PR #567, opened 2019-09-27
501 - Process operations to mark Accounts as revealed/activated

### LINKED ISSUE #501: Improve accounts dataset
Account revelation and activation flags should be added to the accounts table. They should default to false.

ivanopagano:
Closes #501 

A new executable target called Harpoon is available.
It runs a data processor that will mark accounts based on the operations it sees.
The process can be run autonomously from either Lorre/Conseil, but it needs to have a schema-compatible database with those.
It uses write access to the database, so a .conf file is needed to provide `lorre.db` entry with the usual credentials.

## Changes to database schema
Additional columns are added to the `account` table

```slq
activated boolean DEFAULT false NOT NULL
revealed boolean DEFAULT false NOT NULL
```


src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
```diff
@@ -0,0 +1,142 @@
...
+    /* Implement any dependency needed to run the processing. */
+    implicit val serviceApis = injectServices(tagging)
+
+    /* now we can call the processing stage */
+    val processCycle = (processedTopLevel: tagging.ProcessorInput) =>
+      tagging.process
+        .run(processedTopLevel)
+        .flatTap {
+          case (level, flagged) =>
+            logger.pureLog[IO](
```
kubukoz: Might want to use log4cats for nicer syntax

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
```diff
@@ -0,0 +1,142 @@
...
+                "This Harpoon processing cycle marked {} accounts, up to level {}.",
+                flagged.fold("an unknown number of")(String.valueOf),
+                level.fold("unspecified")(String.valueOf)
+              )
+            )
+        }
+        .handleErrorWith(
+          err =>
+            logger.pureLog[IO](_.error("Harpoon processing failed to complete", err)) *>
+                IO(processedTopLevel, 0) //we reset to the previous state as to retry next round
```
kubukoz: Avoid autotupling - `IO((processedTopLevel, 0))` should be used instead (if there are no side effects here, `IO.pure` can be used too)

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
```diff
@@ -0,0 +1,142 @@
...
+        )
+
+    //a wait period before running the process once more
+    val waitRound =
+      IO.shift *>
```
kubukoz: Why shift before logging?

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
```diff
@@ -0,0 +1,142 @@
...
+          logger.pureLog[IO](
+            _.info("Harpoon is now waiting for the next processing cycle")
+          ) *>
+          timer.sleep(conf.cycleSleep)
+
+    //combine the two parts, recursively calling itself, with an updated state
+    def program(level: tagging.ProcessorInput): IO[Unit] =
+      for {
+        out <- processCycle(level)
```
kubukoz: `(levelReached, flagCounts) <- processCycle(level)` should work if you add the better-monadic-for plugin.

kubukoz: Also, flagCounts seems to be unused.

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
```diff
@@ -0,0 +1,142 @@
...
+        (levelReached, flagCounts) = out
+        _ <- waitRound
+        _ <- program(levelReached)
+      } yield ()
+
+    /* we return the exit status into IO. The IOApp will run the actual process */
+    program(Option.empty)
+      .guarantee(
```
kubukoz: I'd totally move this to a Resource wrapping the db handle.

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
```diff
@@ -0,0 +1,142 @@
...
+        IO(dbHandle.close())
+      )
+      .as(ExitCode.Success)
+
+  }
+
+  /* Allows running db-operations as cats-effect IO actions */
+  private def runDbToIO[T] = IOUtil.runDbToIO[T](dbHandle.run _) _
```
kubukoz: I suggest annotating this with a return type.

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
```diff
@@ -0,0 +1,142 @@
...
+
+  /* Allows streaming db-results */
+  private def runDbToStream[R, T](action: StreamingDBIO[R, T]) =
+    IOUtil.publishStream(dbHandle.stream(action))
+
+  /* builds the service instance based on global operations available in the project */
+  private def injectServices(processor: AccountsTaggingProcess[IOState])(implicit ec: ExecutionContext) =
```
kubukoz: `implicit ec` is redundant due to having one in scope due to being in IOApp.

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
```diff
@@ -0,0 +1,142 @@
...
+    new processor.ServiceDependencies {
+
+      override def flagAsActive(accountId: TezosTypes.PublicKeyHash): IOState[Int] =
+        runDbToIO(TezosDatabaseOperations.flagAccount(accountId.value, AccountFlags.activated)).to[IOState]
```
kubukoz: I'd suggest avoiding usage of `to`, as it only works on IO (and requires `cats.effect.Effect` otherwise), and using `State.liftF(...)` instead.

kubukoz: Oh, you have that as `liftIOState` already. You can use that then.

src/main/scala/tech/cryptonomic/conseil/config/HarpoonAppConfig.scala
```diff
@@ -0,0 +1,7 @@
+package tech.cryptonomic.conseil.config
+
+import scala.concurrent.duration._
+
+case object HarpoonAppConfig {
```
kubukoz: Probably no need for `case` here.

src/main/scala/tech/cryptonomic/conseil/process/AccountsTaggingProcess.scala
```diff
@@ -0,0 +1,157 @@
...
+  val operationKinds = Set("activate_account", "reveal")
+
+}
+
+/** Scans the existing blocks stored by conseil for specific account-related operations
+  * i.e. Revelation, Activation.
+  * It then marks such accounts with the appropriate flag (e.g. revealed, activated).
+  * @param highWatermarkRef a pure reference that tracks the highest verified block level for flagging operations
+  */
+class AccountsTaggingProcess[F[_]: Sync] private ( //highWatermarkRef: Ref[F, Option[AccountsTaggingProcess.BlockLevel]])(
```
kubukoz: Dead comment

src/main/scala/tech/cryptonomic/conseil/process/AccountsTaggingProcess.scala
```diff
@@ -0,0 +1,157 @@
...
+          stored.map(_ max newLevel).orElse(newLevel.some)
+      }
+
+    /* find sorted operations and flag the related accounts as they come by, all the while keeping track of the
+     * new block level, as read from the operations
+     */
+    api
+      .readOperations(fromLevel)
+      .evalMap { implicit operation =>
+        (operation.kind match {
```
kubukoz: I'd extract this whole `operation.kind match` as a val

src/main/scala/tech/cryptonomic/conseil/process/AccountsTaggingProcess.scala
```diff
@@ -0,0 +1,157 @@
...
+      .evalTap {
+        //evenutually store the reference level, if needed
+        case (Some(flaggedCount), blockLevel) => updateIfHigher(blockLevel)
+        case (None, _) => Applicative[F].unit
+      }
+      .foldMap {
+        //extracts and sums all integer values (counting the actual updates)
+        case (anyFlag, _) => anyFlag.getOrElse(0)
+      }
+      .last
```
kubukoz: `.compile.last` is probably what you're looking for

src/main/scala/tech/cryptonomic/conseil/util/IOUtil.scala
```diff
@@ -0,0 +1,28 @@
...
+object IOUtil {
+
+  /** Transform any slick action to an IO result which will run the db operation when the IO is executed
+    * @param runner a way to execute the database action, i.e. use the database instance to invoke `db.run`
+    * @param action the db action to execute
+    * @param shifter a necessary context shift to handle threads correctly
+    * @tparam T the result of the operation
+    */
+  def runDbToIO[T](runner: DBIO[T] => Future[T])(action: DBIO[T])(implicit shifter: ContextShift[IO]): IO[T] =
+    IO.fromFuture(IO(runner(action))).guarantee(IO.shift)
```
kubukoz: This `guarantee` is redundant on CE 2.x, btw

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
```diff
@@ -0,0 +1,142 @@
...
+
+  /* This is the main entrypoint, the return type is wrapped in IO and provides an ExitCode for the process
+   * Described in cats-effect apis: https://typelevel.org/cats-effect/api/cats/effect/IOApp.html
+   */
+  override def run(args: List[String]): IO[ExitCode] = {
+
+    //we read custom conf, usually provided from the environment
+    val conf = HarpoonAppConfig
+
+    implicit val ctx = scala.concurrent.ExecutionContext.global
```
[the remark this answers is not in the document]
ivanopagano: Thanks, I thought I'd removed this actually

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
ivanopagano: Interesting, does it support lightbend loggers?
I've gone with a home-made very straightforward wrapper, I'll look into that

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
ivanopagano: Cool, mistake

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
ivanopagano: it's actually shift as first thing in the waiting sequence, it wasn't specifically for logging.

I thought to give more chance to switch context before each cycle. Is it a costly operation and possibly not worth doing here?

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
ivanopagano: I'm not too much into adding plugins if not strictly necessary, one thing at a time...

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
ivanopagano: I think I had to add it because something was requiring it and didn't pick anything from the IOApp.

I'll try removing it

src/main/scala/tech/cryptonomic/conseil/config/HarpoonAppConfig.scala
ivanopagano: typo?

src/main/scala/tech/cryptonomic/conseil/process/AccountsTaggingProcess.scala
ivanopagano: Undead stuff... I thought I'd removed this too

Nice one

src/main/scala/tech/cryptonomic/conseil/util/IOUtil.scala
ivanopagano: Does the `fromFuture` automatically shift context?

build.sbt
```diff
@@ -44,7 +46,11 @@ libraryDependencies ++= Seq(
   "com.fasterxml.jackson.module" %% "jackson-module-scala"          % "2.9.6",
   "com.chuusai"                  %% "shapeless"                     % "2.3.3",
   "org.typelevel"                %% "cats-core"                     % catsVersion,
+  "org.typelevel"                %% "cats-effect"                   % "1.3.1",
```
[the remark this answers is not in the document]
ivanopagano: Is it out and compatible, I'll try that, thanks

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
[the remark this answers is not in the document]
ivanopagano: Well, it seems that there's an implicit ContextShift[IO], but no ready-made ExecutionContext... 🦊

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
ivanopagano: I actually never considered that there's a difference between `IO.apply` and `IO.pure`, great insight

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
ivanopagano: Indeed, it didn't work out

src/main/scala/tech/cryptonomic/conseil/process/AccountsTaggingProcess.scala
ivanopagano: leftovers of previous trials... thanks a lot

ivanopagano:
Please don't merge: adding missing tests

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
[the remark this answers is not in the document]
kubukoz: Ohh, that's true. Sorry :P

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
kubukoz: It doesn't support them, afaik. You'd use string interpolation/concatenation to insert parameters.

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
kubukoz: It's probably not as bad as it sounds... but if you want to increase fairness it's a fair approach - worth documenting in a comment

src/main/scala/tech/cryptonomic/conseil/util/IOUtil.scala
kubukoz: Yes.

kubukoz: It also requires a ContextShift for that reason :)

build.sbt
```diff
@@ -44,7 +47,11 @@ libraryDependencies ++= Seq(
   "com.fasterxml.jackson.module" %% "jackson-module-scala"          % "2.9.6",
   "com.chuusai"                  %% "shapeless"                     % "2.3.3",
   "org.typelevel"                %% "cats-core"                     % catsVersion,
-  "org.typelevel"                %% "mouse"                         % "0.20",
+  "org.typelevel"                %% "cats-effect"                   % catsVersion,
```
kubukoz: I wouldn't use the same version val (cats core and effect can have different veresions)

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
```diff
@@ -0,0 +1,153 @@
...
+          }
+          .handleErrorWith(
+            err =>
+              logger.pureLog[IO](_.error("Harpoon processing failed to complete", err)) *>
+                  IO.pure((processedTopLevel, 0)) //we reset to the previous state as to retry next round
+          )
+
+      //a wait period before running the process once more
+      val waitRound =
+        IO.shift *>
```
kubukoz: Now that I think though, is there a reason to shift if you're only doing this once at a time? I don't see how switching contexts would help.

src/main/scala/tech/cryptonomic/conseil/process/AccountsTaggingProcess.scala
```diff
@@ -0,0 +1,156 @@
...
+            //log unexpected kind
+            logger.pureLog[F](
+              _.error(
+                """I didn't expect to process such operation kind: "{}", while flagging accounts. It should be one of {}""",
+                operation,
+                operationKinds.mkString(", ")
+              )
+            ) >> missingResult
+        }
+        effect.map { flagResult =>
```
kubukoz: `effect.tupleRight(operation.blockLevel)`?

src/main/scala/tech/cryptonomic/conseil/process/AccountsTaggingProcess.scala
```diff
@@ -0,0 +1,156 @@
...
+      }
+      .compile
+      .last
+
+  }
+
+  /* will have the method that accepts data and actualy processes it*/
+  def process(implicit api: ServiceDependencies): F[ProcessorOutput] =
+    for {
+      initialMark <- highWatermark.get
```
kubukoz: Note that this `process` function isn't atomic because you `get` and then `set`. If that matters and you predict that there might be concurrent calls to this function in the future, maybe wrap this in a `Semaphore(1)` to ensure atomicity?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -546,4 +547,61 @@ object TezosDatabaseOperations extends LazyLogging {
...
+    * @param fromLevel the lowest block-level to start from, zero by default
+    * @return the matching operations, sorted by ascending block-level
+    */
+  def fetchRecentOperationsByKind(
+      ofKind: Set[String],
+      fromLevel: Int = 0
+  ): DBIOAction[Seq[OperationsRow], Streaming[OperationsRow], Effect.Read] =
+    Tables.Operations
+      .filter(
+        row => (row.kind inSet (ofKind)) && (row.blockLevel >= fromLevel)
```
kubukoz: Just FYI `filter(x && y)` is the same as `filter(x).filter(y)` in slick

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
ivanopagano: Thanks, I'll keep it in mind, but not for this PR.
I want to keep the current logging library here, will need a more through investigation of which logging to use on a dedicated task.
There we might consider switching to log4cats on top of the logging of choice.
😉

build.sbt
ivanopagano: ouch... I thought they were aligning

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
ivanopagano: I'm using a single execution context for the whole application, which uses fs2 streams.
If the stream schedule tasks on different threads, I can feel assured that the main loop will do his job by yielding control to other threads, if needed (consider that we're doing IO on those too, such as logging).

src/main/scala/tech/cryptonomic/conseil/process/AccountsTaggingProcess.scala
ivanopagano: I'm consciously keeping it sequential, for the time being (might add a comment here to be sure).
Might change in the future, but it's not something I expect, actually. This is a simple side-process that won't probably evolve anymore.
But I see your point and note it.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: I can imagine that, but I prefer to have my boolean algebra at work to make things more "direct"... 😄 

Good to know though.

build.sbt
kubukoz: Just in the majors, I believe, and that's sort of by accident...

src/main/scala/tech/cryptonomic/conseil/Harpoon.scala
kubukoz: I'm just not sure there's a point if everything is sequential :P

vishakh:
@ivanopagano @anonymoussprocket @ba205 @bhaskarkishore Friends, I am closing this PR for now. We should soon agree on an approach for the task and adapt or modify this PR.

Here are my original comments from Riot:

> Regarding the PR for accounts tagging (https://github.com/Cryptonomic/Conseil/pull/567), the code itself appears well written but I am strongly opposed to any mutation of the database. In the PR, a secondary process runs and identifies relevant operations, based on which records in the accounts table are marked as activated or revealed.

> As we want to avoid data duplication and create another table with enriched data, I think it makes sense to do the tagging during the main Lorre loop. Already, in each iteration, Lorre examines all operations that happened in the relevant blocks and uses them to figure out which accounts were 'touched' so as to write new accounts records. This process could be enhanced so instead of simply tracking which accounts were touched we also track how they were touched. Based on this and any existing information in the accounts table, when new accounts records are written, we can add the additional fields we need.

> Let's discuss this over the coming days and nail down an agreement on how to proceed.

PR #568, opened 2019-09-30
516 - allow grouping in predicates using OR combinator

### LINKED ISSUE #516: Allow cross-predicate OR operation
Create syntax to enable an OR condition between predicates. Currently it can only be AND. Only support a single-level nesting along the lines of `(predicate or predicate) and predicate` and `(predicate and predicate) or predicate`.

- Predicate group may contain more than two predicates.
- There may be multiple predicate groups.

ivanopagano:
closes #516 

Add an extra optional property to each predicate, such that they can be "grouped" via a common name.
By default within each group the predicates will be AND-ed, and between groups they'll be OR-ed.
A predicate with no group name belongs to a default name-less group.

No changes to the DB schema nor configuration are needed.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -354,6 +354,7 @@ It is possible to narrow down the fields returned in the result set by listing j
 - `set` – an array of values to compare against. 'in' requires two or more elements, 'between' must have exactly two, the rest of the operations require a single element. Dates are expressed as epoch milliseconds.
 - `inverse` – boolean, setting it to `true` applied a `NOT` to the operator
 - `precision` – for numeric field matches this specifies the number of decimal places to round to.
+- `group` - an optional naming to collect predicates in groups that will be eventually OR-ed amongst each other
 
 #### `limit`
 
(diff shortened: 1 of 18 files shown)


docs/README.md
```diff
@@ -613,6 +614,33 @@ Send this query to `/v2/data/tezos/<network>/accounts`
...
+#### Blocks by end-user-transactions outside April 2019
+
+Send this query to `/v2/data/tezos/<network>/operations`
+
+```json
+{
+    "fields": ["block_level", "operation_group_hash"],
+    "predicates": [
+        { "field": "timestamp", "set": [1554076800000], "operation": "before", "inverse": false, "group": "before" },
+        { "field": "kind", "set": ["transaction"], "operation": "in", "inverse": false, "group": "before" },
```
ba205: Will group names other "before"/"after" be considered? Or more than two groups, or is this intentionally to restrict "or" nesting?

anonymoussprocket: group names are arbitrary. there is no nesting

ivanopagano: Do you think I might be making a more detailed explanation of how the example is interpreted?

ba205: It's probably fine, I just got confused because "before" and "after" is a binary way at looking at groups. Maybe different group names would be enough.

Issue #521, opened 2019-08-26
Post-process ballot and proposal data


anonymoussprocket:
This issue depends on #520.

Create a new table that combines the ballot and proposal information in a meaningful way. This is conceptually similar to the way we already process fees. The idea is to associate block level, vote count (yea, nay, pass), voted stake total (yea, nay, pass) (sum of the balances of the voters) and proposal hash.

Note this process should occur outside of the mail Lorre loop and should potentially be runnable independently of the rest of the chain loading code on separate hardware.

ba205:
Logic for Algorithm:

**Tezos RPC Routes**

GET ../<block_id>/context/delegates/<pkh>/staking_balance
{{protocol}}://{{hostname}}:{{port}}/{{path_prefix}}/chains/main/blocks/head/votes/current_period_kind

**Database Schema**

  case class VotesRow(
      timestamp: java.sql.Timestamp,
      cycle: Option[Int] = None,
      level: Option[Int] = None,
      proposalHash: String,
      yayCount: Option[Int] = None,
      nayCount: Option[Int] = None,
      passCount: Option[Int] = None,
      yayStake: Option[scala.math.BigDecimal] = None,
      nayStake: Option[scala.math.BigDecimal] = None,
      passStake: Option[scala.math.BigDecimal] = None,
      totalStake: Option[scala.math.BigDecimal] = None,
      periodKind: String,
      periodIndex: Option[Int] = None
     )

**Algorithm**
-Fetch range of block levels to be queried for operations, from head level in Votes Table to head Level in Blocks Table
-For each block level
  -fetch voting fields from database, via ballot operation if proposalKind is in testing vote or promotion vote period, proposal operation otherwise, group by block level, as aggregation is done via that
  -do post processing calculations for stakes/counts, if ballot
  -create database rows, and write

piotrkosecki:
@ba205 I think there is something wrong with the Tezos routes or maybe I misunderstood the way to use them - I can't make those routes to work, can you double check them?

vishakh:
@piotrkosecki  The previous spec need to be greatly amended. Here is the latest spec. I would recommend creating a fresh PR and reusing the bits from the old PR which make sense. 

A new governance table should track the currently active Tezos governance proposal and the corresponding votes. 

The table schema should be:

CREATE TABLE tezos.governance (
    voting_period integer NOT NULL,
    voting_period_kind character varying NOT NULL,
    cycle integer,
    level integer,
[...]

PR #569, opened 2019-09-30
Feature/conseil 521 votes table

ba205:
Logic is finished aside from separate process step, I have to fix the database conversion as well, handles this ticket: #521


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -289,6 +285,91 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+          tezosNodeOperator.getAccountBalanceForBlock(x._1, x._2)
+        }
+        .map(balances => balances.reduceOption(_ + _).getOrElse(0))
+
+    def processVoteAggregate(votingFields: Seq[VotingFields]): Future[TezosTypes.VoteAggregates] = {
+      val sampleVotingField = votingFields.head
+      val timestamp = sampleVotingField._5
+      val cycle = sampleVotingField._4
+      val level = sampleVotingField._2
+      val proposalHash = sampleVotingField._1
```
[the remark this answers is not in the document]
ivanopagano: Additionally, are we sure that those fields will match on a single `level`?
Any potential discrepancy, e.g. timestamp?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -289,6 +285,91 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+  private[this] def processTezosVotes()(implicit ex: ExecutionContext): Future[Option[Int]] = {
+    import cats.instances.list._
+    import cats.instances.future._
+    import cats.syntax.traverse._
+    import cats.implicits._
+    import slick.jdbc.PostgresProfile.api._
+
+    logger.info("Processing latest Tezos votes data...")
+
+    def partitionByVote(fields: Seq[VotingFields]): (Seq[VotingFields], Seq[VotingFields], Seq[VotingFields]) = {
```
ivanopagano: missing docs

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -289,6 +285,91 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+      val (yays, notYays) = fields.partition(x => x._6 == Some("yay"))
+      val (nays, notYaysOrNays) = notYays.partition(x => x._6 == Some("nay"))
+      val (passes, _) = notYaysOrNays.partition(x => x._6 == Some("pass"))
+      (yays, nays, passes)
+    }
+
+    def calculateStake(fields: Seq[VotingFields]): Future[BigDecimal] =
```
ivanopagano: missing docs

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -289,6 +285,91 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+    import cats.implicits._
+    import slick.jdbc.PostgresProfile.api._
+
+    logger.info("Processing latest Tezos votes data...")
+
+    def partitionByVote(fields: Seq[VotingFields]): (Seq[VotingFields], Seq[VotingFields], Seq[VotingFields]) = {
+      val (yays, notYays) = fields.partition(x => x._6 == Some("yay"))
+      val (nays, notYaysOrNays) = notYays.partition(x => x._6 == Some("nay"))
+      val (passes, _) = notYaysOrNays.partition(x => x._6 == Some("pass"))
+      (yays, nays, passes)
```
ivanopagano: I think we can do the same with `groupBy(_._6)` to get an indexed map and then ask for the values
```scala
(index.get(Some("yay")), index.get...)
```

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -289,6 +285,91 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+
+    def calculateStake(fields: Seq[VotingFields]): Future[BigDecimal] =
+      Future
+        .traverse(
+          fields.map {
+            case (_, _, hash, _, _, _, address) =>
+              (BlockHash(hash), AccountId(address.get))
+          }
+        ) { x =>
+          tezosNodeOperator.getAccountBalanceForBlock(x._1, x._2)
```
ivanopagano: Maybe fusing the steps might simplify?
```scala
Future.traverse(fields){
  case (_, _, hash, _, _, _, address) =>
    tezosNodeOperator.getAccountBalanceForBlock(BlockHash(hash), AccountId(address.gett))
}.map[...]
```

Also, what if `address` is not there? Calling `.get` will make it explode and crash the process.

Better to explicitly consider how to deal with the missing data.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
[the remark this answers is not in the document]
vishakh: It seems all of these fields are intrinsic to a block, i.e. @ba205 is pulling all ballot operations for a block and then betting they have the same block metadata. If this is indeed the case, these fields should be pulled from blocks data and not operations data.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
vishakh: Stake should be derived from the `bakers` table, not the `accounts` table. 'Stake` has a very particular meaning in Tezos and the correct way to pull it out is using baker information.

@anonymoussprocket

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -289,6 +285,91 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+        .map(balances => balances.reduceOption(_ + _).getOrElse(0))
+
+    def processVoteAggregate(votingFields: Seq[VotingFields]): Future[TezosTypes.VoteAggregates] = {
+      val sampleVotingField = votingFields.head
+      val timestamp = sampleVotingField._5
+      val cycle = sampleVotingField._4
+      val level = sampleVotingField._2
+      val proposalHash = sampleVotingField._1
+      val (yays, nays, passes) = partitionByVote(votingFields)
+      val yayCount = yays.length
```
vishakh: This information is also available directly from the Tezos client and used to be stored in the `ballots` table. We should revive the `ballots` table and ensure our tally matches their tally.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -163,6 +163,38 @@ object TezosTypes {
...
+    * @param yayCount      Total count of yays for proposal
+    * @param nayCount      Total count of nays for proposal
+    * @param passCount     Total count of passes for proposal
+    * @param yayStake      Total stake of accounts who voted yay this block
+    * @param nayStake      Total stake of accounts who voted nay this block
+    * @param passStake     Total stake of accounts who voted pass this block
+    * @param totalStake    Total stake of accounts who voted
+    *
+    */
+  final case class VoteAggregates(
```
vishakh: I believe this file only contains types that reflect the types returned by the Tezos RPC. `VoteAggregates` is a 'synthetic' type so it should be placed in an appropriate file. As @ivanopagano pointed out, we need a case class instead of a type elsewhere where you have aggregated votes.

doc/conseil.sql
```diff
@@ -323,10 +347,18 @@ ALTER TABLE ONLY public.operations ALTER COLUMN operation_id SET DEFAULT nextval
...
+ALTER TABLE ONLY public.operation_groups
     ADD CONSTRAINT "OperationGroups_pkey" PRIMARY KEY (block_id, hash);
 
 
+--
+-- Name: Votes Votes_pkey; Type: CONSTRAINT; Schema: public; Owner: -
+--
+
+ALTER TABLE ONLY public."Votes"
+    ADD CONSTRAINT "Votes_pkey" PRIMARY KEY (proposal_hash);
```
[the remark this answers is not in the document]
ba205: You're correct, this is an old sql file, I'll update accordingly. Tables.scala is correct :)

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -109,6 +104,7 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
     val noOp = Future.successful(())
     val processing = for {
       _ <- processTezosBlocks()
+      _ <- processTezosVotes()
```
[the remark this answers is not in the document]
ba205: I did this at the request of @anonymoussprocket's original intention for this to be a separate process from Lorre. At the very least, vote aggregation has to be done after operation collection in "processTezosBlocks".

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ba205: @anonymoussprocket and I discussed that there was no easy way to get account balances as of block level from the current database schema, so we hit the tezos node in this function, not the database. 

I'm not a fan of this approach actually, if querying the baker's table (or the db in general) would solve this issue, I'd be okay with it.

ba205: Very good points, I'll add them in.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -289,6 +285,91 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+          fields.map {
+            case (_, _, hash, _, _, _, address) =>
+              (BlockHash(hash), AccountId(address.get))
+          }
+        ) { x =>
+          tezosNodeOperator.getAccountBalanceForBlock(x._1, x._2)
+        }
+        .map(balances => balances.reduceOption(_ + _).getOrElse(0))
+
+    def processVoteAggregate(votingFields: Seq[VotingFields]): Future[TezosTypes.VoteAggregates] = {
```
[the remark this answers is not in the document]
ba205: Got it!

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ba205: @anonymoussprocket thoughts?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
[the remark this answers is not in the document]
ivanopagano: If it's independent, it may well be run whenever it happens.

In any case I'd second the idea of having multiple _independent_ processings.
The main point is to figure out what needs to be absolutely consistent with the fetched blocks, and what might be updated more leniently.
If we can make processing independent of the main loop, we have much less work to do to guarantee consistency of all different information stored in the system, which requires some form of "transaction" logic in the main block-processing loop.
An example are accounts checkpoints.
We've gone to great length to guarantee that no account gets lost if something breaks during processing, and it's an additional burden, mentally and programmatically, to maintain such logic.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -289,6 +285,91 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+        )
+    }
+
+    //get fields from db in Tezos Database Operations, use the block levels as input
+    val computeAndStore = for {
+      blockLevels <- TezosDb.fetchVotingBlockLevels
+      listOfVotingFields <- TezosDb.fetchVotingFields(blockLevels)
+      listOfListOfVotingFields = listOfVotingFields.groupBy(_._2).values
+      votes <- DBIO.from {
+        Future.traverse(listOfListOfVotingFields.toSeq) { x =>
```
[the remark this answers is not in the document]
ba205: I'll do the latter!

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -465,6 +476,43 @@ object TezosDatabaseOperations extends LazyLogging {
       .getOrElse(defaultBlockLevel.toInt)
       .result
 
+  /** Get next batch of voting levels */
```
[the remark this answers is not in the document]
ba205: Sure thing!

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -465,6 +476,43 @@ object TezosDatabaseOperations extends LazyLogging {
...
+        .getOrElse(defaultBlockLevel.toInt)
+        .result
+
+    for {
+      headBlockLevel <- fetchMaxBlockLevel
+      voteMaxLevel <- fetchVotesMaxLevel
+    } yield (voteMaxLevel + 1 to headBlockLevel).toSet
+  }
+
+  type VotingFields = (Option[String], Int, String, Option[Int], java.sql.Timestamp, Option[String], Option[String])
```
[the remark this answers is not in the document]
ba205: That's definitely a better idea :)

src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
ba205: Will move to a separate file.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ba205: Oh, I misread this, do you mean the baker rpc for tezos?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -289,6 +285,91 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+    * Collect data from Blocks Table, Operations Table, and the Tezos RPC,
+    * return the appropriate list of VoteAggregates, and write them to the Votes table.
+    * @param ex the needed ExecutionContext to combine multiple database operations
+    * @return a future result of the number of rows stored to db, if supported by the driver
+    */
+  private[this] def processTezosVotes()(implicit ex: ExecutionContext): Future[Option[Int]] = {
+    import cats.instances.list._
+    import cats.instances.future._
+    import cats.syntax.traverse._
+    import cats.implicits._
```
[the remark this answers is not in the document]
ba205: will remove those which do not break the code

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ba205: On it

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
vishakh: It's best if you can get it from RPC. The table would be second best.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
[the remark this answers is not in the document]
vishakh: OK, sounds like we should have a proper discussion about this.

For this PR, I'm fine with proceeding with this bit as-is. We can refactor it later.

vishakh:
Replaced by #658

PR #572, opened 2019-10-02
570 - fix for the endorsing/baking rights

### LINKED ISSUE #570: Fetching endorsing/baking operations fails because of connections limit in Lorre
The issue can be resolved with using batched queries for fetching baking/endorsing rights with given block.

piotrkosecki:
resolves #570 

Changed way of fetching endorsing rights. Now they are being batched.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
@@ -1,6 +1,6 @@
 package tech.cryptonomic.conseil.tezos
 
-import java.time.Instant
+import java.time.ZonedDateTime
 
 import monocle.Traversal
 import monocle.function.all._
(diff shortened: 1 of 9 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/TezosTypes.scala
```diff
@@ -531,14 +531,15 @@ object TezosTypes {
       level: Int,
       delegate: String,
       priority: Int,
-      estimated_time: java.sql.Timestamp
+      estimated_time: ZonedDateTime
```
bhaskarkishore: Curious, Why have we moved away from simple timestamps?

ivanopagano: I've suggested this to keep the domain model clean of things related to a database.
This helps immediately to decode json datetimes with no custom decoding definitions, and allows (if ever needed) more handle to operate on the datetime values.

PR #573, opened 2019-10-03
493 - updated API operations tests

### LINKED ISSUE #493: ApiOperations has no tests
Some care could be taken to write appropriate UnitTests for the ApiOperations class

piotrkosecki:
resolves #493 

Made a refactor of ApiOperations, added basic tests. I'll need to add more tests, especially those using DB(I wonder if those won't be redundant at some point to the tests in `TezosDatabaseOperations`) - thats why I'm opening draft PR, but for now it can be reviewed.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
@@ -1,29 +1,40 @@
 package tech.cryptonomic.conseil.tezos
 
 import slick.jdbc.PostgresProfile.api._
-import tech.cryptonomic.conseil.generic.chain.{DataOperations, DataTypes, MetadataOperations}
-import tech.cryptonomic.conseil.generic.chain.DataTypes.{
-  Field,
-  FormattedField,
-  OperationType,
-  OrderDirection,
-  Predicate,
-  Query,
-  QueryOrdering,
-  QueryResponse,
-  SimpleField
-}
+import tech.cryptonomic.conseil.generic.chain.{DataOperations, MetadataOperations}
+import tech.cryptonomic.conseil.generic.chain.DataTypes.{Field, FormattedField, Predicate, Query, QueryResponse, SimpleField}
+import tech.cryptonomic.conseil.tezos.ApiOperations.{AccountResult, BlockResult, OperationGroupResult}
 import tech.cryptonomic.conseil.tezos.TezosTypes.{AccountId, BlockHash}
 import tech.cryptonomic.conseil.tezos.{TezosDatabaseOperations => TezosDb}
 import tech.cryptonomic.conseil.util.DatabaseUtil
 
 import scala.concurrent.{ExecutionContext, Future}
 
+
+object ApiOperations {
+  case class BlockResult(block: Tables.BlocksRow, operation_groups: Seq[Tables.OperationGroupsRow])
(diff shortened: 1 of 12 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -1,29 +1,40 @@
...
+}
+
 /**
   * Functionality for fetching data from the Conseil database.
   */
-object ApiOperations extends DataOperations with MetadataOperations {
-
+class ApiOperations extends DataOperations with MetadataOperations {
+  import ApiOperations._
   lazy val dbReadHandle: Database = DatabaseUtil.conseilDb
```
ivanopagano: What about passing this in as a parameter?

src/test/scala/tech/cryptonomic/conseil/tezos/ApiOperationsTest.scala
```diff
@@ -1,11 +1,112 @@
...
+    "sanitizeForSql alphanumeric string with unsupported characters" in {
+      // given
+      val input = ";xyz$%*)("
+
+      // when
+      val result = ApiOperations.sanitizeForSql(input)
+
+      // then
+      result shouldBe "xyz"
+    }
```
ivanopagano: Is it really necessary to use three separate tests instead of a single case containing good and bad chars and verifying that only the good ones are in?

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
piotrkosecki: passing DB as a parameter would require adding DB in the tests which don't use database(only stubs/mocks of the ApiOperations), or maybe do you have other proposal how to handle it?

ivanopagano: I guess the parameterization would look something like `DBIOAction[T] => Future[T]` with no reference to a `Database`.

I know by direct handling this that using the latter as a generic definition is terrible to mock. Instead you can easily get away with it, using the "function" parameter.

Still, do you think it might improve the situation? I can't be sure from this point.

piotrkosecki: I'm not sure what you have in mind. Parametrising by `[T]` would require creating parameter for class and `[T]` would be different for each method. Or am I missing something and I misunderstood your idea?

src/test/scala/tech/cryptonomic/conseil/tezos/ApiOperationsTest.scala
piotrkosecki: no, it's not necessary, but I think reading test output it's easier to understand what happened

src/test/scala/tech/cryptonomic/conseil/tezos/ApiOperationsTest.scala
```diff
@@ -20,16 +209,112 @@ class ApiOperationsTest extends WordSpec with Matchers {
...
+
+    }
+
+
+    "fetchMaxLevel when DB is empty" in {
+      // when
+      val result = sut.fetchMaxLevel().futureValue
+
+      // then
+      result shouldBe -1
```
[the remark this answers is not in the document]
piotrkosecki: it was described in the javadoc of the `fetchMaxLevel` method 
`@return Max level or -1 if no blocks were found in the database.`

PR #575, opened 2019-10-04
Add the ability to configure an API key when launching in docker.

Pierre-Michard:
- Add the ability to configure an API key when running the conseil API in docker.
- Update the Readme regarding this feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ Conseil Config:
 
 
 * API_PORT - Conseil API port, default: 80
+* API_KEY - Conseil API key, default: conseil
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -41,6 +41,7 @@ Conseil Config:
 
 
 * API_PORT - Conseil API port, default: 80
+* API_KEY - Conseil API key, default: conseil
```
vishakh: This is good for now but ideally we should take a plural of API keys. Perhaps we can do this under a separate ticket unless @Pierre-Michard feels generous enough to make the change now. :smile:

Pierre-Michard: Sorry, I can't think of an elegant way to configure an array of strings in an env variable.

PR #577, opened 2019-10-07
574 - Shutdown the starting system after retrying if the metadata is not available at init

### LINKED ISSUE #574: Exception in main when starting does not stop conseil process
When starting conseil API in a docker container, we experience the following issue sometimes:
```
conseil.db - Driver does not support get/set network timeout for connections. (Method org.postgresql.jdbc.PgConnection.getNetworkTimeout() is not yet implemented.)
Exception in thread "main" java.util.concurrent.TimeoutException: Futures timed out after [2 seconds]
	at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:259)
	at scala.concurrent.impl.Promise$DefaultPromise.result(Promise.scala:263)
	at scala.concurrent.Await$.$anonfun$result$1(package.scala:219)
	at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:57)
	at scala.concurrent.Await$.result(package.scala:146)
	at tech.cryptonomic.conseil.metadata.MetadataService.<init>(MetadataService.scala:30)
	at tech.cryptonomic.conseil.Conseil$.delayedEndpoint$tech$cryptonomic$conseil$Conseil$1(Conseil.scala:83)
	at tech.cryptonomic.conseil.Conseil$delayedInit$body.apply(Conseil.scala:25)
	at scala.Function0.apply$mcV$sp(Function0.scala:39)
	at scala.Function0.apply$mcV$sp$(Function0.scala:39)
	at scala.runtime.AbstractFunction0.apply$mcV$sp(AbstractFunction0.scala:17)
	at scala.App.$anonfun$main$1$adapted(App.scala:80)
	at scala.collection.immutable.List.foreach(List.scala:392)
	at scala.App.main(App.scala:80)
	at scala.App.main$(App.scala:78)
	at tech.cryptonomic.conseil.Conseil$.main(Conseil.scala:24)
	at tech.cryptonomic.conseil.Conseil.main(Conseil.scala)
```
This would not be an issue if it stopped the whole process, since docker compose and kubernetes have the ability to restart a dead container. 

Nevertheless the process continue but does not start the API server, leaving it in a zombie mood.

Would that be possible to find a way to kill the whole process when the main thread dies?

ivanopagano:
closes #574 

As suggested in the issue, the whole process can be stopped if initialization should fail upon metadata pre-fetching.
The user can now pass a new option on the commandline: `-f` or `--fail-fast`.

If such flag is set, the system will stop as soon as the init fails.
The default behaviour is to retry initialization for a given amount of time - set from config at 1 minute - after which the process will stop.
The retries are done with an exponential backoff policy.

## Configuration changes
the `reference.conf` file has now a new entry
```
conseil.startup-deadline
```
which can be overridden in the `conseil.conf` to allow for shorter or longer startup deadlines.

## No Api changes


src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -31,115 +32,155 @@ object Conseil
     with ConseilOutput {
 
   loadApplicationConfiguration(args) match {
-    case Right((server, platforms, securityApi, verbose, metadataOverrides, nautilusCloud)) =>
+    case Left(errors) =>
```
ba205: You can just have an empty case match with no error handling?

ivanopagano: I'm not sure I get the question...
Are you asking if you're not needed to actually put _some code_ in one of the `case` of the match?
The answer to this is "yes, as long as the return type of all branches of the match are consistent". Here consistent mean that they can be inferred to have a common shared super-type, that will be the return type of the whole match expression.

If the question was another, then I kindly ask you to clarify.

src/main/scala/tech/cryptonomic/conseil/util/Retry.scala
```diff
@@ -0,0 +1,133 @@
+package tech.cryptonomic.conseil.util
```
ba205: This code is extremely readable, wow. We could learn something from this :(

ba205: @vishakh

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -31,115 +42,157 @@ object Conseil
...
+
+    nautilusCloud.foreach { ncc =>
+      system.scheduler.schedule(ncc.delay, ncc.interval)(Security.updateKeys(ncc))
+    }
+
+    // This part is a temporary middle ground between current implementation and moving code to use IO
+    implicit val contextShift: ContextShift[IO] = IO.contextShift(executionContext)
+    val metadataCaching = MetadataCaching.empty[IO].unsafeRunSync()
+
+    lazy val transformation = new UnitTransformation(metadataOverrides)
```
ba205: In hindsight, it's not obvious what UnitTransformation means in this context, just from looking at this line of code. Maybe a rename is in order for a different ticket?

src/main/scala/tech/cryptonomic/conseil/util/Retry.scala
ivanopagano: @ba205 please elaborate more, I'm very interested to hear about possible improvements
What you feel here is more readable? What could we borrow as a general approach and use in other parts?

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
ivanopagano: I's curious too, I guess the name is borrowed from math to mean that it will take an existing object and apply a transform that won't change the basic overall metrics (so to speak)?

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -31,115 +42,157 @@ object Conseil
...
-          contextShift
-        )
+  }
 
-      tezosPlatformDiscoveryOperations.init().onComplete {
-        case Failure(exception) => logger.error("Pre-caching metadata failed", exception)
-        case Success(_) => logger.info("Pre-caching successful!")
-      }
+  /** Reads configuration to setup the fundamental application services
+    * @param appConfig static configuration as read from files
```
vishakh: Please fix the parameter docs.

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -31,115 +42,157 @@ object Conseil
...
-              }
-            )
-          }
+    // this val is not lazy to force to fetch metadata and trigger logging at the start of the application
+    Try(new MetadataService(platforms, transformation, cacheOverrides, tezosPlatformDiscoveryOperations))
+  }
+
+  /** Starts the web server
+    * @param metadataService the metadata information to build the querying functionality
+    * @param appConfig static configuration as read from files
```
vishakh: Please fix the parameter docs.

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -31,115 +42,157 @@ object Conseil
...
-        apiKeyTuple =>
-          val apiKey = apiKeyTuple match {
-            case Tuple1(apiKey) => apiKey
-            case _ => None
-          }
+      val retries = if (failFast.on) Some(0) else None
+
+      val serverBinding =
+        retry(maxRetry = retries, deadline = Some(server.startupDeadline fromNow))(
+          initServices(server, platforms, metadataOverrides, nautilusCloud).get
```
vishakh: How come we `get` from the Try returned by `initServices()` and not explicitly handle any errors?

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -31,115 +42,157 @@ object Conseil
...
+        ).andThen {
+          case Failure(error) =>
+            logger.error(
+              "The server was not started correctly, I failed to create the required Metadata service",
+              error
+            )
+            Await.ready(system.terminate(), 10.seconds)
```
vishakh: Why is this hard coded to 10 seconds?

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -31,115 +42,157 @@ object Conseil
     with ConseilOutput {
 
   loadApplicationConfiguration(args) match {
-    case Right((server, platforms, securityApi, verbose, metadataOverrides, nautilusCloud)) =>
+    case Left(errors) =>
```
vishakh: Instead of writing this out as a code block could we make the running of Conseil a one-line composition of multiple functions? This will make the code a lot clearer to read.

src/main/resources/reference.conf
```diff
@@ -36,6 +36,7 @@ conseil {
   cache-ttl: 15 minutes
   max-query-result-size: 100000
   high-cardinality-limit: 100
+  startup-deadline: 1 minute
```
vishakh: I would recommend a longer deadline. We use this functionality when multiple containers are being brought up together. It's possible the database container might take longer to boot up. How about five minutes?

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
```diff
@@ -43,13 +43,13 @@ class MetadataService(
         entities.map(entity => networkPath.addLevel(entity.name))
     }.toSet
 
-    val result = networkPaths.map { path =>
+    val result = Future.traverse(networkPaths){ path =>
```
vishakh: Do I understand correctly that this change makes the code more parallel, i.e. multiple futures are spawned together instead of being spawned sequentially?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosPlatformDiscoveryOperations.scala
```diff
@@ -391,28 +391,27 @@ class TezosPlatformDiscoveryOperations(
...
   def initAttributesCache: Future[Unit] = {
-    val result = for {
+    for {
       _ <- caching.updateCachingStatus(InProgress)
       entCache <- caching.getEntities(networkName)
       attributes <- caching.getAllAttributes
       _ <- contextShift.shift
-      updatedAttributes <- entCache.fold(IO(Map.empty[String, List[Attribute]]))(
-        entC => getAllUpdatedAttributes(entC.value, attributes)
+      updatedAttributes <- entCache.foldMapM(
```
vishakh: Could you explain briefly the change you made here?

ivanopagano: The original code used fold on an Option to define what to do 
- when the value was there: get the attributes, returning a map wrapped in IO
- when the value was missing: return an empty map wrapped in IO

`foldMap` would do the exact same, using the fact that a Map has a Monoid instance that defines an empty value, so you won't need to explicitly say what it is, it's the most natural one.

`foldMapM` takes into account that the mapping part produced a wrapping Monad (IO), therefore it automatically matches the returned empty monoid result by wrapping it into the needed monad.

It's essentially doing grunt work for free, adding what is expected to be the most natural empty value, instead of needing from the outside.

ivanopagano: To complete the picture, the original code then mapped on such IO to cache each entry of the map into the cache (using putAttributes(...)).

Each of those produced a Unit value. Hence the original `result` was a `Seq[IO[Unit]]` wrapped in a extra IO, which was useless, unless you sequenced those into a single IO value to be run.
Thus the `flatMap(_.toList.sequence)` action.

I simply used `traverse` to do the same within the same for comprehension, ignoring the useless result, and finally adding the last IO operation (originally `>> caching.updateCachingStatus`) to the same comprehension again.

I also removed the conversion to a `Map` and back `.toList` by using a `List` of key/value directly, since the Map structure wasn't actually ever used to index the entries. The list was only ever traversed linearly.

src/main/scala/tech/cryptonomic/conseil/util/Retry.scala
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+import scala.annotation.tailrec
+import scala.concurrent.{blocking, Await, ExecutionContext, Future, Promise}
+import scala.concurrent.duration.{Deadline, Duration, DurationLong}
+import scala.language.postfixOps
+import scala.util.{Failure, Random, Success, Try}
+
+object Retry {
```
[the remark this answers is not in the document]
ivanopagano: Done

src/main/scala/tech/cryptonomic/conseil/metadata/MetadataService.scala
ivanopagano: In fact it's a side-effect of switching from `map` and then `sequence`, I didn't realise the fact.
I did it out of habit, because doing both operations is always equivalent to a single `traverse` and I refactor it without thinking to make the intent explicit.

In this case we might've gained a little performance too, possibly, but it wasn't intentional on my part.

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
ivanopagano: Done

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
ivanopagano: The system is closing anyway, so it's just to give it some room to cleanly close any resource taken, but generally I don't think that anything bad might happen if we don't wait. The Jvm process would be killed anyway.

I think the system will usually be terminated in far less, or will hang on until forced because something is "stuck".
E.g. Lorre's process usually can't close all the streams in a timely manner, but waiting longer won't resolve the thing anyway as far as my trial showed.

tl;dr; I don't consider this value a critical one, it's simply an arbitrary time needed by the code. Making it configurable won't actually change much in my understanding.

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
ivanopagano: That's how the `retry` logic expects to work.
It spawns a "thunk" of code in a Future, and internally catches any error to be able to handle the retry logic independently.

It provides parameters to define how long to retry, how long to wait for each retry and if there's any specific expection to pass through without retrying, and simply failing the call immediately

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
ivanopagano: Do you mind if I join this proposal with #581 to have a more uniform definition for both the main application entrypoints?
@vishakh

src/main/resources/reference.conf
ivanopagano: No problem

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
ivanopagano: Done

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
vishakh: As discussed, this will be done as a separate effort.

PR #582, opened 2019-10-11
Various bug fixes for proto5 logic

vishakh:
High level changes:

- The 'Account' type and table schema have been brought into alignment with the actual proto5 Tezos RPC JSON payload.
- After discussion with @anonymoussprocket the `delegated_contracts` table has been removed as it's proven unnecessary and was causing issues in the new code. 
- New foreign keys from `baking_rights` and `endorsing_rights` to `blocks` have been made as they make sense.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/michelson/renderer/MichelsonRenderer.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/michelson/renderer/MichelsonRenderer.scala
@@ -54,6 +54,6 @@ object MichelsonRenderer {
         .map(_.render())
         .mkString(" ;\n")
         .lines
-        .mkString("\n" + indent)
+      .toString()
   }
 }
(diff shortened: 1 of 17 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/michelson/renderer/MichelsonRenderer.scala
```diff
@@ -54,6 +54,6 @@ object MichelsonRenderer {
         .map(_.render())
         .mkString(" ;\n")
         .lines
-        .mkString("\n" + indent)
+        .mkString(s"\n${indent}")
```
vishakh: @dorians I had to change this line because I was getting a compiler error before: 

```
[error] /home/vishakh/dev/cryptonomic/Conseil/src/main/scala/tech/cryptonomic/conseil/tezos/michelson/renderer/MichelsonRenderer.scala:59:10: value mkString is not a member of java.util.stream.Stream[String]
[error] possible cause: maybe a semicolon is missing before `value mkString'?
[error]         .mkString("\n" + indent)
[error]          ^
```

Could you tell me how to fix it peoperly?

ivanopagano: The issue is with the `.lines` call, which I think we have fixed before in previous PRs.

The jdk9+ introduced a `.lines` method that conflicts with the existing one from scala, and returns a type that's obviously different, so that `mkString` is not available.

tl;dr; you only need to change `.lines` with `linesIterator`

PR #609, opened 2019-10-23
Remove hard coding from conseil.sql

vishakh:
The hard coded user name and database name in the SQL file is breaking our infrastructure deployments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,14 @@ The application expects to read and write from a database compatible with [Types
 
 ##### Using a database instance
 
-Cryptonomic uses Postgres for all its Conseil deployments. Once a Postgres database is set up, `doc/conseil.sql` can be used to set up the latest schema.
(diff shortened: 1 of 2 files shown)


ivanopagano:
We might want to add automated handling of the search_path for 
- docker images built with our dockerfile, adding a psql command that runs the "ALTER ROLE ..." to the `entrypoint.sh` script before running the java command to run the app
- local development docker dbs run with the docker-compose.yml file, adding an additional `conseil.init.sql` file that run the `ALTER ROLE ...` after conseil.sql is executed by the docker image... yet this might be redundant (I would have to test it) if the role creating the schema sees it on its search_path

vishakh:
@ivanopagano I would like @piotrkosecki to first investigate whether we can do without setting the search path entirely (#610).

ivanopagano:
> @ivanopagano I would like @piotrkosecki to first investigate whether we can do without setting the search path entirely (#610). This PR is mean to simply remove the offending lines from the SQL file.

I see, I only have the feeling that my local postgres instances for development will now be broken by the change...

PR #615, opened 2019-10-31
404 - changes to temporal queries

### LINKED ISSUE #404: Make accounts table temporal
Currently there is no way to view account history as Lorre updates account data in place in the database. We must therefore make the accounts table temporal. 

We can accomplish this by adding an `asof` column to the `accounts` table. Here is a caricature:

```
account_id | balance | asof | block_level
tz1kjdjk41o421kj2 | 20203437843 | 4pm | 500k
tz1kjdjk41o421kj2 | 20203437850 | 2pm | 475k
```

A query to get the latest balance for a given account would be of this form:

```sql
select * from accounts where account_id = 'tz1kjdjk41o421kj2' where asof<= now() order by asof desc limit 1;
```

A query to get the balance as of a given time would be of this form:

```sql
select * from accounts where account_id = 'tz1kjdjk41o421kj2' where asof<= 3 pm order by asof desc limit 1;
```

Finally, account history can be obtained with a simple query:

```sql
select * from accounts where account_id = 'tz1kjdjk41o421kj2' 
```

piotrkosecki:
Resolves #404 

Made changes to the the query, updated doc.

## Required changes to the DB

`accounts_history` table needs to change the `asof` column type because metadata couldn't recognize correct one:
```
asof timestamp without time zone NOT NULL
```

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
@@ -45,7 +45,7 @@ object DataTypes {
       implicit executionContext: ExecutionContext
   ): Future[Query] = {
     val fetchedAttributes = metadataService.getTableAttributesWithoutUpdatingCache(path)
-    val newSnapshot =
+    val newSnapshot: Future[Option[Snapshot]] =
       fetchedAttributes.map { maybeAttributes =>
         query.snapshot.flatMap { snapshot =>
           maybeAttributes.flatMap { attributes =>
@@ -58,9 +58,9 @@ object DataTypes {
         }
       }
 
-    val newPredicates =
+    val newPredicates: Future[List[Predicate]] =
       fetchedAttributes.map { maybeAttributes =>
-        query.predicates.map { predicate =>
+        query.predicates.flatMap { predicate =>
           maybeAttributes.flatMap { attributes =>
             attributes.find(_.name == dropAggregationPrefixes(predicate.field)).map {
               case attribute if attribute.dataType == DataType.DateTime =>
@@ -72,7 +72,7 @@ object DataTypes {
       }
     (newPredicates, newSnapshot).mapN {
       case (pred, snap) =>
-        query.copy(predicates = pred.flatten, snapshot = snap)
+        query.copy(predicates = pred, snapshot = snap)
     }
   }
 
@@ -326,23 +326,22 @@ object DataTypes {
       query: Query,
       entity: EntityPath,
       metadataConfiguration: MetadataConfiguration
-  ): List[InvalidSnapshotField] =
-    query.snapshot match {
-      case Some(snapshot) =>
-        metadataConfiguration.entity(entity).flatMap { entityCfg =>
-          entityCfg.attributes.filter {
-            case (_, v) =>
-              v.temporalColumn.getOrElse(false)
-          }.find {
-            case (k, _) =>
-              snapshot.field == k
-          }
-        } match {
-          case Some(_) => List.empty
-          case None => List(InvalidSnapshotField(snapshot.field))
+  ): List[InvalidSnapshotField] = {
+    import cats.instances.option._
+    import cats.instances.list._
+
+    query.snapshot.foldMap { snapshot =>
+      metadataConfiguration.entity(entity).flatMap { entityCfg =>
+        entityCfg.attributes.find {
+          case (k, v) if k == snapshot.field =>
+            v.temporalColumn.getOrElse(false)
         }
-      case None => List.empty
+      } match {
+        case Some(_) => List.empty
+        case None => List(InvalidSnapshotField(snapshot.field))
+      }
     }
+  }
 
   /** Class representing query got through the REST API */
   case class ApiQuery(
@@ -389,10 +388,10 @@ object DataTypes {
             updatedQuery
         ) =>
           invalidNonExistingFields :::
-            invalidAggregationFieldForTypes :::
-            invalidPredicateFilteringFields :::
-            invalidQueryFieldFormats :::
-            invalidSnapshotField match {
+              invalidAggregationFieldForTypes :::
+              invalidPredicateFilteringFields :::
+              invalidQueryFieldFormats :::
+              invalidSnapshotField match {
             case Nil => Right(updatedQuery)
             case wrongFields => Left(wrongFields)
           }
--- a/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
@@ -41,20 +41,40 @@ object DataTypes {
   val defaultLimitValue: Int = 10000
 
   /** Replaces timestamp represented as Long in predicates with one understood by the SQL */
-  private def replaceTimestampInPredicates(path: EntityPath, query: Query, metadataService: MetadataService)(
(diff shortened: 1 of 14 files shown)


src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -41,20 +41,40 @@ object DataTypes {
...
+    val newPredicates =
+      fetchedAttributes.map { maybeAttributes =>
+        query.predicates.map { predicate =>
+          maybeAttributes.flatMap { attributes =>
+            attributes.find(_.name == dropAggregationPrefixes(predicate.field)).map {
+              case attribute if attribute.dataType == DataType.DateTime =>
+                predicate.copy(set = predicate.set.map(x => new Timestamp(x.toString.toLong).toString))
+              case _ => predicate
+            }
+          }.toList
```
ivanopagano: Sorry, I'm having quite a hard time figuring out what's happening here, due to the Future[Option[List]] nesting and the interaction with the query optional values.

Could add a type annotation for all the `val`s  help understanding how the code behaves?

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -277,14 +301,58 @@ object DataTypes {
...
   }
 
+  /** Class representing snapshot */
+  case class Snapshot(field: String, value: Any) {
+
+    /** Transforms snapshot to predicate */
+    def toPredicate: Predicate =
+      Predicate(
+        field = field,
+        operation = OperationType.after,
```
ivanopagano: why is that the correct operation type?

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -277,14 +301,58 @@ object DataTypes {
...
+  /** Finds invalid snapshot fields*/
+  def findInvalidSnapshotFields(
+      query: Query,
+      entity: EntityPath,
+      metadataConfiguration: MetadataConfiguration
+  ): List[InvalidSnapshotField] =
+    query.snapshot match {
+      case Some(snapshot) =>
+        metadataConfiguration.entity(entity).flatMap { entityCfg =>
+          entityCfg.attributes.filter {
```
ivanopagano: Maybe I'm wrong here: is it filtering by value type and then looking for the right key?

Wouldn't it be simpler then to do a single `.find` that verifies both predicates, on key and value?

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
```diff
@@ -277,14 +301,58 @@ object DataTypes {
...
+      )
+  }
+
+  /** Finds invalid snapshot fields*/
+  def findInvalidSnapshotFields(
+      query: Query,
+      entity: EntityPath,
+      metadataConfiguration: MetadataConfiguration
+  ): List[InvalidSnapshotField] =
+    query.snapshot match {
```
ivanopagano: If you like to consider an alternative, you could try importing 
```scala
cats.instances.option._
cats.instances.list._

``` 

to have an extension `query.snapshot.foldMap` operation that would only need filling the mapping part, as the `List.empty` will be given for granted as the empty value for the list monoid.

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
```diff
@@ -3038,6 +3079,118 @@ class TezosDatabaseOperationsTest
 
       }
 
+      "should get the balance of an account at a specific timestamp where there are multiple entiries for given account_id" in {
```
ivanopagano: typo: "entiries"

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
```diff
@@ -3038,6 +3079,118 @@ class TezosDatabaseOperationsTest
...
+
+        val ahrs = List(
```
ivanopagano: more meaningful name? 😀

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
```diff
@@ -3038,6 +3079,118 @@ class TezosDatabaseOperationsTest
...
+          ),
+          Map(
+            "account_id" -> Some("id2"),
+            "block_id" -> Some("blockid2"),
+            "asof" -> Some(new Timestamp(2)),
+            "r" -> Some(1)
+          )
+        )
+
+      }
```
ivanopagano: Maybe you might wanna test filtering failures on invalid field choice for the snapshot query field?

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
```diff
@@ -3038,6 +3079,118 @@ class TezosDatabaseOperationsTest
...
+        } yield found
+
+        val result = dbHandler.run(populateAndTest.transactionally).futureValue
+
+        result shouldBe List(
+          Map(
+            "account_id" -> Some("id1"),
+            "block_id" -> Some("blockid2"),
+            "asof" -> Some(new Timestamp(2)),
+            "r" -> Some(1)
```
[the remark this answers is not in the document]
piotrkosecki: it's a rank in temporal query

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
piotrkosecki: added test for that

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
piotrkosecki: account history rows - changed

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
piotrkosecki: changed

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: sure, changed

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: `after` is the same as `gt`, if you noticed it has also `inverse` set to `true`.
We want to tae the latest snapshot field by rank, and combining this predicate wit ordering we get that.

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: sure, done

src/main/scala/tech/cryptonomic/conseil/generic/chain/DataTypes.scala
piotrkosecki: done

Issue #603, opened 2019-10-20
Add cycle and governance period to bakers table


anonymoussprocket:
Add cycle and period fields.

vishakh:
@ba205  The new columns should be called cycle and governance_period. These can be fetched from the block data.

Issue #600, opened 2019-10-20
Add cycle and governance period to baking and endorsement rights tables


anonymoussprocket:
Add cycle and period fields. This will allow for some interesting statistics to be presented.

Period in this case is the same as period index in the blocks table.

godenzim:
Yes, this improvement would be very helpful to us.

vishakh:
@ba205 The new columns should be called `cycle` and `governance_period`. These can be fetched from the block data.

PR #617, opened 2019-11-07
Conseil 600/603 add cycle and governance period to bakers and rights tables

ba205:
addresses #600 , #603


src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -604,26 +604,30 @@ object DatabaseConversions {
     }
   }
 
-  implicit val bakingRightsToRows = new Conversion[Id, (BlockHash, BakingRights), Tables.BakingRightsRow] {
-    override def convert(from: (BlockHash, BakingRights)): tezos.Tables.BakingRightsRow = {
-      val (blockHash, bakingRights) = from
+  implicit val bakingRightsToRows = new Conversion[Id, ((Option[Int], Option[Int], BlockHash), BakingRights), Tables.BakingRightsRow] {
+    override def convert(from: ((Option[Int], Option[Int], BlockHash), BakingRights)): tezos.Tables.BakingRightsRow = {
+      val (blockHashWithCycleAndGovernancePeriod, bakingRights) = from
```
piotrkosecki: I suggest getting rid of `_1`,`_2` etc with something like that:
```
      val ((cycle, governancePeriod, blockHash), bakingRights) = from
      bakingRights
        .into[Tables.BakingRightsRow]
        .withFieldConst(_.blockHash, blockHash.value)
        .withFieldConst(_.estimatedTime, toSql(bakingRights.estimated_time))
        .withFieldConst(_.cycle, cycle)
        .withFieldConst(_.governancePeriod, governancePeriod)
        .transform
```

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -604,26 +604,30 @@ object DatabaseConversions {
...
         .transform
     }
   }
 
-  implicit val endorsingRightsToRows = new Conversion[List, (BlockHash, EndorsingRights), Tables.EndorsingRightsRow] {
-    override def convert(from: (BlockHash, EndorsingRights)): List[Tables.EndorsingRightsRow] = {
-      val (blockHash, endorsingRights) = from
+  implicit val endorsingRightsToRows = new Conversion[List, ((Option[Int], Option[Int], BlockHash), EndorsingRights), Tables.EndorsingRightsRow] {
+    override def convert(from: ((Option[Int], Option[Int], BlockHash), EndorsingRights)): List[Tables.EndorsingRightsRow] = {
+      val (blockHashWithCycleAndGovernancePeriod, endorsingRights) = from
```
piotrkosecki: similar to what I suggested in `bakingRightsToRow`

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
```diff
@@ -155,24 +155,25 @@ trait BlocksDataFetchers {
...
     override type Encoded = String
 
-    private val makeUrl = (hash: BlockHash) => s"blocks/${hash.value}/helpers/baking_rights"
+    private val makeUrl = (blockData: (Option[Int], Option[Int], BlockHash)) => s"blocks/${blockData._3.value}/helpers/baking_rights"
 
     /** an effectful function from a collection of inputs `T[In]`
       * to the collection of encoded values, tupled with the corresponding input `T[(In, Encoded)]`
       */
-    override val fetchData: Kleisli[Future, List[BlockHash], List[(BlockHash, String)]] =
+    override val fetchData: Kleisli[Future, List[(Option[Int], Option[Int], BlockHash)], List[((Option[Int], Option[Int], BlockHash), String)]] =
```
ivanopagano: If you don't want to get mad with heavy signatures, you might want to use the alias `In` for the tuple, defined a little above

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
```diff
@@ -155,24 +155,25 @@ trait BlocksDataFetchers {
...
       Kleisli(
-        hashes => {
+        hashesWithCycleAndGovernancePeriod => {
+          val hashes = hashesWithCycleAndGovernancePeriod.map(_._3)
```
ivanopagano: I can't see how the `hashes` variable is ever used...

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
```diff
@@ -206,24 +207,25 @@ trait BlocksDataFetchers {
...
```
ivanopagano: Same as for BakingRights fetcher, is `hashes` needed?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -219,7 +202,32 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+      }
+
+      val governancePeriod = safeHead(fetchingResults).flatMap{
+        _._1.data.metadata match {
+          case GenesisMetadata => None
+          case BlockHeaderMetadata(_,_,_,_,_,level) => Some(level.voting_period)
+        }
+      }
+
+      val blockHashesWithCycleAndGovernancePeriod = bh.map(hash => (cycle, governancePeriod, hash))
```
ivanopagano: I'm not sure I follow, but this way it looks like for the whole batch of block-hashes, we always assign the cycle and voting period of the first block in the batch.
Naively I'd think that each block would need to be assigned to his own level.cycle/level.voting_period.
Hence cycle, and governancePeriod might be turned into functions to combine when mapping over all results, to get the triple.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -219,7 +202,32 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+
+      (tezosNodeOperator.getBatchBakingRights(blockHashesWithCycleAndGovernancePeriod),
+        tezosNodeOperator.getBatchEndorsingRights(blockHashesWithCycleAndGovernancePeriod)).mapN {
```
ivanopagano: Here it looks like we have to adapt the fetcher to accept the triple, but still use only the block-hash internally.

Why not fetch the rights passing only the block-hash, and then "build" the triplet only when calling the `writeXXXRights` methods?

We can, e.g., build an index to map hashes to the correct cycle/period
```scala
//define two extractor functions
val readCycle: BlockData => Option[Int] = data => ...
val readPeriod: BlockData => Option[Int] = data => ...

(getBatchBakingRights..., getBatchEndorsingRights...).mapN {
  case (br, er) =>
    (db.run(TzDB.writeBakingRights(withLevelData(br)), db.run(TzDB.writeEndorsingRights(withLevelData(er)))
      .mapN((_, _) => ())
}

//define an index and a way to attach cycle/period to rights map
//the index associated with the hash the actual level data
val levelDataIndex: Map[BlockHash, (Option[Int], Option[Int])] = fetchingResults.map {
  case (Block(data, _, _), _) =>
    data.hash -> (readCycle(data), readPeriod(data))
}.toMap

//this should convert inputs to add the missing level info, by mapping the keys and pairing them with those from the index
def withLevelData[T](tMap: Map[BlockHash, T]): Map[(Option[Int], Option[Int], BlockHash), T] = ...
```

ivanopagano: This would by-pass the need to update the rights fetchers completely.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -219,7 +202,32 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
     def processBakingAndEndorsingRights(fetchingResults: tezosNodeOperator.BlockFetchingResults): Future[Unit] = {
       import cats.implicits._
       val bh = fetchingResults.map(_._1.data.hash)
-      (tezosNodeOperator.getBatchBakingRights(bh), tezosNodeOperator.getBatchEndorsingRights(bh)).mapN {
+
+      def safeHead[A](list : List[A]): Option[A] = {
```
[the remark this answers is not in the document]
ba205: Oh wow that's really cool! Thanks :)

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
ba205: logging is based of hashes, not other metadata

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
ba205: same idea as above

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
ivanopagano: I don't follow you here. 😕 
I only see a `val hashes` defined here, but I never see any other code use it.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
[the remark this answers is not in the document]
ba205: IntelliJ lists the type of "head" to not be wrapped in a Some. For expediency, I'll leave this as is.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ba205: You're absolutely right, I'll refactor this so that each block hash is coupled with it's own cycle/period.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
ba205: See line 182.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ba205: I agree with this fundamentally, but may leave for another pr, as we have to get this functionality out soon.

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
ba205: Duh! I agree. :)

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
ba205: Same! :)

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
ba205: Sure.

sql/conseil.sql
```diff
@@ -54,30 +58,14 @@ CREATE TABLE tezos.accounts (
...
     balance numeric NOT NULL,
     block_level numeric DEFAULT '-1'::integer NOT NULL,
-    manager character varying, -- retro-compat from protocol 5+
-    spendable boolean, -- retro-compat from protocol 5+
-    delegate_setable boolean, -- retro-compat from protocol 5+
-    delegate_value char varying -- retro-compat from protocol 5+
+    manager character varying,
+    spendable boolean,
+    delegate_setable boolean,
+    delegate_value character varying
```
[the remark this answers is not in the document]
ba205: I don't understand the context. I didn't remove data in this pr.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ba205: Will refactor to abide by this soon.

sql/conseil.sql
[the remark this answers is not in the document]
ivanopagano: The data is still there, sure. 😄 
It's the comments that are missing, and they where there to warn off anyone from being tempted to remove those fields in the future.
They might look like outdated fields from the PoV of the latest protocol. But we need to keep historical memory of why they're here, to let lorre work when trying to inject the pre-babylon data.

I hope it's now more clear what I'm talking about.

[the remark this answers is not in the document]
ba205: OH, I didn't notice the comments being removed, sorry! I'll add them back in.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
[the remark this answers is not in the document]
piotrkosecki: in scala you have also `list.headOption` which returns `Option[A]`

sql/conseil.sql
```diff
@@ -96,7 +99,26 @@ CREATE TABLE tezos.accounts_checkpoint (
 
 
 --
--- TOC entry 193 (class 1259 OID 99668)
+-- Name: accounts_history; Type: TABLE; Schema: tezos; Owner: -
+--
+
+CREATE TABLE tezos.accounts_history (
```
vishakh: This is a duplicate entry. Please double check diffs in the future as there's no way this ticket could have introduced this table.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
```diff
@@ -155,24 +155,27 @@ trait BlocksDataFetchers {
     import JsonDecoders.Circe.Rights._
 
     /** the input type, e.g. ids of data */
-    override type In = BlockHash
+    override type In = (Option[Int], Option[Int], BlockHash)
```
vishakh: Instead of a tuple, could we please use a case class? It will make the code more idiot proof.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
```diff
@@ -155,24 +155,27 @@ trait BlocksDataFetchers {
...
 
     /** the output type, e.g. the decoded block data */
     override type Out = List[BakingRights]
 
     /** the encoded representation type used e.g. some Json representation */
     override type Encoded = String
 
-    private val makeUrl = (hash: BlockHash) => s"blocks/${hash.value}/helpers/baking_rights"
+    private val makeUrl = (blockData: In) => s"blocks/${blockData._3.value}/helpers/baking_rights"
```
vishakh: The `._3` is why the code needs to be made idiot proof.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
```diff
@@ -206,24 +209,27 @@ trait BlocksDataFetchers {
     import JsonDecoders.Circe.Rights._
 
     /** the input type, e.g. ids of data */
-    override type In = BlockHash
+    override type In = (Option[Int], Option[Int], BlockHash)
```
vishakh: Same deal here..please use a case class here.

PR #620, opened 2019-11-08
597 - Lorre performance improvements

### LINKED ISSUE #597: Examine Lorre for performance improvements
As Lorre is now capturing many more data types than before and for a greater number of blocks, the time for full sync has gone up. We should examine the Lorre logic, especially database writes, to identify and remedy bottlenecks and generally improve performance. 

Suggested approach:
- Run Lorre locally for around 10,000 blocks and save the logs.
- Examine the logs for a typical iteration and find the bits which seem to be taking the longest.
- Validate any findings from above and try to find remedies. 
- Try to find ways overall of reducing the number of distinct database inserts.

piotrkosecki:
resolves #597 

I was performing a lot of testing and most of the delays come from Tezos node calls and the complexity of the operations we need to make to fill some tables. 

For example fetching baking/endorsing rights takes soo long because it fetches them for whole page of blocks(500) using batched fetchers - changing concurrency level does not seem to matter, and inserts don't take a lot of time.

Regarding removing foreign keys/indexes unfortunately that only helps when inserting and affects query performance - `accounts_checkpoint` is not being queried at all so I introduced change suggested by @ivanopagano in the ticket. 

Other possible change I've noticed is inserting voting rolls using single query instead of transactionally. That can improve performance a bit, but from what I've tested many times there was nothing to insert.

If anyone has any suggestions what else could be improved feel free to comment here or in the ticket #597.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -272,27 +272,16 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
     * Fetches voting data for the blocks and stores any relevant
     * result into the appropriate database table
     */
-  private[this] def processVotesForBlocks(blocks: List[TezosTypes.Block]): Future[Option[Int]] = {
-    import cats.syntax.traverse._
-    import cats.syntax.foldable._
-    import cats.instances.list._
-    import cats.instances.option._
-    import cats.instances.int._
-    import slickeffect.implicits._
-    import slick.jdbc.PostgresProfile.api._
-
+  private[this] def processVotesForBlocks(blocks: List[TezosTypes.Block]): Future[Option[Int]] =
     tezosNodeOperator.getVotingDetails(blocks).flatMap {
       case (proposals, bakersBlocks, ballotsBlocks) =>
         //this is a nested list, each block with many baker rolls
-        val writeBakers = bakersBlocks.traverse {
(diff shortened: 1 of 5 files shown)


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -272,27 +272,16 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
-        val writeBakers = bakersBlocks.traverse {
-          case (block, bakersRolls) => TezosDb.writeVotingRolls(bakersRolls, block)
-        }
+        val writeBakers = TezosDb.writeVotingRolls(bakersBlocks)
 
-        val combinedVoteWrites = writeBakers.map(_.combineAll.map(_ + proposals.size + ballotsBlocks.size))
+        val combinedVoteWrites = writeBakers.map(_.map(_ + proposals.size + ballotsBlocks.size))
 
-        db.run(combinedVoteWrites.transactionally)
+        db.run(combinedVoteWrites)
```
ivanopagano: @piotrkosecki What would happen if the db fails in the middle?
Do you think the `Tables.Rolls ++= bakers` operation is actually atomic at the db level? Or it might lose only part of  the vote writes?

piotrkosecki: According to the comment in the slick if any part of the batch fails it throws an exception - it uses JDB~C batch update feature.

PR #622, opened 2019-11-08
581 - Simplify Lorre main body

### LINKED ISSUE #581: Refactor main body of Lorre
In order to make the core logic of Lorre more readable, the code should be refactored so the core is a small composition of multiple obviously-understandable functions.

ivanopagano:
Fixes #581 

This is essentially a basic rewrite of the Lorre app object designed to make all the processing steps more explicit and cleanly separated.

Apart from reshuffling the code there are a couple of enhancements
- using the first tezos connection trial to gain info on the current head block and use it, avoiding repeated calls for that
- reordering of the operations on db, hence requiring smaller transactions
- using cats effects to sequence all the operations, this allows the following
- more control on the real execution of remote calls, especially considering the limits we must ensure to avoid overriding the http connection pool
- more combinators for the results, which lighten up the syntax
- resource safety guaranteed by cats-effect Resources management

I ran the regression testing in `it` and everything seems to work correctly

## Database schema changes

I removed a foreign key from the accounts checkpoint, needed to reorder operations

```sql
DROP CONSTRAINT checkpoint_block_id_fkey;	
```

---
## for the reviewer
The Lorre file is essentially rewritten from scratch, so I guess that doing side-by-side comparison is quite futile.
It might be better to read the logic top to bottom on both files to compare for readability and feature parity


vishakh:
I think Lorre.scala should be refactored to be no more than 50 lines, with the code only providing a high level description of what the program does. Most of the processing logic should be moved to specific files and packages. What do you think?

ivanopagano:
> I think Lorre.scala should be refactored to be no more than 50 lines, with the code only providing a high level description of what the program does. Most of the processing logic should be moved to specific files and packages. What do you think?

I'll rework in that direction, figuring out any issue as I go. Thank you.

ivanopagano:
> I think Lorre.scala should be refactored to be no more than 50 lines, with the code only providing a high level description of what the program does. Most of the processing logic should be moved to specific files and packages. What do you think?

The latest changes are there to separate the logic in different logical modules

src/main/scala/tech/cryptonomic/conseil/application/LorreAccountsOperations.scala
```diff
@@ -0,0 +1,264 @@
...
+  BlockTagged,
+  Protocol4Delegate,
+  PublicKeyHash
+}
+import cats.Foldable
+import cats.syntax.all._
+import cats.effect.{ContextShift, IO}
+import slick.jdbc.PostgresProfile.api._
+import scala.concurrent.{ExecutionContext, Future}
+import scala.collection.SortedSet
```
piotrkosecki: unused import

src/main/scala/tech/cryptonomic/conseil/application/LorreBakerRightsOperations.scala
```diff
@@ -0,0 +1,85 @@
...
+import cats.effect.{IO, Timer}
+import scala.{Stream => _}
+import com.typesafe.scalalogging.LazyLogging
+import cats.implicits._
+import com.typesafe.scalalogging.Logger
+import scala.concurrent.ExecutionContext
+import akka.stream.ActorMaterializer
+import akka.stream.scaladsl.{Sink, Source}
+
+class LorreBakerRightsOperations(resources: LorreAppConfig.ConfiguredResources) extends LazyLogging with IOLogging {
```
piotrkosecki: why did you change name `BakingRights` -> `BakerRights`? also this class fetches endorsing rights

src/main/scala/tech/cryptonomic/conseil/application/LorreBakerRightsOperations.scala
```diff
@@ -0,0 +1,85 @@
+package tech.cryptonomic.conseil.application
+
+import tech.cryptonomic.conseil.util.IOUtils.{lift, IOLogging}
```
piotrkosecki: `lift` is unused and few lines below `Timer`

src/main/scala/tech/cryptonomic/conseil/config/LorreAppConfig.scala
```diff
@@ -61,60 +91,139 @@ trait LorreAppConfig {
...
+      }
+
+      /** Use the pureconfig convention to handle configuration from the command line */
+      def readArgs(args: List[String]): ConfigReader.Result[ArgumentsConfig] =
+        argsParser
+          .parse(args, ArgumentsConfig())
+          .toRight[ConfigReaderFailures](ConfigReaderFailures(new ThrowableFailure(new IllegalArgumentException, None)))
+
+      //applies convention to uses CamelCase when reading config fields
+      @silent("local method hint in method loadApplicationConfiguration is never used")
```
piotrkosecki: because you changed `loadApplicationConfiguration` to val  `@silent` annotations are not working, please fix them

src/main/scala/tech/cryptonomic/conseil/tezos/Tables.scala
```diff
@@ -1351,53 +1351,55 @@ trait Tables {
...
-            r(33).asInstanceOf[Option[String]].get,
-            r(34).asInstanceOf[Option[Int]].get,
-            r(35).asInstanceOf[Option[String]],
-            r(36).asInstanceOf[Option[Boolean]].get,
-            r(37).asInstanceOf[Option[Int]],
-            r(38).asInstanceOf[Option[java.sql.Timestamp]].get
-          ),
-        (_: Any) => throw new Exception("Inserting into ? projection not supported.")
-      )
+      (branch :: numberOfSlots :: cycle :: Rep.Some(operationId) :: Rep.Some(operationGroupHash) :: Rep.Some(kind) :: level :: delegate :: slots :: nonce :: pkh :: secret :: source :: fee :: counter :: gasLimit :: storageLimit :: publicKey :: amount :: destination :: parameters :: managerPubkey :: balance :: proposal :: spendable :: delegatable :: script :: storage :: status :: consumedGas :: storageSize :: paidStorageSizeDiff :: originatedContracts :: Rep
```
piotrkosecki: something I've noticed - this always differs between our PRs, looks like scalafmt works differently on our machines 😄

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -1,729 +1,167 @@
...
-              Future.successful(events)
-          }
-        } else Future.successful(events)
-      } yield updated
-    } else Future.successful(events)
-
-  /**
-    * Fetches all blocks not in the database from the Tezos network and adds them to the database.
-    * Additionally stores account references that needs updating, too
+  /** This is the core logic of Lorre, based on additional resources and services
```
piotrkosecki: I would prefer to see everything from this line below in the separate file.

src/main/scala/tech/cryptonomic/conseil/application/LorreBakerRightsOperations.scala
ivanopagano: I tried to find a naming that would generally include both baking and endorsing, so I tried to refrain from using exactly the "baking rights" word to make it more general (like operations related to what bakers do, where baker is a guy involved in baking blocks)

I'd like to find a good way to change this constant repetition of bakingEndorsingXXX

src/main/scala/tech/cryptonomic/conseil/application/LorreAccountsOperations.scala
ivanopagano: thanks

src/main/scala/tech/cryptonomic/conseil/application/LorreBakerRightsOperations.scala
ivanopagano: Thanks, I'll check it out

src/main/scala/tech/cryptonomic/conseil/config/LorreAppConfig.scala
ivanopagano: Does the annotation only work on `def`s?

src/main/scala/tech/cryptonomic/conseil/tezos/Tables.scala
ivanopagano: I've noticed something like that too, do you have any idea of where that might be coming from? 🤔

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: I'll check this out more throughly then.

src/main/scala/tech/cryptonomic/conseil/config/LorreAppConfig.scala
piotrkosecki: no, the warning is phrased differently - there is `value` instead of `method` and `$anonfunction`(if I remember correctly) instead of `loadApplicationConfiguration`

src/main/scala/tech/cryptonomic/conseil/tezos/Tables.scala
piotrkosecki: nope, no idea

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -1,771 +1,53 @@
...
+  //defines the main logic, given all necessary dependencies and configurations
+  override def run(args: List[String]): IO[ExitCode] = {
+    import LorreAppConfig.ConfiguredResources
+    import cats.arrow.Arrow
+    import cats.instances.function._
+
+    //builds operations on which Lorre depends, applying the configured resources
+    //it uses cats.arrow idioms to pair the input with the output
+    val buildDeps: ConfiguredResources => (ConfiguredResources, application.CompositeOps) =
+      Arrow[Function].lift(identity[ConfiguredResources]).merge((application.makeOperations _).tupled)
```
ivanopagano: rewrite this simply as something like: 
```
res => {
  val ops = makeOps(res)
  (res, ops)
}
```

ivanopagano: The idea here is that we need a function that builds on top of "general" resources (database connection pool, actor system, execution context, purely functional async context, ...) a set (tuple) of "specialized" operators/operation-set/services.
ConfiguredResources contains the general resources, built from configuration params.
application.CompositeOps is the set of specialized operations.
They're created from the ConfiguredResources using the `application.makeOperations` method.

The `Arrow` plumbing is only needed to make the types fit correctly, based on the assumption that `makeOperations` is to be written in the most straightforward way, without any hint to _how_ it will be used.

Here we make use of `cats.arrow.Arrow` to be able to convert the makeOperations function such that it also returns the passed-in argument as part of the output.

We'll see that this is needed later on so that both will be available later in the actual code that runs the application, which needs both general and specialized resources.

ivanopagano:
I'm adding [more than] a few comments here and there to be used as reference to
* understand the changes
* write some docs to collect the knowledge needed to understand the design from scratch even when unfamiliar with the whole cats/cats-effect approach and api

src/it/scala/tech/cryptonomic/conseil/client/EndpointClient.scala
```diff
@@ -153,35 +161,29 @@ object DataEndpointsClientProbe {
...
-        case Right(json) =>
-          Either.cond(
-            json == expected,
-            right = json,
-            left = new Exception(
-              s"Failed to match on $endpoint: expected \n${expected.spaces2} \n but found \n ${json.spaces2}"
-            )
-          )
+        case Right(received) =>
+          Diff[Json].apply(received, expected) match {
```
ivanopagano: Notes: switching to software-mill's diff allows to receive a better-detailed log of how two complex data-structure differ between expectations and test results.

src/it/scala/tech/cryptonomic/conseil/client/EndpointClient.scala
```diff
@@ -190,21 +192,95 @@ object DataEndpointsClientProbe {
...
+                s"Failed to match on $endpoint: diff result is \n${different.show}"
+              ).asLeft[Json]
+          }
+        case left => left
+      }
+    }
+
+  }
+
+  def bakingRightsQueryEndpoint = {
```
ivanopagano: Notes: Testing bakers' rights in the integration to make sure that the code redesign will not break the concurrent operations.

It's a safety measure needed to handle the fact that the scheduling is now done within cats-effect's pure system.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -1,771 +1,53 @@
...
-    }
-  }
-
-  private object DelegatesProcessor {
-
-    /* Fetches the data from the chain node and stores delegates into the data store.
-     * @param ids a pre-filtered map of delegate hashes with latest block referring to them
+    val indexingLoop: Kleisli[IO, ConfiguredResources, Unit] = Kleisli(
+      (LorreIndexing.run _).tupled
+    ).local(buildDeps)
```
ivanopagano: Notes: here we can see that in the end, the main loop is defined elsewhere and will need both general and specialized lorre's resources and operations to work (that's why we needed to compose those with the above `buildDeps` function.

The resulting Kleisli built from `LorreIndexing.run` will take both `ConfiguredResources` and `CompositeOps`, but here we "adapt" the input using the `.local` call, which somehow "tells to the Kleisli how to pre-compose the simpler input (only the ConfiguredResources) to obtain the composed one by itself.

The resulting `Kleisli` value is a "effectful function" that accepts a `ConfiguredResources` to compute the `IO` effect with no return value (`Unit` is the return type).

Such value of `IO[Unit]` is a "description of how to run a program" that will execute IO effects and eventually return no real value (Unit)

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -1,771 +1,53 @@
...
+
+    /* the bakers rights scheduled update */
+    val bakerRightsSchedule: Kleisli[IO, ConfiguredResources, Unit] = Kleisli(
```
ivanopagano: Notes: here again we build an effectful Kleisli function to build a scheduled operation from `ConfiguredResources`.

The result type will produce IO effects but won't return anything meaningful (Unit), since it's the effect of scheduling another process, which doesn't compute anything per-se.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -1,771 +1,53 @@
...
-    * @param processed how many entities were processed at the current checkpoint
-    */
-  private[this] def logProcessingProgress(entityName: String, totalToProcess: Int, processStartNanos: Long)(
-      processed: Int
-  ): Unit = {
-    val elapsed = System.nanoTime() - processStartNanos
-    val progress = processed.toDouble / totalToProcess
-    logger.info("================================== Progress Report ==================================")
-    logger.info("Completed processing {}% of total requested {}s", "%.2f".format(progress * 100), entityName)
+    val program = Parallel.parProduct(indexingLoop, bakerRightsSchedule).void
```
ivanopagano: This simply combines the Kleisli functions as two process descriptions that will run in parallel when the `program` combined IO description will be executed by the `IOApp` object

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -1,771 +1,53 @@
...
 
-    val etaMins = Duration(scala.math.ceil(elapsed / progress) - elapsed, NANOSECONDS).toMinutes
-    if (processed < totalToProcess && etaMins > 1) logger.info("Estimated time to finish is around {} minutes", etaMins)
-    logger.info("=====================================================================================")
+    //builds the actual resources and then pass the adapted program as a function to execute
+    useConfiguredResources(args)(program.run).as(ExitCode.Success)
```
ivanopagano: Notes: this combines all the prepared pieces together:
* useConfiguredResources takes the command line args to produce a set of `ConfiguredResources` (the "generic" ones) then
* passes the result to the `program.run` function, which will build the IO[Unit] value
* the `.as` will change the return value of a successful `IO` execution to the exitcode type, which is the expected return for the global `IOApp.run` function.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -1,771 +1,53 @@
...
-      _ <- if (iteration % lorreConf.feeUpdateInterval == 0)
-        FeeOperations.processTezosAverageFees(lorreConf.numberOfFeesAveraged)
-      else
-        noOp
-    } yield Some(nextRefreshes)
-
-    /* Won't stop Lorre on failure from processing the chain, unless overridden by the environment to halt.
-     * Can be used to investigate issues on consistently failing block or account processing.
-     * Otherwise, any error will make Lorre proceed as expected by default (stop or wait for next cycle)
+object Lorre extends IOApp with LazyLogging with IOLogging with LorreAppConfig {
```
ivanopagano: Notes: cats.effect.IOApp is the functional equivalent of the traditional java Main object, where the `main(...)` method is instead defined as a `run` method that is expected to return an `IO[ExitCode]` value.

Such value describes a computation to be executed (internally by the `IOApp`) where side-effects are correctly defined inside IO values.
The value contained in the returned IO can be one of an enumeration defined in `ExitCode`'s type hierarchy.

Each corresponds roughly to a Unix process exit code (0, for success, negative to signal any form of codified error)

vishakh:
Closing stale PRs. We can revive them later.

Issue #604, opened 2019-10-20
Baker flag in accounts table


anonymoussprocket:
Similar to the work in #501, add a flag to the accounts table to show if the address is baking.

vishakh:
In order to add this flag while saving down account data, we should use the data from the `/chains/main/blocks/head/votes/listings` route of the Tezos node. If a given account has an entry in the results, it's `isBaker` value should be set to true, false otherwise.

The above route gives a list of active bakers and the number of rolls they control. If an account is not mentioned there, it is not a baker. 

Please keep in mind we might already be querying that route so we should ideally not query it a second time (if possible).

ba205:
> In order to add this flag while saving down account data, we should use the data from the `/chains/main/blocks/head/votes/listings` route of the Tezos node. If a given account has an entry in the results, it's `isBaker` value should be set to true, false otherwise.

As mentioned in the call yesterday, if I understood correctly, an account becomes a baker if it has a delegation operation involving itself. An account can also have nonzero balance of rolls without being a baker.

My concern: is it possible for an account to be listed here without performing that self-delegation operation? If not, then this approach works. Otherwise, we have to check and see if a specific account is touched by that kind of operation before performing this check.

vishakh:
@ba205 I checked a few accounts using Arronax. `tz1Syu3KacZ8cy4286a4vaCeoMtwqVKHkaoj` has  22,774,503 tz but does not show up in the `rolls` results.

I therefore think we are good.

ivanopagano:
Does the flag switch on and off based on baking cycle?

PR #628, opened 2019-11-14
621 - added isBaker flag to the accounts_history

### LINKED ISSUE #621: add baker flag to accounts_history table
Similar to work in #604

piotrkosecki:
resolves #621 and #604 

## Required database changes
`accounts_history`  and `accounts` tables need one more column:
```is_baker boolean NOT NULL DEFAULT false```

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -294,10 +294,16 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
             TezosDb.updateAccountsHistoryWithBakers(bakersRolls, block)
         }
 
+        val updateAccounts = bakersBlocks.traverse {
+          case (block, bakersRolls) =>
+            TezosDb.updateAccountsWithBakers(bakersRolls, block)
+        }
+
         val combinedVoteWrites = for {
           bakersWritten <- writeBakers
-          accountsUpdated <- updateAccountsHistory
-        } yield bakersWritten.combineAll.map(_ + proposals.size + ballotsBlocks.size + accountsUpdated.size)
+          accountsHistoryUpdated <- updateAccountsHistory
+          accountsUpdated <- updateAccounts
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/main/scala/tech/cryptonomic/conseil/Lorre.scala [hunk 1]
```diff
@@ -285,10 +285,25 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+          case (block, bakersRolls) =>
+            TezosDb.updateAccountsWithBakers(bakersRolls, block)
         }
 
-        val combinedVoteWrites = writeBakers.map(_.combineAll.map(_ + proposals.size + ballotsBlocks.size))
+        val combinedVoteWrites = for {
+          bakersWritten <- writeBakers
+          accountsHistoryUpdated <- updateAccountsHistory
+          accountsUpdated <- updateAccounts
+        } yield bakersWritten.combineAll.map(_ + proposals.size + ballotsBlocks.size + accountsHistoryUpdated.size + accountsUpdated.size)
```
ivanopagano: The final sum of all data is meant to be logged in the call to `logVotingOutcome`, I'm not sure that the logging is still up-to-date with the numbers being given from here

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -88,6 +88,7 @@ object DatabaseConversions {
         blockAccountsToAccountRows.convert(from).map {
           _.into[Tables.AccountsHistoryRow]
             .withFieldConst(_.asof, Timestamp.from(from.timestamp.getOrElse(Instant.ofEpochMilli(0))))
+            .withFieldConst(_.isBaker, false)
```
ivanopagano: Is this actually needed? If the db-column defaults to false, I guess there's no need to explicitly set the value on the row scala type generated via slick

piotrkosecki: you're, right good catch

Issue #607, opened 2019-10-21
Investigate protocol 005 airdrop balance update


anonymoussprocket:
It appears that Conseil is not picking up the 1utz airdrop that occurred at the time of the upgrade. Consider these two results:

[A](https://arronax-beta.cryptonomic.tech?e=Tezos%20Mainnet/accounts&q=eyJmaWVsZHMiOlsiYWNjb3VudF9pZCIsIm1hbmFnZXIiLCJkZWxlZ2F0ZV92YWx1ZSIsImJhbGFuY2UiXSwicHJlZGljYXRlcyI6W3siZmllbGQiOiJhY2NvdW50X2lkIiwib3BlcmF0aW9uIjoiZXEiLCJzZXQiOlsidHoxV1JkV3A5NjljdG5hSFV3VXZma3JLRlV0SkFhQXZ3ajE3Il0sImludmVyc2UiOmZhbHNlfV0sIm9yZGVyQnkiOlt7ImZpZWxkIjoiYmxvY2tfbGV2ZWwiLCJkaXJlY3Rpb24iOiJkZXNjIn1dLCJhZ2dyZWdhdGlvbiI6W10sImxpbWl0IjoxMDAwfQ) vs [B](https://tzstats.com/account/tz1WRdWp969ctnaHUwUvfkrKFUtJAaAvwj17)

A shows 0.000000, B has 0.000001

ba205:
Here is a potential spec for the logic:

In Lorre.scala:
-In mainLoop function:
  -At the end of processing:
    -using db handle in Lorre.scala, run ProtocolUpgrades.isProtocol005Active
    -if protocol005 is active and !config.hasPath("protocol005Updated") 
      -run ProtocolUpgrades.updateAccountBalancesForProtocol005
      -write new value "protocol005Updated" to config, with value true

reference.conf
-New variable "protocol005BlockLevel" set to 655361.
-After protocol005 has been detected, and the account balance fix has been updated, a new value "protocol005Updated" will be written as true, so that logic is run exactly once.

ProtocolUpgrades.scala
-File under tezos namespace.
-Retrieve new "protocol005BlockLevel" variable from configuration.
-Function "isProtocol005Active": DBIO[Boolean]
  -Use TezosDatabaseOperations "maxBlockLevel" function.
  -Return whether or not "maxBlockLevel" >= protocol005BlockLevel.
-Function "fetchProtocol005UpdatedAccounts" (type?).
[...]

PR #629, opened 2019-11-18
607 - Handles forced reload of account data at given blocks like the babylon airdrop

ivanopagano:
Addresses #607 

 * create a new step in the lorre regular process that periodically checks for needed refreshs
 * add a config entry to identify the levels
 * add a db table to keep track of processed levels
 * add logic into Lorre to do the processing at each cycle
 * add database and api operations to support all of that

### how it works
Handling the missed micro-tezos for the babylon updated is handled by
 * defining configuration entries to identify custom block levels where a special processing should happen, in this case an account data reload from the node data
 * defining a new persistent table that stores the levels that actually triggered the special processing, as to make the operation an exactly-once event
 * when the process starts, we read any configuration values that identifies levels to process, and at the same time read the list of those already processed from db, and produce a diff of the twos
 * at the beginning of every lorre complete cycle, we check if there are configuration values for block levels to process from the diff of the previous step
 * if the current reached level crossed one of the configured boundaries, then corresponding block hash and level are loaded and used to populate the account's checkpoint for the next cycle step, using all account ids currently available on conseil
 * the recorded entries from the checkpoint will force a refetch of all accounts at the given level, unless a new update was included in this cycle's blocks, in which case the latest data is loaded

Adding on mainnet or any other network the correct level for the babylon switch, should reload all accounts' data, hence correcting the balances to account for the micro-tezos airdrop.

## Changes to the database
We have a new table to add, with no indexing, as the cardinality expected is really low

```sql
CREATE TABLE tezos.processed_chain_events (
    event_level numeric PRIMARY KEY
);
```

## Changes to the .conf files
The reference conf (used as base for any application) introduces new sub-entry in the `lorre` entry.

To define the correct upgrade level in the applied lorre.conf add the following entry (considering the estimated block update on mainnet)
```coffee
chainEvents {
  accountsRefreshLevels: [655361]
}
```


ivanopagano:
TODO: I need to add some unit-tests before finalising the draft

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -151,11 +166,41 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+   * @param epochs the relevant epochs that calls for a refresh
+   */
+  private def processAccountRefreshes(epochs: SortedSet[Int]): Future[SortedSet[Int]] =
+    if (epochs.nonEmpty) {
+      for {
+        storedHead <- apiOperations.fetchMaxLevel
+        updated <- if (epochs.exists(_ <= storedHead)) {
+          val (past, toCome) = epochs.partition(_ <= storedHead)
+          logger.info(
+            "A block level was reached requiring to update the accounts data. Relevant block levels are {}",
```
vishakh: Can we change this to: The current block, $block, requires an update of account data as specified in the config file. A full refresh is now underway.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -218,6 +218,32 @@ object TezosDatabaseOperations extends LazyLogging {
...
     Tables.AccountsCheckpoint.distinctOn(_.accountId).length.result
 
+  /**
+    * Takes all existing account ids and puts them in the
+    * checkpoint to be later reloaded, based on the passed block reference
+    */
+  def refillAccountsCheckpointFromExisting(hash: BlockHash, level: Int, timestamp: Instant)(
+      implicit ec: ExecutionContext
+  ): DBIO[Option[Int]] = {
+    logger.info("Relaoding all ids for existing accounts and checkpointing with block hash {}, level {} and time {}",
```
vishakh: Could we change this to the following?

"Fetching all ids for existing accounts and checkpointing them with block hash {}, level {} and time {}"

src/main/resources/reference.conf
```diff
@@ -84,6 +84,11 @@ lorre {
   GETResponseEntityTimeout: 90 s
   POSTResponseEntityTimeout: 1 s
 
+  # defines custom updates to be run at specific epochs (i.e. upon reaching a given level)
+  chain-epoch-updates {
```
vishakh: `epoch` is a dangerous word to use as it has specific meanings in blockchain protocols. Could we please remove all instances of it in these code changes? Over here, I would suggest something like:

```
account-refreshes {
  account-refresh-levels: []
}
```

I would recommend renaming the table to something like `account_refresh_levels`.

src/main/resources/reference.conf
```diff
@@ -84,6 +84,11 @@ lorre {
...
+    accounts-refresh-epochs: []
```
vishakh: Can we also accept a SQL pattern which we can use to filter accounts which are refreshed, e.g. `tz1%` or `KT19EnJqbfWuQ6oWAZZWLjBzibNMqMFGPcW2`? This will make this new functionality a lot more useful.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
vishakh: There should also be a log message printed when the refresh is complete.

ivanopagano: I changed the sentence to mostly correspond to your suggestions.
I'm not keen on referring to a "current block", because the check will be made between lorre cycles (*), meaning that we might have already passed the relevant block by a significant margin (and even more than one actually, but only the most recent refresh will take effect).

(*) this additionally implies that the refresh effect will not be generally apparent until a full cycle is completed, which might be a long run when bootstrapping from blank.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: done

src/main/resources/reference.conf
ivanopagano: Ouch, I didn't know. I was trying to find some significant naming to identify a specific point in time... I'll work on that

src/main/resources/reference.conf
ivanopagano: Absolutely... I suggest I might work on that on next PR

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: I will add

sql/conseil.sql
```diff
@@ -41,6 +41,10 @@ $$;
 
 SET default_with_oids = false;
 
+CREATE TABLE tezos.processed_chain_events (
+    event_level numeric PRIMARY KEY
```
vishakh: As discussed, there should be an additional column here to state the event type.

ivanopagano: I'm almost done with the changes

Issue #619, opened 2019-11-08
accounts_history table updates


anonymoussprocket:
- [x] remove `manager` field
- [x] remove `delegate_setable` field
- [x] remove `spendable` field
- [ ] remove `script` - it will never change
- [x] `counter` doesn't seem to update
- [ ] add `cycle` field
- [ ] update metadata to read "Account History" for both `displayName` and `displayNamePlural`

vishakh:
@anonymoussprocket Changing a lot of these fields will make `accounts_history` inconsistent with `accounts`. Is this intentional?

anonymoussprocket:
This table contains changes. `script` can never change once deployed.

`manager`, etc are deprecated, they will never change going forward and it would be fine to trash the changes from pre-P005 era.

`isBaker` has a ticket for the accounts table: #604.

vishakh:
@anonymoussprocket  The original intention was to test the new table and then kill `accounts` once `accounts_history` was proven to be good, with the further intention of making certain other tables temporal as well. If we change the shape of the new table, this might not be possible. Is tat OK with you?

Regarding deprecated fields, they do capture data that was available on the blockchain. So far we have tried to capture Tezos RPC data as expansively as possible. This would be a change in philosophy. Is that OK with you? If a future protocol upgrade further changes the way accounts work, would we then retrospectively delete some of the fields active now?

vishakh:
@anonymoussprocket `isBaker` potentially changes the architecture of the application so it should be done as a dedicated ticket.

vishakh:
As discussed with @anonymoussprocket, we are going ahead with the changes he specified.

PR #630, opened 2019-11-19
Feature/619 accounts history table updates

piotrkosecki:
- [x] remove `manager` field
- [x] remove `delegate_setable` field
- [x] remove `spendable` field
- [x] remove `script` - it will never change
- [x] ~~`counter` doesn't seem to update <-- it seems to be working fine, can you give an example @anonymoussprocket ?~~
- [x] add `cycle` field
- [x] update metadata to read "Account History" for both `displayName` and `displayNamePlural`

## Required database changes
As expected above for the `accounts_history` table:
```
- manager
- delegate_setable
- spendable
- script
+ cycle integer
```

### DIFF
--- a/sql/conseil.sql
+++ b/sql/conseil.sql
@@ -65,15 +65,12 @@ CREATE TABLE tezos.accounts_history (
     account_id character varying NOT NULL,
     block_id character varying NOT NULL,
     counter integer,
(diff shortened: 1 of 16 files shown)


piotrkosecki:
resolved #619

PR #646, opened 2019-12-10
602 - changes to future rights fetching

piotrkosecki:
Those are additional changes requested by @vishakh.
I was testing this functionality and I've noticed that estimated time can be missing - that's why I've introduced change with making `estimated_time` columns nullable.

## Required DB changes
`estimated_time` in both tables `baking_rights` and `endorsing_rights ` must be changed to nullable

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -144,11 +144,11 @@ lorre {
 
   # Configuration for fetching upcoming baking and endorsing rights
   blockRightsFetching {
-    initDelay: 2 minutes
-    interval: 60 minutes
-    cyclesToFetch: 5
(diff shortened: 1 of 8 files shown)


src/main/resources/reference.conf
```diff
@@ -144,11 +144,11 @@ lorre {
 
   # Configuration for fetching upcoming baking and endorsing rights
   blockRightsFetching {
-    initDelay: 2 minutes
-    interval: 60 minutes
-    cyclesToFetch: 5
-    cycleSize: 4096
-    fetchSize: 200
+    initDelay: 2 minutes # time before starting fetching rights
```
ivanopagano: ❤️

PR #657, opened 2019-12-16
501 - adding is_activated to accounts and accounts_history tables

### LINKED ISSUE #501: Improve accounts dataset

piotrkosecki:
resolves #501 

## Necessary DB changes
tables `accounts` and `accounts_history` need one more column
```
is_activated boolean NOT NULL DEFAULT false
```


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -582,13 +583,49 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+        } yield res
+      }.andThen(logWriteFailure)
+
+      def updateTaggedAccountsWithIsActivated(
+          taggedAccounts: List[BlockTagged[AccountsIndex]],
+          activatedOperations: Map[Int, Seq[Tables.OperationsRow]],
+          activatedAccountIds: List[String]
+      ): List[BlockTagged[Map[AccountId, Account]]] =
+        taggedAccounts.map { taggedAccount =>
+          val activatedAccountsHashes = activatedOperations.get(taggedAccount.blockLevel).toSeq.flatten.map(_.pkh)
```
ivanopagano: Is it necessary to cast the Option with `toSeq`? Isn't there the implicit conversion to flatten it directly with its content?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -582,13 +583,49 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+          taggedAccount.copy(
+            content = taggedAccount.content.mapValues { account =>
+              val hash = account.manager.map(_.value)
+              if (activatedAccountsHashes.contains(hash) || activatedAccountIds.contains(hash.getOrElse(""))) {
```
ivanopagano: Can't we convert the activated container to only have one type of values, so that here we only make a single check? Each of these verifications might be scanning the whole sequence of activations...
I mean something similar to first lowercasing a string, instead of checking both for lower and upper case versions.

Moreso, if the accounts are not just a few, it might be useful to first prepare the operations-map into something directly useful to all checks, as a first step. E.g. make it a Map[Int, Set[Hash]] or similar.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -582,13 +583,49 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+            }
+          )
+        }
+
+      def fetchActivatedOperationsByLevel(levels: List[Int]): Future[Map[Int, Seq[Tables.OperationsRow]]] = {
+        import slick.jdbc.PostgresProfile.api._
+        db.run {
+          DBIO.sequence {
+            levels.map { level =>
+              TezosDb.fetchRecentOperationsByKind(Set("activate_account"), level).map(x => level -> x)
```
ivanopagano: Would it be an improvement if this function already returned the pkhs instead of a whole row?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -666,6 +666,53 @@ object TezosDatabaseOperations extends LazyLogging {
...
+  def activateAccount(id: String): DBIO[Int] =
+    Tables.Accounts
+      .findBy(_.accountId)
+      .extract(id)
+      .map(_.isActivated)
+      .update(true)
+
+  /** Updates an account history by id and asof, marking it with a flag.
+    * @param id the account identifier on database (corresponds to the pkh)
+    * @param asof which flag to set "on"
```
ivanopagano: These seem to be outdated?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -666,6 +666,53 @@ object TezosDatabaseOperations extends LazyLogging {
...
+    * @return the action to run against the db, which will tell how many rows where touched
+    */
+  def activateAccountHistory(ids: List[String], asof: java.sql.Timestamp): DBIO[Int] =
```
ivanopagano: Styl-ish: possibly `activateAccountInHistory`? 😄

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -666,6 +666,53 @@ object TezosDatabaseOperations extends LazyLogging {
...
+    Tables.AccountsHistory
+      .filter(acc => acc.asof >= asof && acc.accountId.inSet(ids))
+      .map(_.isActivated)
+      .update(true)
+
+  /** Finds activated accounts - useful when updating accounts history
+    * @return sequence of activated account ids
+    */
+  def findActivatedAccounts: DBIO[Seq[String]] =
```
ivanopagano: What about `findActivatedAccountIds`?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -666,6 +666,53 @@ object TezosDatabaseOperations extends LazyLogging {
...
+      .map(_.accountId)
+      .result
+
+  /** Load all operations referenced from a block level and higher, that are of a specific kind.
+    * @param ofKind a set of kinds to filter operations, if empty there will be no result
+    * @param fromLevel the lowest block-level to start from, zero by default
+    * @return the matching operations, sorted by ascending block-level
+    */
+  def fetchRecentOperationsByKind(
+      ofKind: Set[String],
```
ivanopagano: Do we have typed enumerations of the existing operation kinds somewhere, to use instead of treacherous strings? 😁

piotrkosecki: nope, I did not find that, we have a trait which extends case classes

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: sure, changed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: removed this method because after change in logic it was no longer of use

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: removed

src/main/scala/tech/cryptonomic/conseil/Lorre.scala

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: nope, unfortunately it does not work this way

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: made a change to make one check on a single list

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -582,13 +582,49 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+            content = taggedAccount.content.mapValues { account =>
+              val hash = account.manager.map(_.value)
+              if ((activatedAccountsHashes ::: activatedAccountIds.map(Some(_))).contains(hash)) {
+                account.copy(isActivated = Some(true))
+              } else account
+            }
+          )
+        }
+
+      def fetchActivatedOperationsByLevel(levels: List[Int]): Future[Map[Int, Seq[Option[String]]]] = {
```
vishakh: `fetchActivatedOperationsByLevel` => `fetchActivationOperationsByLevel`

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -666,6 +666,43 @@ object TezosDatabaseOperations extends LazyLogging {
     }
   }
 
+  /** Updates an account by id, marking it with a flag.
+    * @param id the account identifier on database (corresponds to the pkh)
+    * @return the action to run against the db, which will tell how many rows where touched
+    */
+  def activateAccount(id: String): DBIO[Int] =
```
vishakh: Do we still need this functions? If not, please remove it.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -582,13 +582,49 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
       def processAccountsPage(
           taggedAccounts: List[BlockTagged[AccountsIndex]],
           taggedDelegateKeys: List[BlockTagged[DelegateKeys]]
-      ): Future[(Option[Int], Option[Int], List[BlockTagged[DelegateKeys]])] =
-        db.run(TezosDb.writeAccountsAndCheckpointDelegates(taggedAccounts, taggedDelegateKeys))
-          .map {
+      ): Future[(Option[Int], Option[Int], List[BlockTagged[DelegateKeys]])] = {
+        for {
+          activatedOperations <- fetchActivatedOperationsByLevel(taggedAccounts.map(_.blockLevel).distinct)
+          activatedAccounts <- db.run(TezosDb.findActivatedAccountIds)
```
vishakh: This might become very expensive if the number of accounts grows a lot. Can we put in an index for now?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: renamed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: good catch, it wasn't needed

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: sure, added

PR #658, opened 2019-12-17
521 - post process ballots and proposal data 

piotrkosecki:
Created PR with adjusted changes from @ba205 but now points to `master` so the changes are more visible.
Updated with master and refactored.


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -450,6 +462,107 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+
+    logger.info("Processing latest Tezos votes data...")
+
+    /**
+      * * Helper function, takes a sequence of GovernanceData, separates them by ballot types.
+      * @param votingData
+      * @return
+      */
+    def partitionByVote(votingData: Seq[VotingData]): (Seq[VotingData], Seq[VotingData], Seq[VotingData]) = {
+      val (yays, notYays) = votingData.partition(x => x.ballot.contains("yay"))
```
ivanopagano: I guess here you can `groupBy` and collect the values of the returned map in a single pass

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -450,6 +462,107 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+    }
+
+    /**
+      * * Helper function, takes a sequence of GovernanceData, calculates stake.
+      * @param votingData
+      * @return
+      */
+    def calculateStake(votingData: Seq[VotingData]): Future[BigDecimal] =
+      Future
+        .sequence(
```
ivanopagano: `.traverse` please... 😁

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -450,6 +462,107 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+          votingData.map {
+            case VotingData(_, _, hash, _, _, _, address) =>
+              tezosNodeOperator.getAccountBalanceForBlock(BlockHash(hash), AccountId(address.get))
```
ivanopagano: What's `address` here? What if `.get` fails?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -450,6 +462,107 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+      val sampleVotingField = votingData.head
+      val timestamp = sampleVotingField.timestamp
+      val cycle = sampleVotingField.cycle
+      val level = sampleVotingField.level
+      val proposalHash = sampleVotingField.proposalHash
+      val (yays, nays, passes) = partitionByVote(votingData)
+      val yayCount = yays.length
+      val nayCount = nays.length
+      val passCount = passes.length
+      for {
```
ivanopagano: Here I guess we can profit from starting the Futures immediately and not within the for comprehension, to parallelize work and get results a bit faster

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -450,6 +462,107 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+          passStake,
+          totalStake
+        )
+    }
+
+    //get fields from db in Tezos Database Operations, use the block levels as input
+    db.run(TezosDb.fetchVotingBlockLevels).flatMap { levels =>
+      Source
+        .fromIterator(() => levels.toIterator)
+        .grouped(batchingConf.blockPageSize) // we use here block page size, because
```
ivanopagano: the comment seems half-baked?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -450,6 +462,107 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+          val computeAndStore = for {
+            listOfVotingFields <- TezosDb.fetchVotingData(blockLevels.toSet)
+            ballotOperationsByClock = listOfVotingFields.groupBy(_.level).values
+            votes <- DBIO.from {
+              Future.traverse(ballotOperationsByClock.toSeq)(processVoteAggregate)
+            }
+            dbWrites <- TezosDb.writeVotes(votes.toList)
+          } yield dbWrites
+
+          db.run(computeAndStore).andThen {
```
ivanopagano: Might we wanna wait to fold all iterated levels in a single result here?

Failure cases might need to be handled consistently to get reasonable results even if one "batch" of data fails to save... what happens in such case?
The stream will probably fail... I guess.

Just thinking out my worries, not saying that it's not correct as it is now... I'm just looking for a clear idea of how we keep under control the failures.

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
```diff
@@ -80,6 +79,19 @@ class ApiOperations extends DataOperations with MetadataOperations {
     optionalMax.map(_.getOrElse(-1))
   }
 
+  /**
+    * Fetches the level of the most recent votes stored in the database.
+    *
+    * @return Max level or -1 if no votes were found in the database.
+    */
+  def fetchVotesMaxLevel()(implicit ec: ExecutionContext): Future[Int] = {
+    val optionalMax: Future[Option[Int]] = runQuery(Tables.Votes.map(_.level).max.result)
```
ivanopagano: just wondering... is it possible to `max.getOrElse(-1).result` directly in the query?

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -49,6 +49,23 @@ object DatabaseConversions {
 
   //implicit conversions to database row types
 
+  implicit val votesToVotesRow = new Conversion[Id, VoteAggregates, Tables.VotesRow] {
+    override def convert(from: VoteAggregates) =
+      Tables.VotesRow(
+        timestamp = from.timestamp,
+        cycle = from.cycle,
+        level = Some(from.level),
+        proposalHash = from.proposalHash.get,
```
ivanopagano: Is this `.get` safe?

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
ivanopagano: By the way I guess we should have some code constant to describe `-1` as a value instead of a "magic number"

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -592,6 +603,54 @@ object TezosDatabaseOperations extends LazyLogging {
...
+      ballot: Option[String],
+      address: Option[String]
+  )
+
+  /** Get necessary fields to populate votes table from ballot operations */
+  def fetchVotingData(levels: Set[Int])(implicit ec: ExecutionContext): DBIO[Seq[VotingData]] =
+    Tables.Operations
+      .filter(_.kind inSet Set("ballot"))
+      .filter(_.blockLevel inSet levels)
+      .result
```
ivanopagano: I'm not sure it will perform better, but I think we can directly map to the fields of interest (project the query result) and then use slick's api to get directly the VotingData objects in the `.result` outcome

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -130,6 +130,16 @@ class TezosNodeOperator(
       .runAsyncGetQuery(network, s"blocks/${blockHash.value}/context/contracts/${accountId.id}")
       .map(fromJson[Account])
 
+  /**
+    * Given an account's address, get it's balance/stake.
```
ivanopagano: typo: it's -> its

ivanopagano: The doc comments are no longer up-to-date with the method signature

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -79,6 +80,17 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
     apiOperations
   )
 
+  // starts processing Tezos votes
+  system.scheduler.scheduleOnce(1.minute)(processTezosVotes())
+
+  /** Recursive process to start processing Tezos votes */
+  @tailrec
+  def processVotes: Future[Done] = {
```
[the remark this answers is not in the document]
piotrkosecki: yes, you're right, that was leftover from my changes - replaced it with unit

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -450,6 +462,107 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
     }
   }
 
+  /**
+    * Collect data from Blocks Table, Operations Table, and the Tezos RPC,
+    * return the appropriate list of VoteAggregates, and write them to the Votes table.
+    * @param ex the needed ExecutionContext to combine multiple database operations
```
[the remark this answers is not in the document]
piotrkosecki: updated

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: used groupBy and foldLeft, not sure what did you mean by using collect?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: that's the part about which I'm guessing that the vote should always have address, but it was fetched from the operations table, but there address is optional

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: ok, changed

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: yes, finished it 😄

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: yes, but when it starts again it starts from the max votes level

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
piotrkosecki: good point, changed

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
piotrkosecki: I guess its situation as above

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
piotrkosecki: you're right, changed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: I picked the columns I needed into columns, but I couldn't map them into a case class before the result because those were `Rep[]`, I changed it a bit, tell me if it's ok with you

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: Oh, I meant simply  `groupBy(...).values` 😁

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: I see

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: We just need to make sure that we don't have half-processed "in-flight" batches that will not be re-processed at next restart. 👍

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -493,18 +492,21 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
       */
     def calculateStake(votingData: Seq[VotingData]): Future[BigDecimal] =
       Future
-        .sequence(
+        .traverse(
```
ivanopagano: oops, I thought it might become something like
```scala
Future.traverse(votingData){
  case VotingData(...) =>
    tezosNodeOperator.getAccountBalanceForBlock(BlockHash(hash), AccountId(...))
}.map(...)
```?

File: src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala [hunk 14]
```diff
@@ -42,6 +42,7 @@ object ApiOperations {
 class ApiOperations extends DataOperations with MetadataOperations {
   import ApiOperations._
   lazy val dbReadHandle: Database = DatabaseUtil.conseilDb
+  private val MAX_NOT_FOUND = -1
```
ivanopagano: Just to be sure, there should be another constant like this already defined, but it might be in another file, so probably redefining the same here might help keep things properly decoupled...

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: I guess the idea was that you can call `mapTo[VotingData]` as the last operation using slick support for case classes

see https://scala-slick.org/doc/3.3.0/gettingstarted.html#case-class-mapping

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: I'm not sure if I understand your approach correctly, but I fee like it would fail on edge cases

src/main/scala/tech/cryptonomic/conseil/tezos/ApiOperations.scala
piotrkosecki: oh, yes it is in `TezosDatabaseOperations`

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: but from what I understand you'll need to define the mapping beforehand, right?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: oh, I get it, changed 😄

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: do you think adding `.transactionally` would do the job?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: I think you should be able to call that right after the second `.filter`? But I'm not totally sure with that.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: I think I didn't properly evaluate what the algorithm is doing, let's ignore my comments.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: I think there might not be anything more needed, my goal was to make sure that corner cases and failures were accounted for

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: either we have something different in mind or it does not work the way you're suggesting

ivanopagano: It's entirely possible that I'm wrong, so you can leave it at that if it doesn't work.

vishakh:
Closing this PR for now as it might now be out of date relative to the latest spec.

PR #625, opened 2019-11-12
598 - migration to testconteiners

dorians:
fixes #598 (maybe)

PR #664, opened 2019-12-30
Chore: upgrade postgres version on tests and local docker to 11.x

ivanopagano:
Relates to #625 
Since we've now switched to test-containers for db integration tests, we can now upgrade local docker definitions and the tests versions of postgres to `11.x`, more in line with production.

### DIFF
--- a/src/test/scala/tech/cryptonomic/conseil/tezos/InMemoryDatabase.scala
+++ b/src/test/scala/tech/cryptonomic/conseil/tezos/InMemoryDatabase.scala
@@ -14,7 +14,15 @@ trait InMemoryDatabase extends BeforeAndAfterAll with BeforeAndAfterEach {
   self: TestSuite =>
   import slick.jdbc.PostgresProfile.api._
 
-  val dbInstance = new PostgreSQLContainer("postgres:11.6-alpine")
+  //#JavaThankYou
+  val dbInstance =
+    new PostgreSQLContainer("postgres:11.6")
+      .asInstanceOf[PostgreSQLContainer[_]]
+      .withInitScript("in-memory-db/init-script.sql") //startup will prepare the schema
+      .asInstanceOf[PostgreSQLContainer[_]]
+      .withCommand("-c full_page_writes=off") //should improve performance for the tests
+      .asInstanceOf[PostgreSQLContainer[_]]
+
   dbInstance.start()
 
   /** how to name the database schema for the test */
@@ -66,8 +74,6 @@ trait InMemoryDatabase extends BeforeAndAfterAll with BeforeAndAfterEach {
   override protected def beforeAll(): Unit = {
     super.beforeAll()
     dbInstance.start()
-    Await.result(dbHandler.run(sql"SET client_encoding = 'UTF8'".asUpdate), 1.second)
-    Await.result(dbHandler.run(sql"CREATE SCHEMA IF NOT EXISTS tezos".as[Int]), 1.second)
     Await.result(dbHandler.run(dbSchema.create), 1.second)
   }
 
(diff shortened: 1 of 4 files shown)


src/test/scala/tech/cryptonomic/conseil/tezos/InMemoryDatabase.scala
```diff
@@ -87,8 +79,9 @@ trait InMemoryDatabase extends BeforeAndAfterAll with BeforeAndAfterEach {
   }
 
   override protected def beforeEach(): Unit = {
-    Await.ready(dbHandler.run(truncateAll), 1.second)
     super.beforeEach()
+    Await.ready(dbHandler.run(truncateAll), 1.second)
+    ()
```
dorians: Isn't this line useless?

ivanopagano: Essentially yes, but it removes a warning about discarding non-unit values

src/test/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperationsTest.scala
```diff
@@ -2361,8 +2362,8 @@ class TezosDatabaseOperationsTest
 
         val result = dbHandler.run(populateAndTest.transactionally).futureValue
         result shouldBe List(
```
piotrkosecki: that does not look right, we are ordering by hash descending and after change of postgres version it works different way than expected

ivanopagano: It depends on expectations for String sorting.
Do we expect all upper-case chars to come first or do we mix upper/lower?

My understanding is that it all depends on the encoding done by postgres for comparison of chars, which is decided usually on the database setup.

I simply experienced that the test would work on a previous version of postgres container and would break on the 11.x series, giving inverted ordering.
Therefore I tried to figure out how those differ wrt sorting rows, but I only found that the sorting depends on COLLATION of the column

>Character-string data is sorted according to the collation that applies to the column being sorted. That can be overridden at need by including a COLLATE clause in the expression, for example ORDER BY mycolumn COLLATE "en_US". For more information see Section 4.2.10 and Section 23.2.

If I get it correctly we use UTF-8 compatible sorting based on characters' code points.

Whatever the solution, we must ensure that the database ORDER BY works correspondingly and consistently on every deployment/environment.

See: https://www.postgresql.org/docs/11/collation.html (§23.2.2)
https://www.postgresql.org/docs/11/sql-expressions.html#SQL-SYNTAX-COLLATE-EXPRS

piotrkosecki: If I remember correctly I was fixing this inconsistency issue some time ago(it either worked on travis or locally) - that's why in previous version of `InMemoryDatabase` we had something like this:
```
  /* turns off anti-corruption guarantees settings that will improve performance on testing
   * override to change or add test-specific settings
   */
  protected val pgInitParams = List("--nosync", "--lc-collate=C")
```

ivanopagano: The `nosync` option is already set by default by testcontainers.

I will see how to restore the collation option too, to make the behaviour like before

ivanopagano: Fixed by avoiding the 'alpine' based image.
Configuration parameters might have changed names between versions, e.g. `nosync`

Issue #656, opened 2019-12-15
account_history missing some events


anonymoussprocket:

vishakh:
@anonymoussprocket Could you please provide more details so we can verify whether this issue was fixed?

PR #666, opened 2020-01-02
662 - fix the incorrect update of balances with chain events

### LINKED ISSUE #662: The order in which account refresh records are written seems to be wrong
From our staging environment..

Observe:

```
conseil=# select block_level, balance from tezos.accounts where account_id='tz1N3LYhafWJUc2cfmTnRzJgpQmp82WUXmKF';
 block_level | balance 
-------------+---------
      655361 | 1993731
(1 row)

conseil=# select block_level, balance from tezos.accounts_history where account_id='tz1N3LYhafWJUc2cfmTnRzJgpQmp82WUXmKF' order by block_level desc;
 block_level |  balance   
-------------+------------
      687014 |          0
      655361 |    1993731
      485110 |    1993731
      373255 |          0
      371998 | 1880116290
      316302 |          0
      220026 |      47311
      219738 | 1999950000
      206086 |          0
      204062 | 1000900001
      200766 |     950001
      175246 |          1
      172648 |    9742999
         677 | 7800000000
(14 rows)
```

It seems the account refresh record from an older block overwrites the data from a newer block in the `accounts` table. Therefore, there seems to be a bug in the order in which we write records.

ivanopagano:
Fixes #662 

Essentially the timing with which the chain event update requests where "stored" as checkpoint entries conflicted with the ordering of events, since they've been overwriting fresher data which was just updated by the previous cycle using in-memory data.

The current solution drops accounts and delegates refreshes from the checkpoint tables, if newer data is available.
This only applies for checkpoint-read data, and not from the in-memory data updates, to reduce the performance impact, since the verification requires a number of extra database queries to be made at each cycle.


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -685,24 +721,37 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+        e => {
           val error = "I failed to fetch delegates from client and update them"
           logger.error(error, e)
-          Failure(DelegatesProcessingFailed(message = error, e))
-        case success => Success(Done)
-      }
+          DelegatesProcessingFailed(message = error, e)
+        }
+      )
+      // (saveDelegates flatTap cleanup).transform {
```
piotrkosecki: why this part is commented?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -608,8 +611,19 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
           .map(cleaned => logger.info("Done cleaning {} accounts checkpoint rows.", cleaned))
       }
 
+      //if needed, we get the stored levels and only keep updates that are more recent
+      def prunedUpdates(): Future[Map[AccountId, BlockReference]] =
+        if (onlyProcessLatest) db.run {
+          val selection = ids.map(_._1).toSet
```
piotrkosecki: you can replace `ids.map(_._1).toSet` with `ids.keySet`

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: nice catch!

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: yes!

ivanopagano:
> Left few comments, but looks good. Do you think your fix is related to #656 ?

Doesn't seem probable... the bug actually stored the history as it came, yet the current value was overwritten by previous changes.

Unless the idea that history entries were missing was incorrect, and the entries were simply incorrectly ordered when applied, giving the illusion that some history entry were not there.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: yup!

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -471,6 +471,20 @@ object TezosDatabaseOperations extends LazyLogging {
 
   }
 
+  /** Fetch the latest block level available for each account id stored */
+  def getLevelsForAccounts(ids: Set[AccountId]): DBIO[Seq[(String, BigDecimal)]] =
+    Tables.Accounts
+      .map(table => (table.accountId, table.blockLevel))
+      .filter(_._1 inSet ids.map(_.id))
```
vishakh: Do we not have to put in some max condition? Maybe I am missing something but how do we guarantee we are getting the latest block level for each account?

ivanopagano: The `accounts` table updates the rows - unlike the accounts_history - so we should only have a single entry per id/level pair.

PR #668, opened 2020-01-07
565 - added period to all kinds of operations

### LINKED ISSUE #565: Operations: Period always empty
It looks like that for `operations` the `period` column is alway empty I would expect for ballots cast in the current period a return of `19`.

piotrkosecki:
resolves #565 

Period was added previously only to the ballot operations. Now it will be in all of them.


vishakh:
@piotrkosecki `period` has a specific meaning for ballot operations and is pulled from the operations payload from the Tezos RPC data. Here, we are now mixing this data with 'voting_period' from the blocks payload of the  Tezos RPC data. Let's not mix the two. Please create a new column called `voting_period` in the `operations` table and ensure we do not change the contents of the `period` field in any way.

@ivanopagano @anonymoussprocket FYI, so you also understand the differences in these fields.

piotrkosecki:
> @piotrkosecki `period` has a specific meaning for ballot operations and is pulled from the operations payload from the Tezos RPC data. Here, we are now mixing this data with 'voting_period' from the blocks payload of the Tezos RPC data. Let's not mix the two. Please create a new column called `voting_period` in the `operations` table and ensure we do not change the contents of the `period` field in any way.
> 
> @ivanopagano @anonymoussprocket FYI, so you also understand the differences in these fields.

@anonymoussprocket is this what you've expected in this ticket? because that would mean `period` column would still be empty if the operation was not a ballot

anonymoussprocket:
period/cycle/block is a a temporal hierarchy in Tezos. If an operation has a block level, it therefore is part of some cycle and it is part of some period. Use the blocks table for reference.

vishakh:
@anonymoussprocket @piotrkosecki Shall we rename the existing `period` column to `ballot_period` and use `period` for the new column?

PR #669, opened 2020-01-09
597 - Micro optimize database writes of accounts and delegates

ivanopagano:
Related to #597 

Minor tweaks in the code to improve and simplify bulk insert/upserts to accounts/delegates-related tables.

The resulting throughput impact is quite negligible on the scale of observed load, the code quality should be a bit improved nonetheless.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -428,12 +428,16 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
         //this is a nested list, each block with many baker rolls
         val writeBakers = TezosDb.writeVotingRolls(bakersBlocks)
 
-        val updateAccountsHistory = bakersBlocks.traverse {
+        val updateAccountsHistory = bakersBlocks.filterNot {
+          case (_, rolls) => rolls.isEmpty
+        }.traverse {
           case (block, bakersRolls) =>
             TezosDb.updateAccountsHistoryWithBakers(bakersRolls, block)
         }
 
-        val updateAccounts = bakersBlocks.traverse {
+        val updateAccounts = bakersBlocks.filterNot {
+          case (_, rolls) => rolls.isEmpty
+        }.traverse {
           case (block, bakersRolls) =>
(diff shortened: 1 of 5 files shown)


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -428,12 +428,16 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
         //this is a nested list, each block with many baker rolls
         val writeBakers = TezosDb.writeVotingRolls(bakersBlocks)
 
-        val updateAccountsHistory = bakersBlocks.traverse {
+        val updateAccountsHistory = bakersBlocks.filterNot {
```
piotrkosecki: if you look at my PR you'll see that this will be removed as I changed the logic a bit

ivanopagano: I already guessed that, but I kept things independent to avoid coupling separate PRs...

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -49,13 +50,21 @@ object TezosDatabaseOperations extends LazyLogging {
     */
   def writeAccounts(
       accountsInfo: List[BlockTagged[Map[AccountId, Account]]]
-  )(implicit ec: ExecutionContext): DBIO[Int] = {
+  ): DBIO[Option[Int]] = {
+    import CustomPostgresProfile.api._
+
+    val keepMostRecent = (rows: List[Tables.AccountsRow]) =>
```
vishakh: If the purpose of the PR is how change *how* data is written but not *what* is written and in which order, how come we have to do this additional processing to figure out `keepMostRecent`?

ivanopagano: Because previously we sequentially updated each row, so keeping the right order would guarantee that overwrites were correct.
Now we're trying to optimise with a bulk upsert, so we don't want to have duplicates ids in the bulk query.

PR #671, opened 2020-01-10
Changing accounts history logic to remove updates

piotrkosecki:
As requested, remodelled accounts history to not use an updates


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -486,7 +467,11 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
           "I loaded all of {} checkpointed ids from the DB and will proceed to fetch updated accounts information from the chain",
           checkpoints.size
         )
-        AccountsProcessor.process(checkpoints, onlyProcessLatest = true).map(_ => Done)
+        // here we need to get missing bakers for the given block
+        tezosNodeOperator.getBakersByHash(checkpoints.values.map(_._1).toList).flatMap { bakers =>
```
ivanopagano: We can try avoiding to reload data from the node, by using the stored database records.
All the info should be there already, it would imply
* passing a different data type to `AccountsProcessor.process`
* convert the node operator call to a database operator one
* remove the custom baker-rolls fetcher that loads by `block-hash`

In any case, should we not follow this path, I'd suggest adapting the existing node-fetcher to only work on block-hash, instead of writing an -  essentially - duplicate one. I can imagine adapting the client's call to extract the block's hash before fetching the node data for bakers' rolls

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
```diff
@@ -545,6 +545,50 @@ trait BlocksDataFetchers {
     }
   }
 
+  /** a fetcher of baker rolls for blocks */
```
ivanopagano: This might be unnecessary, if we get data from the stored db-records.
Or we might want to convert the existing one to fetch from hashes of blocks, like here.

See also the comment on use site code.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -338,9 +331,11 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
 
       for {
         _ <- db.run(TezosDb.writeBlocksAndCheckpointAccounts(blocks, accountUpdates)) andThen logBlockOutcome
-        delegateCheckpoints <- processAccountsForBlocks(accountUpdates) // should this fail, we still recover data from the checkpoint
+        votingData <- processVotesForBlocks(results.map { case (block, _) => block })
+        delegateCheckpoints <- processAccountsForBlocks(accountUpdates, votingData.map {
```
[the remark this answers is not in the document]
piotrkosecki: done

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -546,6 +531,7 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
      */
     def process(
         ids: Map[AccountId, BlockReference],
+        votingData: List[(BlockHash, List[Voting.BakerRolls])],
```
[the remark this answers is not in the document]
piotrkosecki: done

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: made the changes, removed my fetcher - it was only used during reading from checkpoints, but also had to move up fetching voting data before the checkpoints to ensure that if the writing checpoints will be stopped in the middle we would have corresponding bakers data in the tables

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
piotrkosecki: removed

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -531,7 +532,7 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
      */
     def process(
         ids: Map[AccountId, BlockReference],
-        votingData: List[(BlockHash, List[Voting.BakerRolls])],
+        votingData: Map[BlockHash, List[Voting.BakerRolls]],
```
ivanopagano: The idea to use a map here is to avoid a linear search. Now we should have a simple lookup by key, but I can't see it applied in the following code

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -436,7 +437,7 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
    */
   private[this] def processAccountsForBlocks(
       updates: List[BlockTagged[List[AccountId]]],
-      votingData: List[(BlockHash, List[Voting.BakerRolls])]
+      votingData: Map[BlockHash, List[Voting.BakerRolls]]
```
ivanopagano: The idea to use a map here is to avoid a linear search. Now we should have a simple lookup by key, but I can't see it applied in the following code

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: But am I wrong or now we could end up with stored votes and then a failing save on checkpoints or blocks, leaving "ophaned" records of votes?

Won't those vote be rewritten again on lorre's restart? Please make sure all is kept consistent, let me know if I can help to think of a solution.

piotrkosecki: I think you are right, I'm going to change that, because the previous solution seems more reasonable, although not perfect

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: ok, I removed it from here

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: sure, I forgot to use it and changed it accordingly

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: Yes, I agree, it might've not being apparent, but votes might get lost on failure, at times... 
But making this step more robust might not be that easy

PR #676, opened 2020-01-14
Updating docker build and docker-compose

piotrkosecki:
Making it as a draft because I have issue with building docker image with `testcontainers` library.

So basically what I did is:
- I've updated `Dockerfile` because I wasn't able to build the image - the alphanet metadata file is missing(it was removed at some point). Also updated it with few more additions required to run next steps
- created script for publishing docker images  `docker-publish.sh` which is also capable of publishing them just locally(`./docker-publish.sh local`)
- updated `docker-compose.yml` so now it is possible to run whole stack(Conseil, Lorre, Postgres) with just this file without pulling whole repository(it fetches our image and starts them in correct order with simple `docker-compose up`)

Things left to do:
 ~~- [ ] resolve issue with tests not working during build~~ https://github.com/testcontainers/testcontainers-java/issues/1112
 - [x] update documentation - when the PR is going to be final I'll update it 

I will be grateful for any questions and comments if this is direction you are satisfied with.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ ADD ./docker/wait-for.sh /root/wait-for.sh
 ADD ./sql/conseil.sql /root/sql/conseil.sql
 
 RUN chmod +x /root/entrypoint.sh
-RUN chmod +rwx /root/wait-for.sh
+RUN chmod +rx /root/wait-for.sh
 
 
 ENTRYPOINT ["/root/entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,19 +8,23 @@ USER builduser
 COPY --chown=builduser:builduser . /src
 WORKDIR /src
 RUN rm -rf test-postgres-path
-RUN sbt clean assembly -J-Xss32m
+RUN sbt 'set test in assembly := {}' clean assembly -J-Xss32m
 
 FROM openjdk:13-alpine
 RUN apk --no-cache add ca-certificates
+RUN apk add netcat-openbsd
 WORKDIR /root/
 COPY --from=0 /tmp/conseil.jar conseil.jar
-ADD ./src/main/resources/metadata/tezos.alphanet.conf /root/tezos.alphanet.conf
 ADD ./src/main/resources/metadata/tezos.mainnet.conf /root/tezos.mainnet.conf
 ADD ./src/main/resources/metadata/tezos.babylonnet.conf /root/tezos.babylonnet.conf
 ADD ./src/main/resources/metadata.conf /root/metadata.conf
 ADD ./docker/template.conf /root/template.conf
 ADD ./docker/entrypoint.sh /root/entrypoint.sh
+ADD ./docker/wait-for.sh /root/wait-for.sh
+ADD ./sql/conseil.sql /root/sql/conseil.sql
+
 RUN chmod +x /root/entrypoint.sh
+RUN chmod +rwx /root/wait-for.sh
 
 
 ENTRYPOINT ["/root/entrypoint.sh"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,15 +1,80 @@
 version: "3.1"
 
 services:
+  conseil:
+    image: cryptonomictech/conseil:0.2002.0020-SNAPSHOT
+    ports:
+      - 80:80
+    environment:
+#  List of supported environment variables:
+#  DBHOST=${DB_Host:=db}
+#  DBUSER=${DB_User:=user}
+#  DBPW=${DB_Password:=password}
+#  DBDB=${DB_Database:=conseil}
+#  DBPORT=${DB_Port:=5432}
+#  XTZSCHEME=${XTZ_Scheme:=http}
+#  XTZHOST=${XTZ_Host:=node}
+#  XTZPREFIX=${XTZ_Prefix:=}
+#  XTZPORT=${XTZ_Port:=8732}
+#  XTZNET=${XTZ_Network:=mainnet}
+#  APIPORT=${API_PORT:=80}
+#  APIKEY=${API_KEY:=conseil}
+      DB_Host: conseil-postgres
+      DB_User: conseiluser
+      DB_Password: p@ssw0rd
+      DB_Database: conseil-local
+      DB_Port: 5432
+      XTZ_Scheme: https
+      XTZ_Host: tezos-staging.cryptonomic-infra.tech
+      XTZ_Port: 443
+      XTZ_Network: mainnet
+    entrypoint: ["/root/wait-for.sh", "conseil-postgres:5432", "-t", "120", "--", "/root/entrypoint.sh", "conseil"]
+    volumes:
+      - "conseil-sql:/root/sql/"
+    networks:
+      - conseil-net
+
+  lorre:
+    image: cryptonomictech/conseil:0.2002.0020-SNAPSHOT
+    environment:
+      DB_Host: conseil-postgres
+      DB_User: conseiluser
+      DB_Password: p@ssw0rd
+      DB_Database: conseil-local
+      DB_Port: 5432
+      XTZ_Scheme: https
+      XTZ_Host: tezos-staging.cryptonomic-infra.tech
+      XTZ_Port: 443
+      XTZ_Network: mainnet
+    command: lorre
+    depends_on:
+      - conseil-postgres
+    networks:
+      - conseil-net
+
   conseil-postgres:
     image: postgres:11.6
     ports:
-    - 5432:5432
+      - 5432:5432
     environment:
       POSTGRES_USER: "conseiluser"
       POSTGRES_PASSWORD: "p@ssw0rd"
       POSTGRES_DB: "conseil-local"
       POSTGRES_INITDB_ARGS: "--lc-collate=en_US.UTF-8 -E UTF8"
     volumes:
-    - "./pgdata:/var/lib/postgresql/data"
-    - "./sql/conseil.sql:/docker-entrypoint-initdb.d/conseil.sql"
+#     uncomment if you want your data to be stored in the local directory
+#      - "./pgdata:/var/lib/postgresql/data"
+      - "conseil-sql:/docker-entrypoint-initdb.d/"
+#     use one below if you want sql file modified by yourself
+#      - "./sql/conseil.sql:/docker-entrypoint-initdb.d/conseil.sql"
+    depends_on:
+      - conseil
+    networks:
+      - conseil-net
+
+volumes:
+  conseil-sql:
+
+networks:
+  conseil-net:
+    driver: bridge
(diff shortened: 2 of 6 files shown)


docker-publish.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+if [ "A$1" != "Alocal" ]; then
```
bhaskarkishore: Genuinely curious, why not `"$1" != "local"`?

docker-publish.sh
```diff
@@ -0,0 +1,23 @@
...
+# required to push
+    docker login
```
bhaskarkishore: Would be nice to check if the login was successful before continuing.

docker-compose.yml
```diff
@@ -1,15 +1,80 @@
...
+      DB_Host: conseil-postgres
+      DB_User: conseiluser
+      DB_Password: p@ssw0rd
+      DB_Database: conseil-local
+      DB_Port: 5432
+      XTZ_Scheme: https
+      XTZ_Host: tezos-staging.cryptonomic-infra.tech
+      XTZ_Port: 443
+      XTZ_Network: mainnet
+    entrypoint: ["/root/wait-for.sh", "conseil-postgres:5432", "-t", "120", "--", "/root/entrypoint.sh", "conseil"]
```
bhaskarkishore: Would it be possible to make the timeout setting i.e `-t` configurable somehow?

Dockerfile
```diff
@@ -8,19 +8,23 @@ USER builduser
...
 ADD ./src/main/resources/metadata/tezos.mainnet.conf /root/tezos.mainnet.conf
 ADD ./src/main/resources/metadata/tezos.babylonnet.conf /root/tezos.babylonnet.conf
 ADD ./src/main/resources/metadata.conf /root/metadata.conf
 ADD ./docker/template.conf /root/template.conf
 ADD ./docker/entrypoint.sh /root/entrypoint.sh
+ADD ./docker/wait-for.sh /root/wait-for.sh
+ADD ./sql/conseil.sql /root/sql/conseil.sql
+
 RUN chmod +x /root/entrypoint.sh
+RUN chmod +rwx /root/wait-for.sh
```
bhaskarkishore: Why give write permission to this script?

docker-publish.sh
piotrkosecki: my intension was to run `./docker-publish.sh` to publish to Docker Hub, and `./docker-publish.sh local`, to just build it locally, so when there is no argument in your case it would throw an error(but it would still continue), but this way I'm avoiding additional `if` and it does not throw anything

docker-compose.yml
piotrkosecki: do you mean like argument to the `docker-compose`? in this line there is ` "-t", "120",` which means timeout is 120 seconds

Dockerfile
piotrkosecki: slight overextension from my side, I'll change it 👍

docker-publish.sh
piotrkosecki: ok, I'll add it

docker-compose.yml
bhaskarkishore: Now that i think about it, perhaps that would be overkill. Someone can hand modify this anyways.

docker-publish.sh
bhaskarkishore: Apols but I'm a little lost so i ran a test. I didn't get an error in my case.

```
#!/bin/bash


if [ "$1" != "local" ]; then
	echo "Logging into docker"
fi

echo "Building stuff"
```
`./test.sh` outputs:
```
Logging into docker
Building stuff
```
`./test.sh local` outputs:
```
Building stuff
```

piotrkosecki: oh, my mistake I was testing at first with `if [ $1 != "local" ]; then` and that's why I've got this error:
`./docker-publish.sh: line 3: [: !=: unary operator expected`, thank you for pointing this out!

piotrkosecki:
I don't think there is a possibility to solve `docker build` issue as we cannot run docker container during docker image creation:
https://github.com/testcontainers/testcontainers-java/issues/1112

vishakh:
> I don't think there is a possibility to solve `docker build` issue as we cannot run docker container during docker image creation:
> [testcontainers/testcontainers-java#1112](https://github.com/testcontainers/testcontainers-java/issues/1112)

Does that mean we should disable tests during this process?

piotrkosecki:
> > I don't think there is a possibility to solve `docker build` issue as we cannot run docker container during docker image creation:
> > [testcontainers/testcontainers-java#1112](https://github.com/testcontainers/testcontainers-java/issues/1112)
> 
> Does that mean we should disable tests during this process?

Yes, it is already done in this PR with `'set test in assembly := {}'` in the Dockerfile

docs/README.md
```diff
@@ -53,6 +53,13 @@ env SBT_OPTS="-Dconfig.file=conseil.conf" && sbt -J-Xss32m "runConseil"
 env SBT_OPTS="-Dconfig.file=conseil.conf" && sbt -J-Xss32m "runLorre alphanet"
 ```
 
+### Building from docker-compose
+
+To avoid pulling whole repository you can just download `doccker-compose.yml` from it.
```
bhaskarkishore: Typo in `docker-compose.yml`

docker-compose.yml
```diff
@@ -1,15 +1,80 @@
...
     environment:
       POSTGRES_USER: "conseiluser"
       POSTGRES_PASSWORD: "p@ssw0rd"
       POSTGRES_DB: "conseil-local"
       POSTGRES_INITDB_ARGS: "--lc-collate=en_US.UTF-8 -E UTF8"
     volumes:
-    - "./pgdata:/var/lib/postgresql/data"
-    - "./sql/conseil.sql:/docker-entrypoint-initdb.d/conseil.sql"
+#     uncomment if you want your data to be stored in the local directory
+#      - "./pgdata:/var/lib/postgresql/data"
```
ivanopagano: Changing default, will impact expectations set by the /README.md file, under the paragraph on testing locally using a docker postgres with compose.
I suggest changing that paragraph too, to say that you need to modify the compose file to keep data persistent between restarts.

As a side-note, I think I've lost the reason why we have two different readme files for project-root and `docs` folder? Some information there is overlapping and can easily get inconsistent, can't we find a way to merge them or at least make one refer to the other for details?

docker-compose.yml
```diff
@@ -1,15 +1,80 @@
 version: "3.1"
 
 services:
+  conseil:
+    image: cryptonomictech/conseil:0.2002.0020-SNAPSHOT
```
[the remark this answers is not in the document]
piotrkosecki: yes, if you look at the postgres image it is also hardcoded to `11.6` 😉 we just have it in the different format(based on `BuildInfo.scala` version)

docker-compose.yml
```diff
@@ -1,15 +1,80 @@
...
+#  DBPW=${DB_Password:=password}
+#  DBDB=${DB_Database:=conseil}
+#  DBPORT=${DB_Port:=5432}
+#  XTZSCHEME=${XTZ_Scheme:=http}
+#  XTZHOST=${XTZ_Host:=node}
+#  XTZPREFIX=${XTZ_Prefix:=}
+#  XTZPORT=${XTZ_Port:=8732}
+#  XTZNET=${XTZ_Network:=mainnet}
+#  APIPORT=${API_PORT:=80}
+#  APIKEY=${API_KEY:=conseil}
```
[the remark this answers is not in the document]
piotrkosecki: those are environment variables which could be set by the user for the docker container - I thought it could be good to have them listed all here

docker-compose.yml
```diff
@@ -1,15 +1,80 @@
...
+      XTZ_Port: 443
+      XTZ_Network: mainnet
+    entrypoint: ["/root/wait-for.sh", "conseil-postgres:5432", "-t", "120", "--", "/root/entrypoint.sh", "conseil"]
+    volumes:
+      - "conseil-sql:/root/sql/"
+    networks:
+      - conseil-net
+
+  lorre:
+    image: cryptonomictech/conseil:0.2002.0020-SNAPSHOT
```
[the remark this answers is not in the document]
piotrkosecki: the same answer as above😄 if you want to continue discussion about it please do it in the previous comment - the result should be similar

docker-compose.yml
piotrkosecki: no, this data stays until you remove container with `docker-compose rm`. Now when clearing the data I always run `docker-compose rm` and I delete `/pgdata` directory. With this change you'll only need to `docker-compose rm` after you stop the container.

docker-compose.yml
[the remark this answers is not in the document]
ivanopagano: What I mean is that we're now bound to manually update those values to the latest stable version, at every official release, both here and in the lorre section?

Why not some dynamic value or `latest`?

docker-compose.yml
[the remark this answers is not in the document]
ivanopagano: I don't get from here if the value to be made available in the container are those with or without the underscore, and why would I want to define them via a local env var, with an extra default on top, for some entries?

To be clear, if I happen to uncomment one of those entries, say XTZPORT, what's gonna happen when I execute the compose script?

docker-compose.yml
ivanopagano: I didn't know this, are you saying that if I `docker-compose down`, using the file as in this PR, I would find all the data back when I `docker-compose up` the db?

docker-compose.yml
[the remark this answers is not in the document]
piotrkosecki: about latest tag: https://vsupalov.com/docker-latest-tag/ - it does not work how you probably think it works - been there, done that 😄
`latest` is being called anti-pattern most of the time.
I don't believe we have to change it every release.  What's wrong with user checking docker-hub for the tag of the newest version and replacing it manually? Because I don't think we are going to use it for the development.

docker-compose.yml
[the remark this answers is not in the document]
piotrkosecki: You're right, removed the first part, now when you uncomment the line it should work as expected

docker-compose.yml
piotrkosecki: I don't usually use `-d` parameter, so I use `Ctrl + C` to stop the container.  But you're right `docker-compose down` stops and removes container. Instead try using `docker-compose stop`.

ivanopagano: Ok, got it

docker-compose.yml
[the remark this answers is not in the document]
ivanopagano: Ok, so the idea is that we provide this for the user, not for the developers (as hinted by the current root readme file).
That was the original intention, and I'm ok with changing that as long as we document the purpose correctly to anyone that is approaching the project from github

docs/README.md
piotrkosecki: fixed

docker-compose.yml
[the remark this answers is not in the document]
piotrkosecki: I was also thinking about providing this as a separate docker-compose file and leave current as it is, WDYT?

Issue #680, opened 2020-01-16
The babylon-enabled micheline parser throws away entrypoint info for "parameters"


ivanopagano:
related to #618 

After the change we can correctly parse the post-babylon format for transactions' `parameters`.
During such process though, we're simply discarding information regarding the `entrypoint` of the contract, which is critical to users.

We need to restore such information, evaluating if it can be correctly parsed as part of the whole micheline expression, or if it's to be exposed as part of another wrapping data structure that shows entrypoint and value as separate entries

### Example case
code originally unparsed which showed as
```json
{
"parameters": {
    "entrypoint": "transfer",
    "value": {
      "prim": "Pair",
      "args": [
        {
          "string": "tz1d75oB6T4zUMexzkr5WscGktZ1Nss1JrT7"
        },
       {
          "prim": "Pair",
          "args": [
            {
              "string": "tz1Mj7RzPmMAqDUNFBn5t5VbXmWW4cSUAdtT"
            },
            {
              "int": "11"
            }
          ]
        }
      ]
    }
  }
}
```
is now converted to michelson as
```
{
    "parameters": "Pair \"tz1d75oB6T4zUMexzkr5WscGktZ1Nss1JrT7\" (Pair \"tz1Mj7RzPmMAqDUNFBn5t5VbXmWW4cSUAdtT\" 11)"
}
```

vishakh:
It seems there can be multiple parameter entries per transaction. However, we are blocked on adding array types to our database. Could we instead use a column of `json` type for now? It could look something like this:

```json

[
{
"entrypoint": "transfer",
"value": "// value in Michelson"
},
{
"entrypoint": "withdraw",
"value": "// value in Michelson"
}
]
```

@ivanopagano @dorians  ^^

ivanopagano:
AFAIK Json columns are only supported using the custom-postgres extension for slick, just as with arrays.

Fixing one should fix the other I bet

PR #681, opened 2020-01-16
680 - reverting parsing of entrypoint

dorians:
temporary fix for #680


vishakh:
Since our friends are directly using this branch, we don't need to merge the code into `master`. Instead we can ageee on a strategic approach and create a new PR.

PR #686, opened 2020-01-22
678 - add registered tokens table

### LINKED ISSUE #678: Add a registered tokens table
As the first step towards supporting Tezos tokens through our software stack, we would like to build the skeleton for a registered tokens table in Conseil. The table will contain a list to tokens we know about and officially support. The data in the table will be used to power our wallets and explorers.

For now, we will maintain the list as a set of hard-coded CSV files. In the future, we will pull the data from an on-chain token registry. 

For each Tezos network, e.g. mainnet, babylonnet, etc, we should have a CSV file in our `resources` tree which will contain data in the following format:

`id|name|standard|account_id`

An example record might be:

`1 | USDTez | FA1.2 | tz1eEnQhbwf6trb8Q8mPb2RaPkNk2rN7BKi9`

(Here, for those curious, we refer to the Tezos FA1.2 standard- https://assets.tqtezos.com/)

We could either:
- Load the CSV data into a table on startup and query it like any other table. We could update the table at every block to simulate a 'real' run.
- Point the Conseil API directly to the CSV files.

Either approach is fine for the initial implementation.

piotrkosecki:
resolves #678 

## Required DB changes
New table is being added in this PR
```
CREATE TABLE tezos.registered_tokens (
    id integer PRIMARY KEY,
    name text NOT NULL,
    standard text NOT NULL,
    account_id text NOT NULL
);
```

### DIFF
--- a/src/main/resources/csv/mainnet.csv
+++ b/src/main/resources/csv/mainnet.csv
@@ -1,0 +1,3 @@
+id|name|standard|account_id
+1 | USDTez | FA1.2 | tz1eEnQhbwf6trb8Q8mPb2RaPkNk2rN7BKi9
+
--- a/src/main/resources/metadata/tezos.babylonnet.conf
+++ b/src/main/resources/metadata/tezos.babylonnet.conf
@@ -978,8 +978,8 @@
         }
         account_id {
           visible: true
-          display-name: "Block Level"
-          data-type: "hash"
+          display-name: "Address"
+          data-type: "accountAddress"
         }
       }
     }
--- a/src/main/resources/metadata/tezos.babylonnet.conf
+++ b/src/main/resources/metadata/tezos.babylonnet.conf
@@ -957,5 +957,31 @@
         }
       }
     }
+    registered_tokens {
+      visible: true
+      attributes {
+        id {
+          display-name: "Id"
+          visible: true
+          reference: {
+            entity: "blocks"
+            key: "hash"
+          }
+        }
+        name {
+          display-name: "Name"
+          visible: true
+        }
+        standard {
+          display-name: "Standard"
+          visible: true
+        }
+        account_id {
+          visible: true
+          display-name: "Block Level"
+          data-type: "hash"
+        }
+      }
+    }
   }
 }
--- a/src/main/resources/metadata/tezos.carthagenet.conf
+++ b/src/main/resources/metadata/tezos.carthagenet.conf
@@ -978,8 +978,8 @@
         }
         account_id {
           visible: true
-          display-name: "Block Level"
-          data-type: "hash"
+          display-name: "Address"
+          data-type: "accountAddress"
         }
       }
     }
--- a/src/main/resources/metadata/tezos.carthagenet.conf
+++ b/src/main/resources/metadata/tezos.carthagenet.conf
@@ -957,5 +957,31 @@
         }
       }
     }
+    registered_tokens {
+      visible: true
+      attributes {
+        id {
+          display-name: "Id"
+          visible: true
+          reference: {
+            entity: "blocks"
+            key: "hash"
+          }
+        }
+        name {
+          display-name: "Name"
+          visible: true
+        }
+        standard {
+          display-name: "Standard"
+          visible: true
+        }
+        account_id {
+          visible: true
+          display-name: "Block Level"
+          data-type: "hash"
+        }
+      }
+    }
   }
 }
--- a/src/main/resources/metadata/tezos.mainnet.conf
+++ b/src/main/resources/metadata/tezos.mainnet.conf
@@ -978,8 +978,8 @@
         }
         account_id {
           visible: true
-          display-name: "Block Level"
-          data-type: "hash"
+          display-name: "Address"
+          data-type: "accountAddress"
         }
       }
     }
--- a/src/main/resources/metadata/tezos.mainnet.conf
+++ b/src/main/resources/metadata/tezos.mainnet.conf
@@ -957,5 +957,31 @@
         }
       }
     }
+    registered_tokens {
+      visible: true
+      attributes {
+        id {
+          display-name: "Id"
+          visible: true
+          reference: {
+            entity: "blocks"
+            key: "hash"
+          }
+        }
+        name {
+          display-name: "Name"
+          visible: true
+        }
+        standard {
+          display-name: "Standard"
+          visible: true
+        }
+        account_id {
+          visible: true
+          display-name: "Block Level"
+          data-type: "hash"
+        }
+      }
+    }
   }
 }
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -80,9 +80,11 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
     apiOperations
   )
 
-  readCSV
+  /** Inits registered tokens at startup */
+  initRegisteredTokensFromCsv()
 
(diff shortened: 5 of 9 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -763,6 +763,19 @@ object TezosDatabaseOperations extends LazyLogging {
...
 
+  /** Adds any new level for which a custom event processing has been executed
+    *
+    * @param rows the type of event to record
+    * @return the number of entries saved added to the
+    */
+  def writeRegisteredTokensRowsIfEmpty(
+      rows: List[Tables.RegisteredTokensRow]
+  )(implicit ec: ExecutionContext): DBIO[Option[Int]] =
+    Tables.RegisteredTokens.length.result.flatMap {
```
ivanopagano: There's a `RegisteredTokens.exists` method that will tell if there's any row, and I suspect it to be much more efficient than counting all the rows.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -763,6 +763,19 @@ object TezosDatabaseOperations extends LazyLogging {
   def writeProcessedEventsLevels(eventType: String, levels: List[BigDecimal]): DBIO[Option[Int]] =
     Tables.ProcessedChainEvents ++= levels.map(Tables.ProcessedChainEventsRow(_, eventType))
 
+  /** Adds any new level for which a custom event processing has been executed
```
ivanopagano: Copy-paste leftover?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -763,6 +763,19 @@ object TezosDatabaseOperations extends LazyLogging {
...
+    *
+    * @param rows the type of event to record
+    * @return the number of entries saved added to the
+    */
+  def writeRegisteredTokensRowsIfEmpty(
+      rows: List[Tables.RegisteredTokensRow]
+  )(implicit ec: ExecutionContext): DBIO[Option[Int]] =
```
ivanopagano: Personally I'd express the return value more explicitly to understand what happened, to avoid confusion.

Right now, does an empty list of rows to save look exactly as if the db table wasn't empty?

Also, at call site (in Lorre), I feel the message handling the `None` return value is incorrect, as it's returned only if the database driver will not be able to provide the added row count, even when succeeding.

src/main/resources/csv/mainnet.csv
```diff
@@ -0,0 +1,3 @@
+id|name|standard|account_id
+1 | USDTez | FA1.2 | tz1eEnQhbwf6trb8Q8mPb2RaPkNk2rN7BKi9
+
```
ivanopagano: I swear I'm having a deja-vu but I'll ask the same: why are we choosing csv over adding some more json config?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -79,6 +80,32 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+    import kantan.csv.generic._
+    val rawData: java.net.URL = getClass.getResource("/" + tezosConf.network + ".csv")
+    val reader = rawData.asCsvReader[Tables.RegisteredTokensRow](rfc.withHeader.withCellSeparator('|'))
+
+    def trimFields(cc: Tables.RegisteredTokensRow): Tables.RegisteredTokensRow =
+      cc.copy(name = cc.name.trim, standard = cc.standard.trim, accountId = cc.accountId.trim)
+
+    val (errors, rows) = reader.toList.foldRight((List[ReadError](), List[Tables.RegisteredTokensRow]()))(
+      (acc, pair) => acc.fold(l => (l :: pair._1, pair._2), r => (pair._1, r :: pair._2))
+    )
```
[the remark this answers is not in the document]
piotrkosecki: yes, I googled looking for solution, and apparently there is a way to do it cleanly with Either(pun intended 😄 ) with scalaz or with scala 2.13, but for now this seemed like an efficient solution.

src/main/resources/metadata/tezos.babylonnet.conf
```diff
@@ -957,5 +957,31 @@
...
+        }
+        standard {
+          display-name: "Standard"
+          visible: true
+        }
+        account_id {
+          visible: true
+          display-name: "Block Level"
+          data-type: "hash"
+        }
```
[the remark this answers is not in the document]
piotrkosecki: you're right, changed

src/main/resources/metadata/tezos.carthagenet.conf
```diff
@@ -957,5 +957,31 @@
...
+        standard {
+          display-name: "Standard"
+          visible: true
+        }
+        account_id {
+          visible: true
+          display-name: "Block Level"
+          data-type: "hash"
+        }
+      }
```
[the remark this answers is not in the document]
piotrkosecki: you're right, changed

src/main/resources/metadata/tezos.mainnet.conf
```diff
@@ -957,5 +957,31 @@
...
+          display-name: "Standard"
+          visible: true
+        }
+        account_id {
+          visible: true
+          display-name: "Block Level"
+          data-type: "hash"
+        }
+      }
+    }
```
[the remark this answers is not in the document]
piotrkosecki: you're right, changed

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -79,6 +80,32 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+
+    errors.foreach(err => logger.error(s"Error while reading registered tokens file: ${err.getMessage}"))
+
+    db.run(TezosDb.writeRegisteredTokensRowsIfEmpty(rows.map(trimFields))).map {
```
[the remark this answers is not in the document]

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: ok, used exists, good catch

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: yes, changed

src/main/resources/csv/mainnet.csv
piotrkosecki: not sure, in the ticket there is mentioned specific format of CSV, so I would assume that there is an easy way to extract data in it

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: I made simple changes, but can you elaborate on how do you think it should look like?

src/main/resources/csv/mainnet.csv
vishakh: We can definitely switch over to JSON subsequently if it makes more sense. The choice of CSV in the spec was pretty arbitrary.

ivanopagano: let's put this in a new issue?

PR #687, opened 2020-01-30
Feature/631 record roll count

### LINKED ISSUE #631: Record roll count
- [ ] Store roll count in `baker` table
- [ ] Store roll count in `baker_history` table
- [ ] Remove `rolls` table

dorians:
closes #631 

Stores `rolls` with `delegates` instead of in a separate table.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
@@ -502,8 +502,51 @@ trait BlocksDataFetchers {
   }
 
   /** a fetcher of baker rolls for blocks */
-  implicit val bakersFetcher = new FutureFetcher {
+  implicit val bakersFetcherByHash = new FutureFetcher {
     import JsonDecoders.Circe.Votes._
+    import cats.instances.future._
+
+    type Encoded = String
+    type In = String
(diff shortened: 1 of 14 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
```diff
@@ -502,27 +502,27 @@ trait BlocksDataFetchers {
...
     type Out = List[Voting.BakerRolls]
 
-    private val makeUrl = (block: Block) => s"blocks/${block.data.hash.value}/votes/listings"
+    private val makeUrl = (blockHash: String) => s"blocks/$blockHash/votes/listings"
 
     override val fetchData =
       Kleisli(
         blocks => {
-          logger.info("Fetching bakers in levels {}", blocks.head.data.header.level to blocks.last.data.header.level)
+          logger.info("Fetching {} bakers", blocks.size)
```
vishakh: This is misleading, no? It says bakers but it actually displays the number of blocks.

piotrkosecki: changed

vishakh:
> Looks good for now, but according to the ticket there should be `delegates_history` table created. I think we can do it in the separate PR.

Good catch but I don't think we need the new table right now.

PR #688, opened 2020-01-31
680 - entrypoint support

### LINKED ISSUE #680: The babylon-enabled micheline parser throws away entrypoint info for "parameters"

dorians:
closes #680 

Since we can't write fields as arrays or JSON I wrote `entrypoint` into the separate field.


sql/conseil.sql
```diff
@@ -284,6 +284,7 @@ CREATE TABLE tezos.operations (
     amount numeric,
     destination character varying,
     parameters character varying,
+    parameters_entrypoint character varying,
```
vishakh: Since a transaction can apparently have multiple parameters, we should rename this field to `parameters_entrypoints`.

ivanopagano: I think that to add multiple entrypoints with related scripts, we also need to modify the Model that represents a transaction, for now it doesn't support an array of parameters anyway...

piotrkosecki: @vishakh what do you think about leaving it as `_entrypoint` instead of `_entrypoints` as @ivanopagano suggested?

vishakh: I think `_entrypoints` is more future proof.

piotrkosecki: ok, changed

PR #693, opened 2020-02-18
bugfix - fix for Lorre crashing when querying for inactive bakers.

piotrkosecki:
Fix for Lorre crashing when querying accounts_history table  for inactive bakers.
The query took too long and it occurred in timeout. Testing on my PC it took ~1m45s.
Now querying `accounts` table this time went down to under 1s.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -610,8 +610,8 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
             activatedOperations,
             activatedAccounts.toList
           )
-          accountsWithHistory <- getInactiveBakersWithTaggedAccounts(updatedTaggedAccounts)
-        } yield accountsWithHistory
+          inactiveBakerAccounts <- getInactiveBakersWithTaggedAccounts(updatedTaggedAccounts)
+        } yield inactiveBakerAccounts
 
         accountsWithHistoryFut.flatMap { accountsWithHistory =>
           db.run(TezosDb.writeAccountsAndCheckpointDelegates(accountsWithHistory, taggedDelegateKeys))
@@ -625,22 +625,21 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
 
       def getInactiveBakersWithTaggedAccounts(
           taggedAccounts: List[BlockTagged[Map[AccountId, Account]]]
-      ): Future[List[(BlockTagged[Map[AccountId, Account]], List[Tables.AccountsHistoryRow])]] =
+      ): Future[List[(BlockTagged[Map[AccountId, Account]], List[Tables.AccountsRow])]] =
         Future.traverse(taggedAccounts) { blockTaggedAccounts =>
           if (blockTaggedAccounts.blockLevel % lorreConf.blockRightsFetching.cycleSize == 1) {
             tezosNodeOperator.fetchActiveDelegates(taggedAccounts.map(x => (x.blockLevel, x.blockHash))).flatMap {
               activeBakers =>
                 db.run {
                   TezosDb
-                    .getInactiveBakersFromAccountsHistoryByBlock(
-                      blockTaggedAccounts.blockLevel,
+                    .getInactiveBakersFromAccounts(
                       activeBakers.toMap.apply(blockTaggedAccounts.blockHash)
                     )
                     .map(blockTaggedAccounts -> _)
                 }
             }
           } else {
-            Future.successful(blockTaggedAccounts -> List.empty[Tables.AccountsHistoryRow])
+            Future.successful(blockTaggedAccounts -> List.empty[Tables.AccountsRow])
           }
         }
 
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -630,16 +630,15 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
           if (blockTaggedAccounts.blockLevel % lorreConf.blockRightsFetching.cycleSize == 1) {
             tezosNodeOperator.fetchActiveDelegates(taggedAccounts.map(x => (x.blockLevel, x.blockHash))).flatMap {
               activeBakers =>
+                val activeBakersIds = activeBakers.toMap.apply(blockTaggedAccounts.blockHash)
                 db.run {
                   TezosDb
-                    .getInactiveBakersFromAccounts(
-                      activeBakers.toMap.apply(blockTaggedAccounts.blockHash)
-                    )
+                    .getInactiveBakersFromAccounts(activeBakersIds)
                     .map(blockTaggedAccounts -> _)
                 }
             }
           } else {
-            Future.successful(blockTaggedAccounts -> List.empty[Tables.AccountsRow])
+            Future.successful(blockTaggedAccounts -> List.empty)
           }
         }
 
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
@@ -15,7 +15,6 @@ import tech.cryptonomic.conseil.tezos
 import tech.cryptonomic.conseil.tezos.TezosNodeOperator.FetchRights
 import tech.cryptonomic.conseil.tezos.TezosTypes.{BakingRights, Contract, EndorsingRights}
 import com.typesafe.scalalogging.Logger
-import tech.cryptonomic.conseil.tezos.Tables.AccountsHistoryRow
 
 object DatabaseConversions extends LazyLogging {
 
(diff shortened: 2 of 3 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -153,10 +152,21 @@ object DatabaseConversions extends LazyLogging {
 
         val untouchedAccounts =
           inactiveBakers
-            .map(_.copy(isActiveBaker = Some(false)))
-            .filterNot(accountsHistoryRow => touched.contains(accountsHistoryRow.accountId))
+            .filterNot(inactiveAccountsRow => touched.contains(inactiveAccountsRow.accountId))
```
ivanopagano: style: I guess we could simply use `row` as identifier, without the ceremony, since it's a local and it's very clear from the context what it is.

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -153,10 +152,21 @@ object DatabaseConversions extends LazyLogging {
...
+                )
+                .withFieldConst(_.cycle, blockTaggedAccounts.cycle)
+                .withFieldConst(_.blockLevel, blockTaggedAccounts.blockLevel)
+                .withFieldConst(_.blockId, blockTaggedAccounts.blockHash.value)
+                .withFieldConst(_.isActiveBaker, Some(false))
+                .transform
+            }
 
-        untouchedAccounts ::: touchedAccounts
+        (untouchedAccounts ::: touchedAccounts).distinct
```
ivanopagano: Question: if untouched already filtered out those accounts whose `id` is in `touched`, isn't this last step redundant?
Do you think it'll make things safer?

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -153,10 +152,21 @@ object DatabaseConversions extends LazyLogging {
...
+            .filterNot(inactiveAccountsRow => touched.contains(inactiveAccountsRow.accountId))
+            .map {
+              _.into[Tables.AccountsHistoryRow]
+                .withFieldConst(
+                  _.asof,
+                  Timestamp.from(blockTaggedAccounts.timestamp.getOrElse(Instant.ofEpochMilli(0)))
+                )
+                .withFieldConst(_.cycle, blockTaggedAccounts.cycle)
+                .withFieldConst(_.blockLevel, blockTaggedAccounts.blockLevel)
+                .withFieldConst(_.blockId, blockTaggedAccounts.blockHash.value)
```
ivanopagano: Minor: might it read better if these fields are bound via pattern matching at line 133?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -625,22 +625,21 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
         Future.traverse(taggedAccounts) { blockTaggedAccounts =>
           if (blockTaggedAccounts.blockLevel % lorreConf.blockRightsFetching.cycleSize == 1) {
             tezosNodeOperator.fetchActiveDelegates(taggedAccounts.map(x => (x.blockLevel, x.blockHash))).flatMap {
               activeBakers =>
                 db.run {
                   TezosDb
-                    .getInactiveBakersFromAccountsHistoryByBlock(
-                      blockTaggedAccounts.blockLevel,
+                    .getInactiveBakersFromAccounts(
                       activeBakers.toMap.apply(blockTaggedAccounts.blockHash)
```
ivanopagano: optimize: this should've been catched in earlier reviews but nonetheless...
We want to convert the list to a Map only once for each `flatMap` call... instead we're doing it on every call to the TezosDb method, times the number of flatMap calls.

It might be premature optimization, yet the cost to rewrite is minimal, while each `toMap` will traverse the whole list again to build the new map object, which is also adding to the global GC pressure.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -79,7 +79,7 @@ object TezosDatabaseOperations extends LazyLogging {
     * @return     Database action possibly containing the number of rows written (if available from the underlying driver)
     */
   def writeAccountsHistory(
-      accountsInfo: List[(BlockTagged[Map[AccountId, Account]], List[AccountsHistoryRow])]
+      accountsInfo: List[(BlockTagged[Map[AccountId, Account]], List[Tables.AccountsRow])]
```
ivanopagano: We might want to document this additional piece of input, to keep track of why we need this additional list of existing rows here...

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -479,7 +479,7 @@ object TezosDatabaseOperations extends LazyLogging {
     * @return a database action that stores both arguments and return a tuple of the row counts inserted
     */
   def writeAccountsAndCheckpointDelegates(
-      accounts: List[(BlockTagged[Map[AccountId, Account]], List[AccountsHistoryRow])],
+      accounts: List[(BlockTagged[Map[AccountId, Account]], List[Tables.AccountsRow])],
```
ivanopagano: We might want to only pass in the block-tagged map as a method parameter here, if the existing rows are not needed.
In fact, it's not clear at all, if we look at the method description, what should we pass in for this list of accounts rows... 😉

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -625,22 +625,21 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+      ): Future[List[(BlockTagged[Map[AccountId, Account]], List[Tables.AccountsRow])]] =
         Future.traverse(taggedAccounts) { blockTaggedAccounts =>
           if (blockTaggedAccounts.blockLevel % lorreConf.blockRightsFetching.cycleSize == 1) {
             tezosNodeOperator.fetchActiveDelegates(taggedAccounts.map(x => (x.blockLevel, x.blockHash))).flatMap {
               activeBakers =>
                 db.run {
                   TezosDb
-                    .getInactiveBakersFromAccountsHistoryByBlock(
-                      blockTaggedAccounts.blockLevel,
+                    .getInactiveBakersFromAccounts(
```
ivanopagano: is the existing `getInactiveBakersFrom AccountsHistoryByBlock` used elsewhere still?
Otherwise we might get rid of that.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -563,6 +563,20 @@ object TezosDatabaseOperations extends LazyLogging {
       .as[AccountsHistoryRow]
       .map(_.toList)
 
+  /**
+    * Gets inactive bakers from accounts
+    * @param activeBakers needed to filter out them from all of the bakers
+    * @return inactive baker accounts
+    * */
+  def getInactiveBakersFromAccounts(
+      activeBakers: List[String]
```
ivanopagano: Could we wrap these `strings` as a typed id, or add docs explanation of what should we're expected to pass-in here?

piotrkosecki: I've updated the docs comment with clearer explanation

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: no, I removed it

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: updated docs for that

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: updated the docs

src/main/scala/tech/cryptonomic/conseil/Lorre.scala

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
piotrkosecki: just to make things safer

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
piotrkosecki: ok, changed

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -625,22 +625,21 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
                       activeBakers.toMap.apply(blockTaggedAccounts.blockHash)
                     )
                     .map(blockTaggedAccounts -> _)
                 }
             }
           } else {
-            Future.successful(blockTaggedAccounts -> List.empty[Tables.AccountsHistoryRow])
+            Future.successful(blockTaggedAccounts -> List.empty[Tables.AccountsRow])
```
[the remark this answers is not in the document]
piotrkosecki: before it did not want to infer correctly when it was `.empty[AccountHistoryRow]`, now it works 😄

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
piotrkosecki: I would agree in a smaller method, but I think it is clearer this way in the bigger method

PR #695, opened 2020-02-24
Feature/521 post process ballots and proposal data 2nd approach

### LINKED ISSUE #521: Post-process ballot and proposal data

piotrkosecki:
resolves #521 

## Required DB changes
```
CREATE TABLE tezos.governance (
voting_period integer NOT NULL,
voting_period_kind character varying NOT NULL,
cycle integer,
level integer,
proposal_hash character varying NOT NULL,
yay_count integer,
nay_count integer,
pass_count integer,
yay_rolls numeric,
nay_rolls numeric,
pass_rolls numeric,
total_rolls numeric,
PRIMARY KEY (block_hash, proposal_hash)
);

CREATE INDEX governance_block_hash_idx ON tezos.governance USING btree (block_hash);
CREATE INDEX governance_proposal_hash_idx ON tezos.governance USING btree (proposal_hash);
```

## Required configuration changes
Because I've noticed that for many blocks there was no proposal available when tested against the mainnet I've decided to introduce this parameter so when running governance process for mainnet we could start around block ~360000:
```
batchedFetches.governanceStartLevel: -1
```

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -137,6 +137,10 @@ lorre {
 
     #Used to specify the max-time allowed for each delegate-page to finish processing
     delegatePageProcessingTimeout: 15 minutes
+
+    #In the mainnet there is a huge gap before there are any proposals so we can start at aroud block level ~360000
+    # but by default it should start from the beginning
+    governanceStartLevel: -1
   }
 
   # Custom libSodium settings
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -138,9 +138,6 @@ lorre {
     #Used to specify the max-time allowed for each delegate-page to finish processing
     delegatePageProcessingTimeout: 15 minutes
 
-    #In the mainnet there is a huge gap before there are any proposals so we can start at aroud block level ~360000
-    # but by default it should start from the beginning
-    governanceStartLevel: -1
   }
 
   # Custom libSodium settings
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -7,23 +7,13 @@ import akka.stream.ActorMaterializer
 import mouse.any._
 import com.typesafe.scalalogging.LazyLogging
 import org.slf4j.LoggerFactory
-import tech.cryptonomic.conseil.tezos.{
-  ApiOperations,
-  DatabaseConversions,
-  FeeOperations,
-  ShutdownComplete,
-  Tables,
-  TezosErrors,
-  TezosNodeInterface,
-  TezosNodeOperator,
-  TezosTypes,
-  TezosDatabaseOperations => TezosDb
-}
+import tech.cryptonomic.conseil.tezos.{ApiOperations, DatabaseConversions, FeeOperations, ShutdownComplete, Tables, TezosErrors, TezosNodeInterface, TezosNodeOperator, TezosTypes, TezosDatabaseOperations => TezosDb}
 import tech.cryptonomic.conseil.tezos.TezosTypes._
 import tech.cryptonomic.conseil.io.MainOutputs.LorreOutput
 import tech.cryptonomic.conseil.util.DatabaseUtil
 import tech.cryptonomic.conseil.config._
-import tech.cryptonomic.conseil.tezos.TezosNodeOperator.{FetchRights, LazyPages}
+import tech.cryptonomic.conseil.generic.chain.DataFetcher.fetch
+import tech.cryptonomic.conseil.tezos.TezosNodeOperator.{FetchRights, LazyPages, isGenesis}
 
 import scala.concurrent.duration._
 import scala.annotation.tailrec
@@ -32,6 +22,7 @@ import scala.concurrent.duration.Duration
 import scala.util.{Failure, Success, Try}
 import scala.collection.SortedSet
 import tech.cryptonomic.conseil.tezos.TezosTypes.BlockHash
+import tech.cryptonomic.conseil.tezos.TezosTypes.Voting.Vote
 
 /**
   * Entry point for synchronizing data between the Tezos blockchain and the Conseil database.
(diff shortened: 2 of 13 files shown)


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -83,6 +83,55 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
   /** Inits registered tokens at startup */
   initRegisteredTokensFromCsv()
 
+  /** Inits governance processing */
+  system.scheduler.scheduleOnce(5.seconds)(GovernanceProcessor.processGovernance())
```
vishakh: Since we are storing governance data per-block, can the new table not be populated as part of the regular Lorre iteration process?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -377,6 +386,30 @@ class TezosNodeOperator(
     (fetchProposals, fetchBakers, fetchBallots).tupled.run(blocks.filterNot(b => isGenesis(b.data)))
   }
 
+  /** Fetches detailed data for voting associated to the passed-in blocks */
+  def getVoting(blocks: List[Block]): Future[(List[BakerBlock], List[BallotBlock], List[BallotCountsBlock])] = {
```
vishakh: `getVoting` => `getVotes`

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -672,6 +705,52 @@ class TezosNodeOperator(
     }
   }
 
+  def getBlocksForLevels(
```
vishakh: Please document the new functions.

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: ok, changed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
piotrkosecki: changed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
piotrkosecki: removed, as after change it was unused

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -365,10 +365,33 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+        blockHashesWithProposals = proposals.filter(_._2.isDefined).map(_._1)
+        blocksWithProposals = blocks.filter(blockData => blockHashesWithProposals.contains(blockData.data.hash))
+        (listings, ballots, ballotCounts) <- {
+          if (blocksWithProposals.nonEmpty) {
+            tezosNodeOperator.getVotes(blocksWithProposals)
+          } else {
+            Future.successful((List.empty, List.empty, List.empty))
+          }
+        }
+        governanceRows = (blocksWithProposals.map(_.data), proposals, listings, ballots, ballotCounts)
```
ivanopagano: My opinion on this is that it's more clear to define conversions for each block's bundled data (which might be a custom class made just for this use).
The we map directly with the convertTo function.

It might look more complicated seen from the call site (here), but it simplifies the logic of a conversion function which is too custom tailored to this specific use to be easily refactored/updated later on.

My position on this is that, while we want to simplify how the code is called to make intent clear, we should nonetheless keep the single pieces of logic consistent with an understandable design choice, and easy to compose in a reasonable way.
The idea of a Conversion function is that you have a data type that can be converted to another or more than one. The input data type might be a composition, out of necessity (because we might not have all the required values in  a single object) yet making the input arbitrarily complex for ease of use is going overboard in my opinion.

As a guideline I would strive as much as possible to reduce the input elements to the minimum required to convert to some other data element, eventually creating temporary input class to aggregate multiple pieces of information when they're more than a few.

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -956,4 +958,70 @@ object DatabaseConversions extends LazyLogging {
       }
     }
 
+  implicit val governanceConversion =
+    new Conversion[
```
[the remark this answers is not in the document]
ivanopagano: Refer to the considerations I shared at the point where this function's used

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -350,31 +352,48 @@ class TezosNodeOperator(
       (fetchCurrentQuorum, fetchCurrentProposal).mapN(CurrentVotes.apply)
     }
 
+  /** Fetches proposals for given blocks */
+  def getProposals(blocks: List[BlockData]): Future[List[(BlockHash, Option[ProtocolId])]] = {
```
ivanopagano: Just wondering: is there a way that we could collect all voting details together (maybe in a preparatory method) and then re-use them wherever they're needed.
I mention this as I see we seem to be fetching the same data many times per cycle, both here and in `getVotingDetails` and `getVotes`?

Do that corresponds to the same data?

src/main/resources/reference.conf
```diff
@@ -137,6 +137,10 @@ lorre {
 
     #Used to specify the max-time allowed for each delegate-page to finish processing
     delegatePageProcessingTimeout: 15 minutes
+
+    #In the mainnet there is a huge gap before there are any proposals so we can start at aroud block level ~360000
+    # but by default it should start from the beginning
+    governanceStartLevel: -1
```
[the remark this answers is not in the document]
piotrkosecki: it was removed during update

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -742,6 +749,13 @@ object TezosDatabaseOperations extends LazyLogging {
       case false => Tables.RegisteredTokens ++= rows
     }
 
+  /** Fetches max block level from governance table
+    *
+    * @return  Max level or [[defaultBlockLevel]] if no blocks were found in the database.
```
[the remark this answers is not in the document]
piotrkosecki: no, I didn't, what was your issue?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
piotrkosecki: `getVotingDetails` and `getVotes` are different now - I've noticed that in the only place that method was used most of the values were discarded

ivanopagano: Do you think we can optimize nonetheless?

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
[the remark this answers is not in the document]
piotrkosecki: ok, I've changed it

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: ok, changed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
piotrkosecki: optimized!

ivanopagano: luvely

Issue #697, opened 2020-02-26
Missing metadata


thanlok:
```
[conseil-system-akka.actor.default-dispatcher-5] WARN t.c.c.metadata.UnitTransformation - There're missing metadata overrides for "tezos / carthagenet / operations / errors"
[conseil-system-akka.actor.default-dispatcher-5] WARN t.c.c.metadata.UnitTransformation - There're missing metadata overrides for "tezos / carthagenet / operations / ballot_period"
[conseil-system-akka.actor.default-dispatcher-5] WARN t.c.c.metadata.UnitTransformation - There're missing metadata overrides for "tezos / carthagenet / operations / parameters_entrypoints"
```

I am connected to carthagenet and kept getting these warnings from the conseil part. What is the issue here?

vishakh:
@ivanopagano is working on a fix, thanks.

ivanopagano:
Can we close this?

PR #699, opened 2020-02-26
Add metadata for big maps and related content

### LINKED ISSUE #696: Update metadata to export big maps entities
The metadata configuration needs sections to expose big maps information to conseil's endpoints

ivanopagano:
fixes #696, #697 

Adding metadata definition for big_maps and big_map_contents entities on all available networks

### DIFF
--- a/src/main/resources/metadata/tezos.mainnet.conf
+++ b/src/main/resources/metadata/tezos.mainnet.conf
@@ -745,6 +745,9 @@
         parameters: {
           visible: true
         }
+        parameters_entrypoints: {
+          visible: true
+        }
         manager_pubkey: {
           visible: true
         }
@@ -822,6 +825,10 @@
           visible: true
           display-name: "Vote"
         }
+        ballot_period: {
+          data-type: "int"
+          visible: true
+        }
         number_of_slots: {
           visible: true
           display-name: "Slots"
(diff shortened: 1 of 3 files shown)


ivanopagano:
I left out the table connecting the originated contract with the map, since it only contained the external references.
Would that be needed?

vishakh:
> I left out the table connecting the originated contract with the map, since it only contained the external references.
> Would that be needed?

@ivanopagano We definitely need to expose that table to our users.

src/main/resources/metadata/tezos.mainnet.conf
```diff
@@ -338,9 +338,94 @@
...
+            entity: "accounts"
+            key: "account_id"
+          }
+        }
+      }
+    }
     blocks {
-      display-name-plural: "Blocks"
-      display-name: "Block"
+      display-name-plural: "blocks"
```
piotrkosecki: why did you change it to lowercase only here?

ivanopagano: I guess it must've been an unwilling change that slipped by... thanks

ivanopagano: I think it might've been some semi-automated formatting that I didn't even expect to be there

PR #709, opened 2020-03-06
governance metadata update

anonymoussprocket:


src/main/resources/metadata/tezos.carthagenet.conf
```diff
@@ -1,3 +1,4 @@
+edited
```
ivanopagano: What's this line? Is it an unwanted artifact?

I noticed it on the mainnet conf PR too, now that I checked again

PR #715, opened 2020-03-12
Remove the existing constraint between big maps and content

ivanopagano:
* this should temporarily avoid Lorre from stopping for constraints
   violations
 * further investigation is due to identify how some big maps are not
   stored, possibly because they're from pre-babylon blocks, which
   are not handled in detail

### DIFF
--- a/sql/conseil.sql
+++ b/sql/conseil.sql
@@ -380,9 +380,6 @@ CREATE TABLE tezos.big_map_contents (
     PRIMARY KEY (big_map_id, key)
 );
 
-ALTER TABLE ONLY tezos.big_map_contents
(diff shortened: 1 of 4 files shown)


sql/conseil.sql
```diff
@@ -380,9 +380,6 @@ CREATE TABLE tezos.big_map_contents (
     PRIMARY KEY (big_map_id, key)
 );
 
-ALTER TABLE ONLY tezos.big_map_contents
```
vishakh: How come `Tables.scala` did not change?

ivanopagano: good one! 
weird, indeed... checking out

PR #716, opened 2020-03-12
bugfix - missing account updates from accounts history tables

piotrkosecki:
Added balance updated account ids from the block header metadata.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
@@ -662,6 +662,12 @@ class TezosNodeOperator(
       codeAlter compose storageAlter compose parametersAlter
     }
 
+    def extractAccountIds(blockData: BlockData): List[AccountId] = for {
+      blockHeaderMetadata <- discardGenesis.lift(blockData.metadata).toList
+      balanceUpdate <- blockHeaderMetadata.balance_updates
+      id <- balanceUpdate.contract.map(_.id).toList ::: balanceUpdate.delegate.map(_.value).toList
+    } yield AccountId(id)
+
     //Gets blocks data for the requested offsets and associates the operations and account hashes available involved in said operations
     //Special care is taken for the genesis block (level = 0) that doesn't have operations defined, we use empty data for it
     for {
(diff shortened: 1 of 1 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -682,9 +682,14 @@ class TezosNodeOperator(
       val proposalsMap = proposalsState.toMap
       fetchedBlocksData.map {
         case (offset, md) =>
+          val accountIdsWithUpdatedBalances = for {
```
ivanopagano: It works, might it read better if extracted to a local method?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -682,9 +682,14 @@ class TezosNodeOperator(
...
+            blockHeaderMetadata <- discardGenesis.lift(md.metadata).toList
+            balanceUpdate <- blockHeaderMetadata.balance_updates
+            id <- balanceUpdate.contract.map(_.id).orElse(balanceUpdate.delegate.map(_.value)).toList
```
ivanopagano: Are we sure that `contract` and `delegate` are mutually exclusive in the update?
Maybe you ask me this already but I didn't get exactly your answer possibly.

We might better find a way to compose those "inclusively", like
`update.contract.map(_.id).toList ::: update.delegate.map(_.value).toList`

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
piotrkosecki: ok, extracted

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
piotrkosecki: you are right, I  made this assumption on my own, changed

ivanopagano:
I'm not sure we have test suites at that level of operations, if so, you might wanna add a test case.

PR #718, opened 2020-03-13
bugfix - fixing yay,nay and pass counts in governance table

### LINKED ISSUE #717: Fix yay, nay and pass counts in governance table
The `governance` table columns `yay_count`, `nay_count` and `pass_count` are duplicates of `yay_rolls`, `nay_rolls` and `pass_rolls` respectively. Instead of storing the number of rolls that voted yes, no or pass, the former should store the number of ballot operations for yes, no and pass.

Modified algorithm for these columns:
1. For the current block in Lorre, obtain the current `cycle`. 
1. Fetch all `ballot` operations belong to the current `cycle`.
1. Group by the `vote` column into yay, nay and pass counts.
1. Populate above counts into table.

Alternatively, fetch the counts as of the previous block and simply add the counts for the current block.

piotrkosecki:
resolves #717 

ready for review, but I'll need to test it more thoroughly

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -386,19 +386,13 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
           db.run(TezosDb.getBallotOperationsForCycle(TezosTypes.discardGenesis(block.data.metadata).level.cycle))
             .map(block -> _)
         }
-        ballots <- {
-          if (blocksWithProposals.nonEmpty) {
-            tezosNodeOperator.getVotes(blocksWithProposals)
-          } else {
-            Future.successful(List.empty)
-          }
-        }
+        ballots <- tezosNodeOperator.getVotes(blocksWithProposals)
         governanceRows = groupGovernanceDataByBlock(
-          blocksWithProposals.map(_.data),
-          proposals,
-          listings,
-          ballots,
-          ballotCounts
+          blocksWithProposals,
+          proposals.toMap,
+          listings.toMap,
+          ballots.toMap,
+          ballotCounts.toMap
         ).map(_.convertTo[Tables.GovernanceRow])
         _ <- db.run(TezosDb.insertGovernance(governanceRows))
       } yield ()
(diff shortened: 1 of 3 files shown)


piotrkosecki:
fofund one issue with counting, and fixed it, tested and it is ready for review

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -382,11 +382,15 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
         proposals: List[(BlockHash, Option[ProtocolId])] <- tezosNodeOperator.getProposals(blocks.map(_.data))
         blockHashesWithProposals = proposals.filter(_._2.isDefined).map(_._1)
         blocksWithProposals = blocks.filter(blockData => blockHashesWithProposals.contains(blockData.data.hash))
-        (ballots, ballotCounts) <- {
+        ballotCounts <- Future.traverse(blocksWithProposals) { block =>
+          db.run(TezosDb.getBallotOperationsForCycle(TezosTypes.discardGenesis(block.data.metadata).level.cycle))
+            .map(block -> _)
+        }
+        ballots <- {
```
ivanopagano: I'm surprised that this check is needed, why passing an empty list of "block+proposals" might return something different from a Future empty list?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -406,15 +410,16 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
     ): List[
       (BlockData, Option[ProtocolId], List[Voting.BakerRolls], List[Voting.Ballot], Option[Voting.BallotCounts])
     ] = blockData.map { bd =>
       val hash = bd.hash
+
       val proposal = proposals.find { case (blockHash, _) => blockHash == hash }.flatMap(_._2)
       val listing = listings.find { case (block, _) => block.data.hash == hash }.toList.flatMap(_._2)
       val ballot = ballots.find { case (block, _) => block.data.hash == hash }.toList.flatMap(_._2)
-      val ballotCount = ballotCounts.find { case (block, _) => block.data.hash == hash }.flatMap(_._2)
+      val ballotCount = ballotCounts.find { case (block, _) => block.data.hash == hash }.map(_._2)
```
ivanopagano: Looking at how they're used, I suspect that all those lists should be exposed as Map[BlockHash, Value]?

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: I believe I put there before few changes as this fixed things, now I tested it and it works without this check

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: changed to maps, not exactly [BlockHash, Value], but still it simplified a bit

PR #725, opened 2020-03-17
720 - include transaction parameters in micheline format in operations table

### LINKED ISSUE #720: Include transaction parameters in Micheline format in operations table
Some of our users prefer to look at transaction parameters in Micheline format. We should therefore add a new column called `parameters_micheline` in the `operations` table which simply stores the unparsed Micheline from the Tezos node.

piotrkosecki:
resolves #720

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
@@ -662,11 +662,12 @@ class TezosNodeOperator(
       codeAlter compose storageAlter compose parametersAlter
     }
 
-    def extractAccountIds(blockData: BlockData): List[AccountId] = for {
-      blockHeaderMetadata <- discardGenesis.lift(blockData.metadata).toList
-      balanceUpdate <- blockHeaderMetadata.balance_updates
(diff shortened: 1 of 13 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -653,20 +653,22 @@ class TezosNodeOperator(
     val proposalsStateFetch =
       fetchMerge(currentQuorumFetcher, currentProposalFetcher)(CurrentVotes.apply)
 
-    def parseMichelsonScripts: Block => Block = {
+    def parseMichelsonScripts: Block => Block = { block: Block =>
```
ivanopagano: Why re-introduce the parameter to create an explicit lambda on top of an existing function?

piotrkosecki: because I needed explicit block to extract micheline from the lens before parsing, maybe I misses simpler solution, do you have any suggestions?

ivanopagano: no, I simply missed this detail, thanks!

PR #726, opened 2020-03-18
719 - added dateTime related columns to the operations and blocks tables

### LINKED ISSUE #719: Add day, month, year, time and date to blocks and operations tables
Arronax is supposed to use Conseil data to help users find and chart interesting trends. However, since the only time-oriented column is timestamp, it is not possible to aggregate by month, day or year easily in Arronax and therefore produce useful reports. 

We should add the following columns to `blocks` and `operations` based on timestamp:
- year (yyyy)
- month (mm)
- day (dd)
- time (in 24-hour format)

This can be added to other tables in the future.

piotrkosecki:
resolves #719

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
@@ -7,7 +7,8 @@ import tech.cryptonomic.conseil.util.Conversion
 import cats.{Id, Show}
 import cats.implicits._
 import java.sql.Timestamp
-import java.time.Instant
+import java.time.{Instant, ZoneOffset}
+import java.time.format.DateTimeFormatter
 
 import monocle.Getter
 import io.scalaland.chimney.dsl._
@@ -41,6 +42,17 @@ object DatabaseConversions extends LazyLogging {
   def extractBigDecimal(number: BigNumber): Option[BigDecimal] = number match {
     case Decimal(value) => Some(value)
     case _ => None
+  }
+
+  def extractDateTime(timestamp: java.sql.Timestamp): (Int, Int, Int, String) = {
+    val offsetDateTime = timestamp.toLocalDateTime.atOffset(ZoneOffset.UTC)
+    val format = DateTimeFormatter.ofPattern("HH:mm:ss")
+    (
+      offsetDateTime.getYear,
+      offsetDateTime.getMonth.getValue,
+      offsetDateTime.getDayOfMonth,
+      offsetDateTime.format(format)
+    )
   }
 
   private def extractResultErrorIds(errors: Option[List[OperationResult.Error]]) = {
@@ -176,6 +188,7 @@ object DatabaseConversions extends LazyLogging {
       val header = from.data.header
       val metadata = discardGenesis.lift(from.data.metadata)
       val CurrentVotes(expectedQuorum, proposal) = from.votes
+      val (year, month, day, time) = extractDateTime(toSql(header.timestamp))
       Tables.BlocksRow(
         level = header.level,
         proto = header.proto,
@@ -202,7 +215,11 @@ object DatabaseConversions extends LazyLogging {
         metaVotingPeriod = metadata.map(_.level.voting_period),
         metaVotingPeriodPosition = metadata.map(_.level.voting_period_position),
         expectedCommitment = metadata.map(_.level.expected_commitment),
-        priority = header.priority
+        priority = header.priority,
+        utcYear = year,
+        utcMonth = month,
+        utcDay = day,
+        utcTime = time
       )
     }
   }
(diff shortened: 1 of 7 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
```diff
@@ -591,7 +591,7 @@ trait BlocksDataFetchers {
     override val fetchData =
       Kleisli(
         blocks => {
-          logger.info("Fetching ballots in levels {}", blocks.head.data.header.level to blocks.last.data.header.level)
+          logger.info("Fetching ballots")
```
vishakh: It was nice to have this info.

piotrkosecki: ah, yes, that's an issue which needed to be fixed, and I fixed just for testing - it happens that we have sometimes  empty list of blocks passed here, so `blocks.head` crashed the Lorre. 
Filtering of blocks was done to avoid making queries when there are no proposals - they would be empty.

So should I make another PR with this fix, or just keep it there?

vishakh: Whichever is quicker. :)

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -43,6 +44,17 @@ object DatabaseConversions extends LazyLogging {
     case _ => None
   }
 
+  def extractDateTime(timestamp: java.sql.Timestamp): (Int, Int, Int, String) = {
```
ivanopagano: Not sure why, but I feel like this will blow up in our face later on?

How do we know that the Timestamp was originally from UTC zone?

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -176,6 +188,7 @@ object DatabaseConversions extends LazyLogging {
       val header = from.data.header
       val metadata = discardGenesis.lift(from.data.metadata)
       val CurrentVotes(expectedQuorum, proposal) = from.votes
+      val (year, month, day, time) = extractDateTime(toSql(header.timestamp))
```
ivanopagano: I guess we don't have any guarantee on how the json timestamp will be formatted nah? 🤔 
Otherwise we could simply parse the string...

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
piotrkosecki: because Timestamp is universal and it does not have a zone

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
piotrkosecki: timestamp in tezos blocks has Z in the end, which stands for Zulu time zone, which is equal to UTC, which means its easily converted to timestamp in millis - had to do some investigation beforehand 😉

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeFetchers.scala
piotrkosecki: hopefully we can merge it today :)

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
ivanopagano: That's my point. I wouldn't be able to guarantee that the time part was actually computed with UTC as reference, that bit of information was lost.

I think that noting this on a doc comment could be fine to start with. It might mention that this function needs to be called with that convention being observed?

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
ivanopagano: Do you mean here that the timestamp always refers to Zulu, or that we could use a parser here, instead of converting to sql first?

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
piotrkosecki: ok, added the comment

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
piotrkosecki: I don't think we have a original string  here anymore after parsing it as `ZonedDateTime`, so  converting it to timestamp gives me simple UTC time

PR #727, opened 2020-03-18
Metadata fix for Arronax #304

anonymoussprocket:


src/main/resources/metadata/tezos.carthagenet.conf
```diff
@@ -574,6 +574,10 @@
         hash {
           visible: true
           data-type: "hash"
+          reference: {
+            entity: "operations"
+            key: "operation_group_hash"
+          }
```
ivanopagano: Shouldn't be the other way round? 
The reference is from operations, referring to the containing group, via the `operation_group_hash` field.

src/main/resources/metadata/tezos.mainnet.conf
```diff
@@ -574,6 +574,10 @@
         hash {
           visible: true
           data-type: "hash"
+          reference: {
+            entity: "operations"
+            key: "operation_group_hash"
+          }
```
ivanopagano: Same as for the other file, clearly.

src/main/resources/metadata/tezos.carthagenet.conf
inventorofwheels: operations is the table with the details.

anonymoussprocket: operations is the table with the details.

PR #722, opened 2020-03-16
679 - store token balances

ivanopagano:
Closes #679 

We extract each transaction referring to known token smart contracts and use those to keep a record of balances for each account for any token.

The code to read the actual value of the balance and to which account it's referring is custom tailored to each network's smart ledger.

In the current PR we apply a known FA1.2 contract found on babylon-net.

### Needed updates to Database schema
We introduce the token balance table as

```sql
CREATE TABLE tezos.token_balances (
    token_id integer,
    address text NOT NULL,
    balance numeric NOT NULL,
    block_id character varying NOT NULL,
    block_level numeric DEFAULT '-1'::integer NOT NULL,
    asof timestamp without time zone NOT NULL,
    PRIMARY KEY (token_id, address, block_level)
);
```

### additional libraries
I've used the scorex foundation libraries to compute crypto-related values, in particular for string codecs and hashing.

The library should introduce minimal transitive dependencies and avoid libsodium issue with locally installed native libraries.

```
[info]   +-org.scorexfoundation:scorex-util_2.12:0.1.6 [S]
[info]   | +-com.typesafe.scala-logging:scala-logging_2.12:3.9.2 [S]
[info]   | | +-org.scala-lang:scala-reflect:2.12.8 [S]
[info]   | | +-org.slf4j:slf4j-api:1.7.25
[info]   | |
[info]   | +-org.rudogma:supertagged_2.12:1.4 [S]
[info]   |
[info]   +-org.scorexfoundation:scrypto_2.12:2.1.7 [S]
[info]   | +-com.google.guava:guava:20.0
[info]   | +-com.typesafe.scala-logging:scala-logging_2.12:3.9.2 [S]
[info]   | | +-org.scala-lang:scala-reflect:2.12.8 [S]
[info]   | | +-org.slf4j:slf4j-api:1.7.25
[info]   | |
[info]   | +-org.bouncycastle:bcprov-jdk15on:1.64
[info]   | +-org.rudogma:supertagged_2.12:1.4 [S]
[info]   | +-org.scorexfoundation:scorex-util_2.12:0.1.6 [S]
[info]   | | +-com.typesafe.scala-logging:scala-logging_2.12:3.9.2 [S]
[info]   | | | +-org.scala-lang:scala-reflect:2.12.8 [S]
[info]   | | | +-org.slf4j:slf4j-api:1.7.25
[info]   | | |
[info]   | | +-org.rudogma:supertagged_2.12:1.4 [S]
[info]   | |
[...]

PR #733, opened 2020-03-23
730 - added known_addresses table

### LINKED ISSUE #730: Add a known_addresses table
At the moment, Conseil simply stores Tezos addresses without identifying at all who they belong to. Using manually collected data, we would like to provide hints to users about the owners of these accounts. 

A new `known_addresses` table should be created and its contents read from a CSV file along the lines of the `registered_tokens` table. 

@Taaseen-Ali can provide the CSV file.

piotrkosecki:
resolves #730 

In this PR refactored CSV handling to make it more generic and made some changes compatible with @ivanopagano PR #722 

I believe we still need to consult on how to move forward with CSV handling.


src/main/resources/metadata/tezos.babylonnet.conf
```diff
@@ -1138,5 +1138,18 @@
         }
       }
     }
+    known_addresses {
+      visible: true
+      attributes {
+        address {
+          display-name: "Address"
+          visible: true
+        }
```
ivanopagano: I think adding 
```json
data-type: "accountAddress"
```
can improve UI-wise

src/main/resources/metadata/tezos.carthagenet.conf
```diff
@@ -1091,5 +1091,18 @@
         }
       }
     }
+    known_addresses {
+      visible: true
+      attributes {
+        address {
+          display-name: "Address"
+          visible: true
+        }
```
ivanopagano: I think adding 
```json
data-type: "accountAddress"
```
can improve UI-wise

src/main/resources/metadata/tezos.mainnet.conf
```diff
@@ -1137,5 +1137,18 @@
         }
       }
     }
+    known_addresses {
+      visible: true
+      attributes {
+        address {
+          display-name: "Address"
+          visible: true
+        }
```
ivanopagano: I think adding 
```json
data-type: "accountAddress"
```
can improve UI-wise

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -755,19 +757,68 @@ object TezosDatabaseOperations extends LazyLogging {
...
-  /** Inserts registered tokens to the table if empty
-    *
+  /** Inserts to the table if table is empty
+    * @param table slick TableQuery[_] to which we want to insert
     * @param rows the type of event to record
     * @return the number of entries saved added to the
     */
-  def writeRegisteredTokensRowsIfEmpty(
-      rows: List[Tables.RegisteredTokensRow]
+  def insertIfEmptyTable[A <: AbstractTable[_]](
```
ivanopagano: make-up: possibly renaming to `insertWhenEmpty`? (the table is in the params in any case, it should be clear we talk about that)

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -755,19 +757,68 @@ object TezosDatabaseOperations extends LazyLogging {
   def writeProcessedEventsLevels(eventType: String, levels: List[BigDecimal]): DBIO[Option[Int]] =
     Tables.ProcessedChainEvents ++= levels.map(Tables.ProcessedChainEventsRow(_, eventType))
 
-  /** Inserts registered tokens to the table if empty
-    *
+  /** Inserts to the table if table is empty
+    * @param table slick TableQuery[_] to which we want to insert
     * @param rows the type of event to record
```
ivanopagano: "event" here might be a copy/paste error

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -755,19 +757,68 @@ object TezosDatabaseOperations extends LazyLogging {
...
   )(implicit ec: ExecutionContext): DBIO[Option[Int]] =
-    Tables.RegisteredTokens.exists.result.flatMap {
+    table.exists.result.flatMap {
       case true => DBIO.successful(Some(0))
-      case false => Tables.RegisteredTokens ++= rows
+      case false => table ++= rows
     }
 
+  import shapeless._
+  import shapeless.ops.hlist._
```
ivanopagano: 😱  😄

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -755,19 +757,68 @@ object TezosDatabaseOperations extends LazyLogging {
...
+
+  /** Trims if passed a String value, otherwise returns the value unchanged */
+  object Trimmer extends Poly1 {
+    implicit val stringTrim = at[String] { _.trim }
+    implicit def noop[T] = at[T] { identity }
+  }
+
+  /** Uses a Generic to transform the instance into an HList, maps over it and convert it back into the case class */
```
ivanopagano: I know it wasn't like this before, but I think we might want to move in the direction of

 * specifically define what the function does: "maps over ..." seems quite general
 * add doc line for types used in each param (even implicit)
 * add a short wiki-entry to descrive how Generics work and how we apply them here, for people with no experience in scala and shapeless

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -755,19 +757,68 @@ object TezosDatabaseOperations extends LazyLogging {
...
+      m: Mapper.Aux[Trimmer.type, H, H],
+      ec: ExecutionContext
+  ): Future[List[A#TableElementType]] = {
+
+    val rawData: java.net.URL = getClass.getResource(s"/${table.baseTableRow.tableName}/" + network + ".csv")
+    val reader: CsvReader[ReadResult[A#TableElementType]] =
+      rawData.asCsvReader[A#TableElementType](rfc.withHeader.withCellSeparator(separator))
+
+    // separates List[Either[L, R]] into List[L] and List[R]
+    val (errors, rows) = reader.toList.foldRight((List[ReadError](), List[A#TableElementType]()))(
```
ivanopagano: I'd rather use `List.empty[...]` every time it's possible instead of the `List()` builder

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -755,19 +757,68 @@ object TezosDatabaseOperations extends LazyLogging {
...
+  import shapeless.ops.hlist._
+
+  /** Trims if passed a String value, otherwise returns the value unchanged */
+  object Trimmer extends Poly1 {
+    implicit val stringTrim = at[String] { _.trim }
+    implicit def noop[T] = at[T] { identity }
+  }
+
+  /** Uses a Generic to transform the instance into an HList, maps over it and convert it back into the case class */
+  private def trimCaseClass[C, H <: HList](
```
ivanopagano: Suggestion: `trimStringFields[...](...)`?

For me it conveys more the goal of the function then saying it applies to case classes 🙂

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -755,19 +757,68 @@ object TezosDatabaseOperations extends LazyLogging {
...
+
+  /** Reads and inserts CSV file to the database for the */
+  def initTableFromCsv[A <: AbstractTable[_], H <: HList](table: TableQuery[A], network: String, separator: Char = ',')(
+      implicit hd: HeaderDecoder[A#TableElementType],
+      g: Generic.Aux[A#TableElementType, H],
+      m: Mapper.Aux[Trimmer.type, H, H],
+      ec: ExecutionContext
+  ): Future[List[A#TableElementType]] = {
+
+    val rawData: java.net.URL = getClass.getResource(s"/${table.baseTableRow.tableName}/" + network + ".csv")
```
ivanopagano: Shouldn't this be passed in as a function parameter?

piotrkosecki: I thought I'll use convention which Vishakh suggested in your ticket - he asked you to rename directory to name of the table(not exactly with this words, but this is what happened), so if we keep it this way we avoid duplicating information passed to this method

ivanopagano: Sorry for insisting, it's just that I already refactored that part in my PR, so we should decide how to proceed. Either mine or your PR will eventually need changes 😄

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -805,13 +805,13 @@ object TezosDatabaseOperations extends LazyLogging {
       rawData.asCsvReader[A#TableElementType](rfc.withHeader.withCellSeparator(separator))
 
     // separates List[Either[L, R]] into List[L] and List[R]
-    val (errors, rows) = reader.toList.foldRight((List[ReadError](), List[A#TableElementType]()))(
-      (acc, pair) => acc.fold(l => (l :: pair._1, pair._2), rr => (pair._1, trimCaseClass(rr) :: pair._2))
+    val (errors, rows) = reader.toList.foldRight((List.empty[ReadError], List[A#TableElementType]()))(
```
ivanopagano: there's another List.empty, in the tuple 😛

src/main/resources/metadata/tezos.babylonnet.conf
piotrkosecki: ok, changed

src/main/resources/metadata/tezos.carthagenet.conf
piotrkosecki: ok, added

src/main/resources/metadata/tezos.mainnet.conf
piotrkosecki: ok, added

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: sure, changed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: yes, changed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: sure, changed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: changed

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: missed it 😄

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
piotrkosecki: don't get overexcited, I just used the example from stackoverflow 😛 
https://stackoverflow.com/questions/45885976/scala-is-there-a-generic-way-to-trim-all-string-fields-of-case-classes

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: Ok, I'd say leave it like that and we'll see what happens when I need to merge the other PR

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: What about this @piotrkosecki 
Do you plan anything?

piotrkosecki: That's the comment from stackoverflow. I used the code from the so I left the original comment. I'm not an expert in shapeless 😛

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -80,16 +82,28 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
+      implicit hd: HeaderDecoder[A#TableElementType],
+      g: Generic.Aux[A#TableElementType, H],
+      m: Mapper.Aux[tech.cryptonomic.conseil.util.ConfigUtil.Csv.Trimmer.type, H, H]
+  ): Future[Option[ProposalSupporters]] = {
+    val rows = ConfigUtil.Csv.readTableRowsFromCsv(table, network, separator)
+    db.run(insertWhenEmpty(table, rows)) andThen {
+      case Success(_) => logger.info(s"Written ${rows.size} ${table.baseTableRow.tableName} rows")
+      case Failure(e) => logger.error(s"Could not fill ${table.baseTableRow.tableName} table", e)
     }
+  }
```
ivanopagano: I'd be surprised if this method signature would require to be that complex anymore.

It should do nothing more than what is on the code before the change, that is, read the tokens and call the db.
All those implicits shouldn't be required anymore, and neither should the extra shapeless and kantan imports, apart from the one already there.

ivanopagano: I also think that there is some incorrect type inference that proposed the result type as
`Future[Option[ProposalSupporters]]`, trying to be too smart 😁

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -7,6 +7,7 @@ import akka.stream.ActorMaterializer
 import mouse.any._
 import com.typesafe.scalalogging.LazyLogging
 import org.slf4j.LoggerFactory
+import slick.lifted.{AbstractTable, TableQuery}
```
ivanopagano: these imports might not be needed if the code for initTableFromCsv gets simplified

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: Unfortunately it requires those implicits. They are generated for the specific types, so they need to be passed with the method. If I had implementation for the specific type like just for `Tables.KnownAddresses` I would not require to put it there.

piotrkosecki: and regarding the `ProposalSupporters`, yes, that was IntelliJ helping 😄

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
piotrkosecki: no, they are needed

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
ivanopagano: ooh, my bad, I didn't realize it's generic of his own... what about moving this into TezosDatabaseOperations, to keep the Lorre code more clean?

piotrkosecki: I moved it out of there as TezosDatabaseOperations did not use `db.run()` and I did not want to introduce it there.

PR #735, opened 2020-03-24
731 - add registered bakers table

### LINKED ISSUE #731: Add a registered_bakers table
Conseil should make available details of various Tezos baking services. This data should be sourced from the Tezos Baker Registry stored on the blockchain. As parsing the corresponding smart contract is somewhat complicated, we will first manually pull this data as a CSV file and hard-code it into Conseil. In a future ticket, we will make the data load dynamic.

The current task is to therefore create a `registered_bakers` table which is read from a CSV file just like `registered_tokens`. @Taaseen-Ali can provide the CSV file.

piotrkosecki:
resolves #731 

Needs to be merged after PR https://github.com/Cryptonomic/Conseil/pull/733


src/main/resources/metadata/tezos.babylonnet.conf
```diff
@@ -1138,5 +1138,136 @@
...
+          data-type: "accountAddress"
+          visible: true
+        }
+        alias {
+          display-name: "Alias"
+          visible: true
+        }
+      }
+    }
+    known_addresses {
```
ivanopagano: i assume `bakers_registry` here 😉

src/main/resources/metadata/tezos.carthagenet.conf
```diff
@@ -1088,5 +1088,136 @@
...
+          data-type: "accountAddress"
+          visible: true
+        }
+        alias {
+          display-name: "Alias"
+          visible: true
+        }
+      }
+    }
+    known_addresses {
```
ivanopagano: i assume bakers_registry here, too 😉

src/main/resources/metadata/tezos.mainnet.conf
```diff
@@ -1258,5 +1258,136 @@
...
+          data-type: "accountAddress"
+          visible: true
+        }
+        alias {
+          display-name: "Alias"
+          visible: true
+        }
+      }
+    }
+    known_addresses {
```
ivanopagano: and here too

src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -80,16 +87,40 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
-  val tokenRows: List[Tables.RegisteredTokensRow] =
-    ConfigUtil.Csv.readTableRowsFromCsv(Tables.RegisteredTokens, tezosConf.network, separator = '|')
+  implicit val timestampDecoder: CellDecoder[java.sql.Timestamp] = (e: String) => {
+    val format = DateTimeFormatter.ofPattern("EEE MMM dd yyyy HH:mm:ss 'GMT'Z (zzzz)")
+    StringDecoder
+      .makeSafe("Instant")(s => Instant.from(format.parse(s)))(e)
+      .map(Timestamp.from)
+      .left
+      .map(x => DecodeError.TypeError(x.message))
+  }
```
ivanopagano: This will be impacted by possible changes to #733, I assume.

I suggest putting this inside the initTable.. method rather than here.

piotrkosecki: I cannot. Because I use type parameters in the method it ti needs to have implicits for the defined types beforehand. There is now way of generating decoders for a type `A`, but before method is being called the type is known and this way it can work.

src/main/resources/metadata/tezos.babylonnet.conf
piotrkosecki: thanks! great catch!

src/main/resources/metadata/tezos.carthagenet.conf
piotrkosecki: thanks! great catch!

src/main/resources/metadata/tezos.mainnet.conf
piotrkosecki: thanks! great catch!

PR #736, opened 2020-03-24
Extract the csv to table row parsing to a dedicated object in config

ivanopagano:
This is preliminary common code refactoring for 
#722 #733 #735, all of which reads csv data to database table rows.

Integrating this PR first will provide a common ground to merge the other downstream PRs.
This should simplify the process and avoid additional code conflicts.

Here we extract the code to parse a resource csv file and generate corresponding table rows in a general fashion, supported by generic programming and the csv library.
The code would be now defined in a custom sub-module of the ConfigUtil object, where it might belong better.

Currently such code is used to load registered tokens from the config.

### note
I plan to add a wiki entry for the project as an overview intro to shapeless generic programming, right after that, as a quick reference.


ivanopagano:
@vishakh, @piotrkosecki please consider this PR before the other pending.

We can separately define how to best handle csv config here, without polluting PRs related to different concerns with extra refactors.

PR #740, opened 2020-03-26
Metadata update

anonymoussprocket:


src/main/resources/metadata/tezos.carthagenet.conf
```diff
@@ -303,8 +303,8 @@
       }
     }
     big_maps {
-      display-name-plural: "Big Maps"
-      display-name: "Big Map"
+      display-name-plural: "Map Definitions"
```
vishakh: I feel this name is too abstract. There can be different kinds of maps.

anonymoussprocket: In the current context, I think this is a better way to describe them. We can also add a "description" field if you'd like.

vishakh: A description would be good. Also, this branch now has a merge conflict.

anonymoussprocket: the conflict is due to a file that should die

vishakh: Backwards compatibility should not be so trivially broken. A simple pull from `master` should sort it out so we can merge this PR.

vishakh:
Assuming this is abandoned.

PR #745, opened 2020-03-30
737 - add yay nay pass per block in governance table

### LINKED ISSUE #737: Add per-block yay, nay and pass counts in governance table
This is a follow up to #717. 

The governance table contains cumulative yay, nay and pass counts for the current cycle. We would also like to have per-block counts. Let's therefore add the following columns to the table:
- block_yay_count
- block_nay_count
- block_pass_count
- block_yay_rolls, 
- block_nay_rolls
- block_pass_rolls

The algorithm can be any modification of the existing algorithm as long as it performs well.

piotrkosecki:
resolves #737

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
@@ -553,25 +553,6 @@ object TezosDatabaseOperations extends LazyLogging {
       }
 
   /** Gets ballot operations for given level */
-  def getBallotOperationsCountForLevel(level: Int)(implicit ec: ExecutionContext): DBIO[Voting.BallotCounts] =
-    Tables.Operations
-      .filter(op => op.kind === "ballot" && op.level === level)
-      .groupBy(_.ballot)
-      .map {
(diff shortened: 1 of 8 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -552,6 +552,44 @@ object TezosDatabaseOperations extends LazyLogging {
         Voting.BallotCounts(yaysCount, naysCount, passesCount)
       }
 
+  /** Gets ballot operations for given level */
+  def getBallotOperationsCountForLevel(level: Int)(implicit ec: ExecutionContext): DBIO[Voting.BallotCounts] =
```
ivanopagano: I fail to see the difference with `getBallotOperationsForLevel` 😅

piotrkosecki: check out the filter part 😉

piotrkosecki: yep, double copied, removed this leftover, good catch!

PR #754, opened 2020-04-02
747 - operation tables fields mismatch

### LINKED ISSUE #747: Operation tables field mismatch
In some cases parameters and micheline parameters content is does not match.

piotrkosecki:
Fixed parameters in transaction not being transformed correctly. 
resolves #747

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
@@ -649,7 +649,7 @@ class TezosNodeOperator(
       implicit lazy val _ = logger
       def rewriteParametersInTransactions(transaction: Transaction): Transaction = {
         val expr = transaction.parameters.map {
-          case Left(value) => value.value.expression
+          case Left(Parameters(value, _)) => value.expression
           case Right(value) => value.expression
         }
         transaction.copy(parameters_micheline = expr)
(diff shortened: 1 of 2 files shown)


vishakh:
Could you describe the precise issue and the fix?

piotrkosecki:
@vishakh The issue was that before my changes lenses were modifying every `parameters_micheline` with the first `parameters` that were found in the block. 
After my change, code will be modifying transaction operation in with `parameters_micheline` from the same transaction.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
```diff
@@ -645,14 +645,21 @@ class TezosNodeOperator(
     val proposalsStateFetch =
       fetchMerge(currentQuorumFetcher, currentProposalFetcher)(CurrentVotes.apply)
 
-    def parseMichelsonScripts: Block => Block = { block: Block =>
+    def parseMichelsonScripts: Block => Block = {
       implicit lazy val _ = logger
+      def rewriteParametersInTransactions(transaction: Transaction): Transaction = {
+        val expr = transaction.parameters.map {
+          case Left(value) => value.value.expression
```
ivanopagano, line 652: I'd suggest pattern matching on `value` too, to help when reading the code

ivanopagano:
Another thing I plan to do, related to this issue, is to put a wiki entry to show examples of how the most common optics work, which will come handy when working with this parts of the code.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosNodeOperator.scala
piotrkosecki: ok, patter matched and renamed for clarity

PR #755, opened 2020-04-02
Fix platform tests for added indexing of operations columns

ivanopagano:
This is hotfix PR.

One of the latest database index additions broke the tests for the platform attributes shown by conseil.


ivanopagano:
> Fixed this test in my other PR 👍

I did the same, to make the tests pass on another PR, yet I'm beginning to separate this kind of work if it's not directly relevant to the task at hand.

My approach now is to try fix this small parts in clean commits and then cherry-pick those to a separate PR which will get merged back in the WIP later on.
I'm suggesting this as a general way to do things in the future so we can get a saner review process.
Especially I'm the one that used to put side-track changes inside other PRs so I'm trying to be more strict in the future.

PR #756, opened 2020-04-02
Refactor to extract big-map logic from converters

ivanopagano:
This change concentrate the logic of big map handling more into the BigMapOperations, and simplifies a lot of the conversion code to database objects, which handled such logic previously.

The goal is to make the steps more obvious, use more meaningful names, separate concerns better.

The changes are extracted from ongoing work to extract Tezos Naming Service information from smart contracts transactions, and are needed to make such work approachable, given that too much was delegated to the Conversion classes, a choice that wouldn't fit in this case.


piotrkosecki:
I assume the lenses, which are not in use, are for your other ongoing work with Tezos Naming Service, right?

ivanopagano:
> I assume the lenses, which are not in use, are for your other ongoing work with Tezos Naming Service, right?

The goal was to move all lenses progressively to the same source file, and removing any duplication.
As I mentioned, I'd also rename the objects to something more general like ModelNavigation... to collect any complex but frequently used traversal pattern within the bowels of the nasty Block aggregate.

PR #757, opened 2020-04-04
Point project README to wiki

vishakh:
Since the project wiki now contains sufficient information in a simple and organized way, the readme file should not point to it instead of containing all the content itself. Hopefully, this will make the project easier to run for all future users.


README.md
```diff
@@ -1,311 +1,14 @@
...
-Query API for the Tezos blockchain
 
 [![Build Status](https://travis-ci.org/Cryptonomic/Conseil.svg?branch=master)](https://travis-ci.org/Cryptonomic/Conseil)
 [![Coverage Status](https://coveralls.io/repos/github/Cryptonomic/Conseil/badge.svg?branch=master)](https://coveralls.io/github/Cryptonomic/Conseil?branch=master)
 
-## Running Conseil
+Conseil is an indexer and API for blockchains, currently focused on the [Tezos](https://tezos.com/), which allows you to seamlessly run complex queries against blockchain data. It is written in the functional style in the [Scala](https://scala-lang.org/) programming language. This project forms an essential part of the [Cryptonomic stack](https://cryptonomic.tech/developers.html).
 
+Conseil consists of two processes:
+- Lorre, which indexes the blockchain by downloading data from blockchain nodes and storing it in databases.
```
piotrkosecki, line 9: `and stores it in database.` - I think that would be more accurate, Lorre uses one DB :)

piotrkosecki:
I think removing `lib/bitcoin-lib_2.12-0.9.15-SNAPSHOT.jar` broke the build.

README.md
```diff
@@ -1,311 +1,14 @@
...
+- Conseil, which provides an API into the blockchain data index by Lorre in databases.
```
piotrkosecki, line 10: I would also use here singular `database`.

vishakh:
> It looks great and a very positive improvement.
> 
> I would consider adding a "quick start" paragraph on the README file that at least provides instructions to run the stack in the most immediate and straightforward way (docker, probably)?

Thank you. It's difficult to have a quick start as a decent amount of set up is required. We will have to first make a dump from a testnet available first so the initial setup becomes easy.

PR #760, opened 2020-04-07
317 - Conseil logging incoming queries

### LINKED ISSUE #317: Conseil should log all incoming queries
All queries received by Conseil should be logged into a table called `api_queries`. The columns of this table should be:
- Timestamp
- IP address (if possible)
- URL
- Query body
- API key
- API version (v1 or v2)
- HTTP response code
- Processing time
- Result set size (if not too complicated to do)

piotrkosecki:
resolves #317 

The requirements evolved a bit since the original ticket was posted.
The idea is that the logs will be still read with Filebeat, but now it is possible to log to the console with the the jsons so it will be easier to parse.
I've introduced two env variables `ASYNC_LOGGING_LEVEL`(default INFO), `JSON_LOGGING_LEVEL` which is by default set to `OFF`.

@gites @bhaskarkishore please check if I missed something from our last meeting.


src/main/scala/tech/cryptonomic/conseil/util/RouteUtil.scala
```diff
@@ -0,0 +1,106 @@
...
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.server.directives.BasicDirectives
+import akka.http.scaladsl.server.{Directive, ExceptionHandler, RequestContext, Route}
+import akka.stream.Materializer
+import cats.effect.concurrent.MVar
+import cats.effect.{Concurrent, IO}
+import com.typesafe.scalalogging.LazyLogging
+import org.slf4j.MDC
+
+class RouteUtil(implicit concurrent: Concurrent[IO]) extends LazyLogging {
```
vishakh, line 16: Please document the purpose of this file.

piotrkosecki: Sure, added

src/main/resources/logback.xml
```diff
@@ -11,16 +11,26 @@
...
 
-
     <appender name="ASYNC" class="ch.qos.logback.classic.AsyncAppender">
+        <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
+            <level>${ASYNC_LOGGING_LEVEL:-INFO}</level>
+        </filter>
         <appender-ref ref="STDOUT" />
     </appender>
 
+    <appender name="JSON" class="ch.qos.logback.core.ConsoleAppender">
```
ivanopagano, line 21: This logger should block when a log call is requested, as opposed to the AsyncAppender

There are also [logstash specific ones](https://github.com/logstash/logstash-logback-encoder#async-appenders)

src/main/scala/tech/cryptonomic/conseil/util/RouteUtil.scala
```diff
@@ -0,0 +1,112 @@
...
+  /** Custom exception handler with MDC logging */
+  def loggingExceptionHandler(implicit correlationId: UUID): ExceptionHandler =
+    ExceptionHandler {
+      case e: Throwable =>
+        val response = HttpResponse(InternalServerError)
+
+        (for {
+          requestMap <- requestInfoMap.take
+          _ <- requestInfoMap.put(requestMap.filterNot(_._1 == correlationId))
+        } yield requestMap(correlationId).logResponse(response)).unsafeRunAsyncAndForget()
```
ivanopagano, line 52: this piece could be extracted to a 
```scala
type RequestMap = Map[UUID, RequestValues]
def requestMapModify[A](modify: RequestMap => RequestMap)(useValues: RequestValues => A) = for {
  map <- requestInfoMap.take
  _   <- requestInfoMap.put(modify(map))
} yield useValues(map(correlationId))

val response = HttpResponse(InternalServerError)

requestMapModify(
  modify = _.filterNot(_._1 == correlationId)
){ values => 
  values.logResponse(response)
}.unsafeRunAsyncAndForget()

```
and so on... wdyt?

src/main/scala/tech/cryptonomic/conseil/util/RouteUtil.scala
```diff
@@ -0,0 +1,112 @@
...
+class RouteUtil(implicit concurrent: Concurrent[IO]) extends LazyLogging {
+
+  private val requestInfoMap: MVar[IO, Map[UUID, RequestValues]] =
+    MVar.of[IO, Map[UUID, RequestValues]](Map.empty[UUID, RequestValues]).unsafeRunSync()
+
+  /** Directive adding recorded values to the MDC */
+  def recordResponseValues(
+      ip: RemoteAddress
+  )(implicit materializer: Materializer, correlationId: UUID): Directive[Unit] =
+    BasicDirectives.extractRequestContext.flatMap { ctx =>
```
ivanopagano, line 26: I guess we could simply `extractRequest` since that's what the context is used for.

src/main/scala/tech/cryptonomic/conseil/Conseil.scala
```diff
@@ -176,34 +181,43 @@ object Conseil
...
-                logRequest("Metadata Route", Logging.DebugLevel) {
-                  platformDiscovery.route
-                },
-                logRequest("Data Route", Logging.DebugLevel) {
-                  data.getRoute ~ data.postRoute
+          implicit val correlationId: UUID = UUID.randomUUID()
+          handleExceptions(loggingExceptionHandler) {
+            extractClientIP {
+              ip =>
+                recordResponseValues(ip)(mat, correlationId) {
```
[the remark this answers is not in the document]
piotrkosecki: yes, we need, as this is directive, if I removed the arguments, it would consider what happened in curly braces as those implicits

src/main/scala/tech/cryptonomic/conseil/util/RouteUtil.scala
piotrkosecki: sure, changed

src/main/scala/tech/cryptonomic/conseil/util/RouteUtil.scala
```diff
@@ -0,0 +1,112 @@
...
+import akka.stream.Materializer
+import cats.effect.concurrent.MVar
+import cats.effect.{Concurrent, IO}
+import com.typesafe.scalalogging.LazyLogging
+import org.slf4j.MDC
+
+/** Utility class for recording responses */
+class RouteUtil(implicit concurrent: Concurrent[IO]) extends LazyLogging {
+
+  private val requestInfoMap: MVar[IO, Map[UUID, RequestValues]] =
```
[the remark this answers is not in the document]
piotrkosecki: ok, I'll keep it in mind

src/main/scala/tech/cryptonomic/conseil/util/RouteUtil.scala
```diff
@@ -0,0 +1,112 @@
...
+  ) {
+    /** Logging response with MDC */
+    def logResponse(response: HttpResponse): Unit = {
+      MDC.put("httpMethod", httpMethod)
+      MDC.put("requestBody", requestBody)
+      MDC.put("clientIp", clientIp)
+      MDC.put("path", path)
+      MDC.put("apiVersion", apiVersion)
+      MDC.put("apiKey", apiKey)
+      val requestEndTime = System.currentTimeMillis()
```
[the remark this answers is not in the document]
piotrkosecki: ok, changed

src/main/resources/reference.conf
```diff
@@ -26,6 +26,14 @@ platforms {
...
       }
     }
+    localnet: {
+      node: {
+        protocol: "http",
+        hostname: "localhost",
+        port: 8732
+        pathPrefix: ""
+      }
+    }
```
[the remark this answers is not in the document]
piotrkosecki: my mistake

src/main/scala/tech/cryptonomic/conseil/util/RouteUtil.scala
piotrkosecki: ok, renamed

src/main/resources/logback.xml
piotrkosecki: this appender seems to be not working, why it should block? it should log like current one

ivanopagano: logback's regular appenders simply write to disk, which means that they block on a IO operation, blocking the current thread until it's done.

Akka docs advice against that because the actor system threads are precious (any thread can be blocking thousands of actors messages on queue)...
so they suggest using the [async appender](http://logback.qos.ch/manual/appenders.html#AsyncAppender) on top of the regular one, using a custom dispatcher to log messages.

ivanopagano: > This logger should block when a log call is requested, as opposed to the AsyncAppender

I actually meant that this console appender would block, not "should"

piotrkosecki: ok, I updated logback.xml

src/main/resources/logback.xml
```diff
@@ -11,16 +11,26 @@
         </encoder>
     </appender>
 
-
     <appender name="ASYNC" class="ch.qos.logback.classic.AsyncAppender">
+        <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
+            <level>${ASYNC_LOGGING_LEVEL:-INFO}</level>
```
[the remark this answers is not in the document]
piotrkosecki: added https://github.com/Cryptonomic/Conseil/wiki/Environment-configuration

gites:
@piotrkosecki  This looks really nice! I have only one question. Can you somehow flattened the structure of `requestBody` field? Parsing multiple levels of nested JSON will be a pain. Other than that LGTM.

piotrkosecki:
> @piotrkosecki This looks really nice! I have only one question. Can you somehow flattened the structure of `requestBody` field? Parsing multiple levels of nested JSON will be a pain. Other than that LGTM.

we don't parse request body here, we just put there what is in the request as string

gites:
> > @piotrkosecki This looks really nice! I have only one question. Can you somehow flattened the structure of `requestBody` field? Parsing multiple levels of nested JSON will be a pain. Other than that LGTM.
> 
> We don't parse request body here, we just put there what is in the request as string. Also request body is not really statically typed, so it would be especially hard to flatten it.

Ok, then I will work on adding some filters in logstash to for those fileds.
Would be nice to have JSON logging also in other backend applications.

piotrkosecki:
> > > @piotrkosecki This looks really nice! I have only one question. Can you somehow flattened the structure of `requestBody` field? Parsing multiple levels of nested JSON will be a pain. Other than that LGTM.
> > 
> > 
> > We don't parse request body here, we just put there what is in the request as string. Also request body is not really statically typed, so it would be especially hard to flatten it.
> 
> Ok, then I will work on adding some filters in logstash to for those fileds.
> Would be nice to have JSON logging also in other backend applications.

This will work for Lorre/Conseil - `logback.xml` is shared.

gites:
> > > > @piotrkosecki This looks really nice! I have only one question. Can you somehow flattened the structure of `requestBody` field? Parsing multiple levels of nested JSON will be a pain. Other than that LGTM.
> > > 
> > > 
> > > We don't parse request body here, we just put there what is in the request as string. Also request body is not really statically typed, so it would be especially hard to flatten it.
> > 
> > 
> > Ok, then I will work on adding some filters in logstash to for those fileds.
> > Would be nice to have JSON logging also in other backend applications.
> 
> This will work for Lorre/Conseil - `logback.xml` is shared.

I meant more lake Nautilus Cloud and Nautilus Cloud Metering ;)

PR #764, opened 2020-04-10
Refactor/split monolith into sbt modules

ajuszczak:
This pull request splits the project into separate modules:
- **conseil-api**, which is responsible for exposing data from the database
- **conseil-common**, which contains common parts between conseil-api and conseil-lorre
- **conseil-lorre**, which contains indexer for Tezos blockchain
- **conseil-lorre-schema**, which contains script to create Scala classes based on database
- **conseil-smoke-tests**, which contains integration tests (note, due to the fact that they are broken on master, this pull request **does not** fix that).

Note that upcoming changes will split this repository even more, into generic and Tezos-specific modules.

Please test this code as much as possible to avoid regression problems.

Edit: During this process new module has been added:
- **conseil-common-testkit**, which contains classes that can be re-used across different modules, generalized, for testing purposes only to avoid using `compile->test` in `sbt`, which was causing problems at multiple levels.


build.sbt
```diff
@@ -1,120 +1,110 @@
...
+          ".*\\.EndpointClient",
+          ".*\\.RegressionRun"
+        ).mkString(";")
+  )
+  .addRunCommand(description = "A run smoke tests Task.")
+  .disableAssembly()
 
-lazy val genSchema = taskKey[Unit]("A schema generating task.")
-fullRunTask(genSchema, Runtime, "tech.cryptonomic.conseil.scripts.GenSchema")
+addCommandAlias("runApi", "; api/runTask")
```
piotrkosecki, line 107: `runApi` does not seem to work for me

build.sbt
```diff
@@ -1,120 +1,110 @@
...
-  val args = spaceDelimited("").parsed
-  runInputTask(Runtime, "tech.cryptonomic.conseil.Conseil", args: _*).toTask("")
-}.evaluated
+lazy val lorreSchema = (project in file("conseil-lorre-schema"))
+  .settings(
+    name := "conseil-lorre-schema",
+    mainClass := Some("tech.cryptonomic.conseil.lorre.schema.GenSchema")
+  )
+  .addRunCommand(description = "A run schema generating Task.")
+  .enableAssembly()
```
piotrkosecki, line 90: I don't think we need assembly for that

build.sbt
```diff
@@ -1,120 +1,110 @@
...
+          "<empty>",
+          ".*\\.Lorre",
+          ".*\\.LorreAppConfig"
+        ).mkString(";")
+  )
+  .settings(Defaults.itSettings)
+  .configs(IntegrationTest)
+  .addRunCommand(
+    description = "A run application Task.",
+    javaExtras = Seq("-Xmx512M", "-Xss1M", "-XX:+CMSClassUnloadingEnabled")
```
piotrkosecki, line 79: I think Lorre needs bigger Xmx. for development I use 8192M

conseil-lorre/src/main/resources/application.conf
```diff
@@ -0,0 +1,90 @@
...
+    #Used to specify the max-time allowed for each accounts-page to finish processing
+    accountPageProcessingTimeout: 15 minutes
+
+    #Used to specify the max-time allowed for each delegate-page to finish processing
+    delegatePageProcessingTimeout: 15 minutes
+
+  }
+
+  # Custom libSodium settings
+  sodium.libraryPath: "/usr/lib/x86_64-linux-gnu/libsodium.so.18"
```
ivanopagano, line 78: This is not clear-cut at all... it seems to be used - commented-out - only in conseil-api, yet it's in conseil-lorre's config file...

Can we temporarily move the configuration to the conseil-common reference.conf file, and put it "top-level" instead of keeping it inside the `lorre` sub-tree?

project/BuildInfo.scala
```diff
@@ -0,0 +1,30 @@
...
+import sbt.Project
+import sbtbuildinfo.BuildInfoKeys._
+import sbtbuildinfo.{BuildInfoKey, BuildInfoPlugin}
+
+object BuildInfo {
+
+  implicit class BuildInfoOps(project: Project) {
+
+    //TODO Probably API and Lorre should have BuildInfo structure split across both of these modules
+    // Currently name is HardCoded, should be changed, once used by multiple modules
```
ivanopagano, line 12: Can't we simply read the name of the `Project` object?
In any case I agree with your analysis, yet not sure about splitting into separate module-specific objects.

project/Commands.scala
```diff
@@ -0,0 +1,28 @@
...
+
+object Commands {
+
+  implicit class CommandOps(project: Project) {
+    def addRunCommand(description: String, javaExtras: Seq[String] = Seq.empty): Project = {
+      import complete.DefaultParsers._
+
+      lazy val runTask = inputKey[Unit](description)
+
+      fork in runTask := false
```
ivanopagano, line 12: Could this be a parameter defaulting to `false`, so we could switch it at will in the main build-file?

project/Dependencies.scala
```diff
@@ -0,0 +1,103 @@
...
+    compilerPlugin("com.github.ghik" % "silencer-plugin" % silencerVersion cross CrossVersion.full),
+    "com.github.ghik" % "silencer-lib" % silencerVersion % Provided cross CrossVersion.full
+  )
+
+  val conseilCommonExclude = Seq(
+    "org.consensusresearch" %% "scrypto"
+  )
+
+  val conseilApiInclude = Seq(
+    "org.scalatest"     %% "scalatest"         % "3.0.5"         % "it, test",
```
ivanopagano, line 83: Isn't this somehow transitively inherited via module dependencies?

And the same for a few entries here-after...

build.sbt
ajuszczak: Well, I copied that from original code, so I don't know why it should be changed now?
This change may have bigger impact that we think, so I would rather suggest to change that later, in separate PR.

Old code for reference: https://github.com/Cryptonomic/Conseil/blob/master/build.sbt#L113

build.sbt
ajuszczak: Ok, I removed that, expect the change at GitHub today.

build.sbt
```diff
@@ -1,120 +1,110 @@
...
+          "tech.cryptonomic.conseil.common.config.Security"
+        ).mkString(";")
+  )
+  .settings(Defaults.itSettings)
+  .settings(
+    scalacOptions += "-P:silencer:pathFilters=common/src/main/scala/tech/cryptonomic/conseil/common/tezos/Tables.scala"
+  )
+  .configs(IntegrationTest)
+  .enableBuildInfo()
+  .disableAssembly()
```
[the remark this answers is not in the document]
ajuszczak: Previously, since there was only one project, entire project has been assembled so there was nothing to disable. I think that by default this plugin is disabled indeed, so I just removed this method and it works as expected.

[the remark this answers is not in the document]
ajuszczak: Ah, but you are asking about `disableAssembly()` at `conseil`, you may actually be right there. Let me investigate this part.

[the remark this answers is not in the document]
ajuszczak: I did investigated that, the conclusion is that, the plugin automatically enables assembly for all projects. Turning it off, at root level gives nothing. I managed to found the issue on GitHub for this problem [here](https://github.com/sbt/sbt-assembly/issues/175). There is no solution only workaround to disable plugin for specific modules.

I see different behavior on assembly. With the setup we have (but without disabling it at root, I'll remove it), we are only building modules we want. Otherwise, we are testing and assembling everything, which is not what we want.

build.sbt
```diff
@@ -1,120 +1,110 @@
...
-  "org.testcontainers"           % "postgresql"                     % "1.12.3" % "it, test",
-  "com.softwaremill.diffx"       %% "diffx-scalatest"               % "0.3.3" % "it, test",
-  "org.http4s"                   %% "http4s-blaze-client"           % http4sVersion % IntegrationTest,
-  "org.http4s"                   %% "http4s-dsl"                    % http4sVersion % IntegrationTest,
-  "org.http4s"                   %% "http4s-circe"                  % http4sVersion % IntegrationTest,
-  compilerPlugin("com.github.ghik" % "silencer-plugin" % silencerVersion cross CrossVersion.full),
-  "com.github.ghik" % "silencer-lib" % silencerVersion % Provided cross CrossVersion.full
-)
+lazy val conseil = (project in file("."))
+  .disableAssembly()
```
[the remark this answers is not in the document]
ajuszczak: I removed that, since it brings nothing to the table. See [below comment](https://github.com/Cryptonomic/Conseil/pull/764#discussion_r410084985) for more detailed explanation.

build.sbt
ivanopagano: I guess we're overriding those values locally anyway.

@piotrkosecki are we sure about the relevant values needed here?
In any case we might adapt them after this PR if needed.

conseil-api/src/main/resources/logback.xml
```diff
@@ -17,7 +17,7 @@
     </appender>
 
     <logger name="com.base22" level="TRACE"/>
-    <logger name="tech.cryptonomic.conseil.tezos.BigMapsOperations" level="${LORRE_BIG_MAPS_LOG_LEVEL:-OFF}"/>
+    <logger name="slick.jdbc.JdbcBackend.statement"  level="INFO"/>
```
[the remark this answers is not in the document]
ajuszczak: Not really. Previously there were 2 places to configure logging. For Slick it was `application.conf` (or `reference.conf`) for rest of the classes it was `logback.xml`. In my opinion logbook is a better place to configure logging (since you are configuring everything from single file, single place) I changed the logLevel in `reference.conf` and limited it at this stage. This is the reason for this line. Would you like me to revert it back as it was? I do not observe any difference in behavior at the end of the day, so I suppose it works as previously (for the end user).

conseil-api/src/main/resources/metadata.conf
```diff
@@ -27,9 +27,9 @@ metadata-configuration {
   tezos {
     visible: true
     networks {
-      mainnet { include "metadata/tezos.mainnet.conf" }
-      babylonnet { include "metadata/tezos.babylonnet.conf" }
-      carthagenet { include "metadata/tezos.carthagenet.conf" }
+      mainnet { include "metadata/tezos.mainnet.conf"}
+      babylonnet { include "metadata/tezos.babylonnet.conf"}
+      carthagenet { include "metadata/tezos.carthagenet.conf"}
```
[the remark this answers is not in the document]
ajuszczak: I believe this is formatted in the same way it was in [original file](https://github.com/Cryptonomic/Conseil/blob/master/src/main/resources/metadata.conf).

conseil-common/src/main/resources/reference.conf
```diff
@@ -0,0 +1,76 @@
...
+  # for available properties and their meaning
+  #
+  # The current configuration is based on local benchmarking against zeronet
+  # To improve or check it, look at the tech.cryptonomic.conseil.NodeStreamingBenchmark app
+  # On the streaming http client pool we expect a max of:
+  #   45 connections x
+  #    7 requests/conn ~=
+  #  315 ongoing requests at each moment
+  # The pipelining on each connection might slow down for slow responses, but they should be rare
+  tezos-streaming-client { //TODO Used by Lorre only, but with tests in common
```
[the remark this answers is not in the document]
ajuszczak: Yea, renaming might improve reasoning about that. Waiting for rest of the Team.

build.sbt
```diff
@@ -1,120 +1,110 @@
...
+  .enableAssembly()
+  .dependsOn(common % "compile->test")
 
-scalacOptions ++= ScalacOptions.common
-scalacOptions += "-P:silencer:pathFilters=src/main/scala/tech/cryptonomic/conseil/tezos/Tables.scala"
-import complete.DefaultParsers._
+lazy val lorre = (project in file("conseil-lorre"))
+  .settings(
+    name := "conseil-lorre",
+    mainClass := Some("tech.cryptonomic.conseil.lorre.Lorre"),
```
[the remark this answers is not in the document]
ajuszczak: You are talking about renaming package `lorre` -> `indexer`? I left that, because with Conseil-api we have similar story, but we decided to name it `api` at this stage. if we want to rename it to `indexer` we should name this module like that too, to keep consistency, I think.

build.sbt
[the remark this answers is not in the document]
ivanopagano: good to know, thanks.

build.sbt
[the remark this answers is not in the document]
ivanopagano: I think I see your point, fine for me to leave it like this.

Simply I didn't like the Lorre packaging since it doesn't convey meaning. Product/project-like names are usually for top level packaging in my head, and that part is already reserved for `conseil`.

build.sbt
```diff
@@ -1,120 +1,110 @@
...
+        ).mkString(";")
+  )
+  .settings(Defaults.itSettings)
+  .configs(IntegrationTest)
+  .addRunCommand(
+    description = "A run application Task.",
+    javaExtras = Seq("-Xms1024M", "-Xmx8192M", "-Xss1M", "-XX:+CMSClassUnloadingEnabled")
+  )
+  .enableAssembly()
+  .dependsOn(common % "compile->test")
```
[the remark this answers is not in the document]
ajuszczak: Honestly, I don't know and I can read about that more, but that's not the reason why I decided to use it. The main reason is that there are classes in `test` folder in `common` which are used in `test` folder in `api`. To make `common` available in `test` we need to apply above configuration.

I'm talking for example about: `tech.cryptonomic.conseil.common.metadata.TestPlatformDiscoveryOperations` which is used in `common` as well as in `api`. If you know better way of making in accessible in both, let me know. I do not know any, different than creating yet another module for it, but that would be an overkill.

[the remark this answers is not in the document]
ajuszczak: I checked that. Looks like `sbt api/test` will trigger `common/compile`, `api/compile`, `api/test`. While `sbt common/compile` will trigger `common/compile` only. Once the `common` is already compiled, we are not going to compile it again, we will only compile tests for `common` and run tests for `api`.

build.sbt
```diff
@@ -1,120 +1,110 @@
...
-fork in runConseil := true
-javaOptions in runConseil ++= Seq("-Xms1024M", "-Xmx8192M", "-Xss1M", "-XX:+CMSClassUnloadingEnabled")
-runConseil := Def.inputTaskDyn {
-  val args = spaceDelimited("").parsed
-  runInputTask(Runtime, "tech.cryptonomic.conseil.Conseil", args: _*).toTask("")
-}.evaluated
+lazy val lorreSchema = (project in file("conseil-lorre-schema"))
+  .settings(
+    name := "conseil-lorre-schema",
+    mainClass := Some("tech.cryptonomic.conseil.lorre.schema.GenSchema")
```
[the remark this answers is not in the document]
ajuszczak: You are right. It does not belongs to `Lorre` more to `Tezos` right now, so I can drop `Lorre` part from the name and package if that is ok for you.

conseil-api/src/main/resources/application.conf
```diff
@@ -0,0 +1,34 @@
+include "metadata.conf"
```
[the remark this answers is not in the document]
ajuszczak: Actually, currently we have 3 types of configuration:
- `reference.conf` from the `common`,
- `application.conf` from the `api`/`lorre`,
- `developer.conf` from the "outside".

If you want to use `application.conf` while running either of two mentioned modules, what you need to do is simply add `include application.conf` at the very beginning and then you can overwrite whatever `application.conf` brings (which additionally brings `metadata.conf` so you can overwrite it as well in your `developer.conf` that is the idea behind that. I believe I tested that already but feel free to do it as well, if you have some spare time.

conseil-common/src/main/resources/reference.conf
```diff
@@ -0,0 +1,76 @@
...
+    # essentially keep connections alive across lorre's cycles
+    idle-timeout: 10 minutes
+    pipelining-limit: 7
+    # give more room for async response in head-of-line blocking on the same connection or other slow responses
+    response-entity-subscription-timeout: 60 seconds
+  }
+
+  # this is essentially available to enable composition of database operations
+  # current configuration is based upon the default-blocking-io-dispatcher in akka
+  tezos-dispatcher { //TODO Used by Conseil only
```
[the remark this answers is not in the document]
ajuszczak: Yes, that's correct. I'm going to move that.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilAppConfig.scala
```diff
@@ -1,7 +1,8 @@
-package tech.cryptonomic.conseil.config
+package tech.cryptonomic.conseil.api
```
ivanopagano, line 1: The `argParser` is built with the wrong name ("conseil"), that will show on the console prompt. it should be renamed "conseil-api" for consistency with how the program is run

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/lorre/LorreAppConfig.scala
```diff
@@ -1,14 +1,15 @@
-package tech.cryptonomic.conseil.config
+package tech.cryptonomic.conseil.lorre
```
ivanopagano, line 1: The `argParser` is built with the wrong name ("lorre"), that will show on the console prompt. it should be renamed "conseil-lorre" for consistency with how the program is run

build.sbt
ivanopagano: I've found the app is not actually starting, it completes immediately.
I'm investigating to find the reason.

conseil-common/src/main/resources/reference.conf
```diff
@@ -0,0 +1,76 @@
...
+    type: "Dispatcher"
+    executor: "thread-pool-executor"
+    throughput: 1
+
+    thread-pool-executor {
+      fixed-pool-size: 16
+    }
+  }
+
+  http { //TODO Only used by Conseil, cause Lorre does not expose API
```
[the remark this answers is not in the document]
ajuszczak: Changed

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/FeeOperations.scala
```diff
@@ -12,6 +11,8 @@ import scala.util.{Failure, Success}
   * Helper classes and functions used for average fee calculations.
   */
 object FeeOperations extends LazyLogging {
+  // TODO: This was originally taken from Lorre.db, but we do not have access there.
```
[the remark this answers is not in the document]
ajuszczak: I'll remove TODO, but I'll leave a comment about that.

conseil-lorre/src/main/resources/application.conf
```diff
@@ -0,0 +1,90 @@
+lorre {
```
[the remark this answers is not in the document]
ajuszczak: I believe that we cannot have two `reference.conf` loaded in the classpath, but I'll investigate that. I responded how this case can be managed somewhere above. If you are not satisfied with my answer let me know and open discussion once again :)

conseil-lorre/src/main/resources/application.conf
ajuszczak: Yea, I took it from [here](https://github.com/Cryptonomic/Conseil/blob/master/src/main/resources/reference.conf#L144), but I can move that wherever you want. I'll move that to the top-level without any section.

project/BuildInfo.scala
ajuszczak: Not really, because name from the project would be different than "Conseil". It will be "Conseil-api" and "Conseil-lorre" which in my opinion is exactly how it should work, but since I wanted to reduce number of changes I decided to hardcode that. We **may** introduce different version for API that Lorre part, but here, it will be under the same name, so it might confuse community, since asking API for version will give different value than Lorre for the same project name.

build.sbt
ajuszczak: That might be issue with configuration file. I noticed this couple of times, that the match with `Failure` which contains comment //nothing to do, actually does not return anything besides -1.

conseil-lorre-schema/src/main/scala/tech/cryptonomic/conseil/lorre/schema/GenSchema.scala
```diff
@@ -1,4 +1,4 @@
-package tech.cryptonomic.conseil.scripts
+package tech.cryptonomic.conseil.lorre.schema
```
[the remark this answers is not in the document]
ajuszczak: Yea, we have been talking about that in different comment I believe, to rename the whole module to make it not lorre-related, is that correct?

build.sbt
ivanopagano: @ajuszczak where exactly?

build.sbt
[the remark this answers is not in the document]
ivanopagano: I think I understand that `compile` will only compile the dependent configuration, and not run tests.

Did you instead try using 
`.dependsOn(common % "test -> test")`
?

That should make the test in conseil-api depend on those in conseil-common but no compilation dependency on the main classes

build.sbt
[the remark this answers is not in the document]
ivanopagano: sure, go with it

conseil-api/src/main/resources/application.conf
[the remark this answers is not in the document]
ivanopagano: It seems to make sense, I'm only worried that we tried that in the past and had issues.

But I might be mistaken, we could've missed the `import`?? 🤔 

Let's try it like this and be hopeful.

conseil-api/src/main/resources/logback.xml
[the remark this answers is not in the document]
ivanopagano: Now I understand the slick entry on the configuration file better... it attracted my attention but I didn't recall it well. Now it makes sense.

Fine for me.

conseil-api/src/main/resources/application.conf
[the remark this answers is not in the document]
ivanopagano: Ok, I guess we must make sure that the docs clearly explain how to do this, to avoid users will get confused.

The custom application.conf file should start with a strict requirement to include the module configuration with a line like
```
include: required(classpath("application"))
```

This specifies that the application file to include is expected as a classpath resource, and that it's required, so the startup will fail if something is not where it should be.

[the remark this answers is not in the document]
ivanopagano: Should the same be needed, when building the docker image, which should provide all needed values in the included application file, bundled via the template generation?
Could you please double-check on that @ajuszczak ?

build.sbt
ivanopagano: Ok, I think I found the problem here.
The included configuration has no `conseil.security.apiKeys.keys` entry defined.

If I include a value with my custom application.conf, the app starts correctly (even though I suspect logging might not be properly configured, as nothing gets printed when conseil-api starts.
This could explain the fact that no error message was logged when failing to run properly.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Security.scala
```diff
@@ -1,4 +1,4 @@
-package tech.cryptonomic.conseil.config
+package tech.cryptonomic.conseil.common.config
```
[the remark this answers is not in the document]
ajuszczak: Yes, for the time being I decided to put it into common. Most likely will be split later on in the  next iteration of refactoring.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/io/MainOutputs.scala
```diff
@@ -1,9 +1,9 @@
-package tech.cryptonomic.conseil.io
+package tech.cryptonomic.conseil.common.io
```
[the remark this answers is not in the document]
ajuszczak: Done

build.sbt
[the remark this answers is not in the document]
ajuszczak: I did changed that, so we have `.indexer.` instead of `Lorre`

build.sbt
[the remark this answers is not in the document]
ajuszczak: Yes, I tried to do that (actually all possible, meaningful configurations). Only the existing one solved the issue.

conseil-api/src/main/resources/application.conf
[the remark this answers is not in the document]
ajuszczak: Yes, I can check docker out. I believe that in docker's template I'm already importing `application.conf`.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/JsonDecoders.scala
```diff
@@ -1,8 +1,8 @@
-package tech.cryptonomic.conseil.tezos
+package tech.cryptonomic.conseil.common.tezos
```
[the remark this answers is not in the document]
ajuszczak: I managed to do that. Thanks!

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/TezosNodeInterface.scala
```diff
@@ -1,4 +1,4 @@
-package tech.cryptonomic.conseil.tezos
+package tech.cryptonomic.conseil.common.tezos
```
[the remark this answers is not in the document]
ajuszczak: This is almost possible. Currently `NodeStreamingBenchmarks` uses this. Since this class belongs to `Schema` module, it does not have (and should not have) dependency on Lorre. I think we cannot move it yet.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/TezosNodeOperator.scala
```diff
@@ -1,19 +1,19 @@
...
+import tech.cryptonomic.conseil.common.tezos.TezosTypes.Lenses._
+import tech.cryptonomic.conseil.common.tezos.michelson.JsonToMichelson.toMichelsonScript
+import tech.cryptonomic.conseil.common.tezos.michelson.dto.{MichelsonInstruction, MichelsonSchema}
 import cats.instances.future._
 import cats.syntax.applicative._
-import tech.cryptonomic.conseil.generic.chain.DataFetcher.fetch
-import tech.cryptonomic.conseil.tezos.TezosNodeOperator.FetchRights
-import tech.cryptonomic.conseil.tezos.TezosTypes.{BakingRights, EndorsingRights}
+import tech.cryptonomic.conseil.common.generic.chain.DataFetcher.fetch
+import tech.cryptonomic.conseil.common.tezos.TezosNodeOperator.FetchRights
```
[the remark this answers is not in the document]
ajuszczak: `TezosNodeOperator` has been moved to `conseil-lorre` module, so yes it is possible :)

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/ApiOperations.scala
```diff
@@ -1,22 +1,22 @@
-package tech.cryptonomic.conseil.tezos
+package tech.cryptonomic.conseil.common.tezos
```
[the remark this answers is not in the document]
ajuszczak: No issue at all, but this step will be actually implemented in next iterator most likely. Let me think which moment is the best to do that.

conseil-common/src/test/scala/tech/cryptonomic/conseil/common/tezos/TezosResponseBuilder.scala
```diff
@@ -1,4 +1,4 @@
-package tech.cryptonomic.conseil.tezos
+package tech.cryptonomic.conseil.common.tezos
```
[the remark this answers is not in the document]
ajuszczak: Moved.

conseil-lorre-schema/src/main/scala/tech/cryptonomic/conseil/lorre/schema/GenSchema.scala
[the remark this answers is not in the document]
ajuszczak: I renamed the module and removed `Lorre` from packages.

project/BuildInfo.scala
ivanopagano: Well, I see that we need to figure out how to handle the versioning if modules can be published separately and evolve as such...
It's a whole new level of complexity that I wouldn't add now.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/TezosNodeInterface.scala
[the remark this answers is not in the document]
ivanopagano: The benchmarks are there for historical purpose only, and I suggest you remove them completely if they block us on cleanly separating the modules.
Especially they would need to be refactored anyway to work correctly, I presume, so having them blocking this process makes no sense.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/ApiOperations.scala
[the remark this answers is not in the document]
ivanopagano: Fine for me, please take note of this tho, so we don't have to make the same analysis again later on.

project/BuildInfo.scala
ajuszczak: All right, just wanted to point out my reasoning. I also think we should go with current behavior for the time being.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/TezosNodeInterface.scala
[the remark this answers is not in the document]
ajuszczak: For some reason, I was thinking that these benchmarks are part of the `GenSchema` process. I'll remove them and move above class as well.

build.sbt
ivanopagano: @ajuszczak do you plan to do any change to the `developer.conf` file, so that the missing keys will be available there?

Also, did you check if the logging config is fine on conseil-api?

build.sbt
piotrkosecki: I remember when I tried running Lorre for longer time to get some data for testing I needed to use bigger Xmx because it crashed relatively often. TBH I'm not sure how it works right now, as I'm always using those values.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/ApiOperations.scala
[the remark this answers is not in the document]
ajuszczak: I think I've managed to do that.

build.sbt
ajuszczak: I can, but as far as I know, this problem was before my changes, so we should treat it more as an improvement. Actually, I think "both" issues are from the past, but that does not change that I'm going to look into it in a second.

build.sbt
ajuszczak: We agreed that we are going to change that if needed in upcoming pull-requests.

build.sbt
ivanopagano: Thanks for the good work, I guess we can
 * re-think how to present the developer.conf differently and not require it along the import chain, but to tell it's a template to start from
 * simply add the `required(classpath(application))` to developer.conf and suggest the developer to import `developer.conf` as it works right now

ivanopagano: Regarding logging, there was never such issue in the current  development branch, since the logback file worked for all applications just the same way.
There might be something trivial like a misspelling of a package name or the logback path... I didn't look into that

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/Conseil.scala
```diff
@@ -133,18 +126,18 @@ object Conseil
...
-      metadataOverrides: MetadataConfiguration,
-      securityApi: SecurityApi,
-      verbose: VerboseOutput
+                 metadataService: MetadataService,
+                 conseilOperations: ConseilOperations,
+                 server: ServerConfiguration,
+                 platforms: PlatformsConfiguration,
+                 metadataOverrides: MetadataConfiguration,
+                 securityApi: SecurityApi,
+                 verbose: VerboseOutput
```
ivanopagano, line 135: formatting went awry

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilOperations.scala
```diff
@@ -0,0 +1,104 @@
+package tech.cryptonomic.conseil.api
```
ivanopagano, line 1: 😍

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/DataTest.scala
```diff
@@ -154,7 +154,7 @@ class DataTest
       cacheOverrides,
       platformDiscoveryOperations
     )
-  val apiOps: ApiOperations = new ApiOperations
+  val apiOps: ConseilOperations = new ConseilOperations
```
ivanopagano, line 157: makeup: rename the val ?

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/SqlOperations.scala
```diff
@@ -0,0 +1,85 @@
+package tech.cryptonomic.conseil.common.tezos
```
ivanopagano, line 1: 🤩

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/SqlOperations.scala
```diff
@@ -0,0 +1,85 @@
...
+  Predicate,
+  Query,
+  QueryResponse,
+  SimpleField
+}
+import tech.cryptonomic.conseil.common.util.DatabaseUtil
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object SqlOperations {
```
ivanopagano, line 17: We'd need to add comments here for completeness, even a reminder will do for now, for our team to fill later

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilOperations.scala
```diff
@@ -0,0 +1,104 @@
...
+
+import com.github.ghik.silencer.silent
+import slick.jdbc.PostgresProfile.api._
+import tech.cryptonomic.conseil.common.tezos.SqlOperations.{AccountResult, BlockResult, OperationGroupResult}
+import tech.cryptonomic.conseil.common.tezos.TezosTypes.{AccountId, BlockHash}
+import tech.cryptonomic.conseil.common.tezos.{SqlOperations, Tables, TezosDatabaseOperations => TezosDb}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class ConseilOperations extends SqlOperations {
```
ivanopagano, line 11: We'd need to add comments here for completeness, even a reminder will do for now, for our team to fill later

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosNodeOperator.scala
```diff
@@ -65,7 +64,7 @@ class TezosNodeOperator(
     val node: TezosRPCInterface,
     val network: String,
     batchConf: BatchFetchConfiguration,
-    apiOperations: ApiOperations
+    sqlOperations: LorreOperations
```
ivanopagano, line 67: Naming here might be more consistent as `lorreOperations` since the type dependency is explicit.
Seems no use to hide the fact that we have that dependency here with a more generic name.
We might figure out how to better separate the things in future reworks.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosNodeOperator.scala
```diff
@@ -698,7 +697,7 @@ class TezosNodeSenderOperator(
     network: String,
     batchConf: BatchFetchConfiguration,
     sodiumConf: SodiumConfiguration,
-    apiOperations: ApiOperations
+    apiOperations: LorreOperations
```
ivanopagano, line 700: Same reasoning as above.

We could choose `apiOperations` in all places for this file, it makes some sense.

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/DataTest.scala
ajuszczak: `conseilOps` it is

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilOperations.scala
ajuszczak: Added similar to `SqlOperations`, if you find that incorrect, please update that later :)

conseil-api/src/main/resources/developer.conf
```diff
@@ -1,3 +1,4 @@
 # Configuration file for local development.
+include required(classpath("application.conf"))
```
ivanopagano, line 2: are you sure that you need `"application.conf"` and not simply `"application"` ?

docker/template.conf
```diff
@@ -1,6 +1,4 @@
-include "application.conf"
-
-logger.scala.slick = INFO
+include required(classpath("application.conf"))
```
ivanopagano, line 1: Here too, is the `.conf` correct?

Note: I use the include without it, but it may well work both ways, and I didn't try it.

build.sbt
ajuszczak: Issue was a bit more complicated. Due to the fact that in build.sbt we had "compile->test" for `common` module, Logback (according to documentation) first picked `Logback-test.xml` file (provided by `common` module due to mentioned operation). To prevent that, I had to get rid of `compile->test` operation, by introducing module as a TestKit. Due to the fact that not all of the classes shared across tests can be generalized and put into `conseil-common-testkit` I decided to temporarily copy them and I'll make it right, once modules will be properly split (specific TODO has been added to it can be identified).

conseil-api/src/main/resources/developer.conf
ajuszczak: Both works. I don't need it, but since this file is managed by "us", I don't see the reason to make this change. Extension is not going to change either way, right?

docker/template.conf
ajuszczak: Well, I'm using what was there before actually :D so that's not mine, it was here before and I don't want to do to much changes in that regard, which will change actually nothing.

build.sbt
```diff
@@ -1,121 +1,111 @@
...
+        ).mkString(";")
+  )
+  .addRunCommand(description = "Task to run smoke tests locally")
+  .disableAssembly()
 
-lazy val genSchema = taskKey[Unit]("A schema generating task.")
-fullRunTask(genSchema, Runtime, "tech.cryptonomic.conseil.scripts.GenSchema")
+addCommandAlias("runApi", "; api/runTask")
+addCommandAlias("runLorre", "; lorre/runTask")
+addCommandAlias("runLorreSchema", "; lorreSchema/runTask")
```
ivanopagano, line 110: Is that `lorreSchema` or just `schema`?

conseil-schema/src/main/resources/logback.xml
```diff
@@ -0,0 +1,37 @@
...
+
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder class="ch.qos.logback.core.encoder.LayoutWrappingEncoder">
+            <layout class="ch.qos.logback.classic.PatternLayout">
+                <Pattern>%d{HH:mm:ss.SSS} %-53([%thread] %-5level) %-36logger{36} - %msg%n</Pattern>
+            </layout>
+        </encoder>
+    </appender>
+
+    <appender name="JSON" class="ch.qos.logback.core.ConsoleAppender">
```
ivanopagano, line 14: This shouldn't be needed here?
It's an operational concern only for conseil-api?

Schema generation should only need a basic appender

conseil-schema/src/main/resources/logback.xml
```diff
@@ -0,0 +1,37 @@
...
+    </appender>
+
+    <appender name="ASYNC" class="ch.qos.logback.classic.AsyncAppender">
+        <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
+            <level>${ASYNC_LOG_LEVEL:-INFO}</level>
+        </filter>
+        <appender-ref ref="STDOUT" />
+    </appender>
+
+    <appender name="ASYNC_JSON" class="ch.qos.logback.classic.AsyncAppender">
```
ivanopagano, line 25: Same considerations here.

conseil-schema/src/main/resources/logback.xml
ajuszczak: You are right, I copied that too fast.

ivanopagano:
@ajuszczak you should now have an updated working smoke tests module in `master`.
If you care to give it a try, it might come useful.

You need to have a tezos node to refer to, which contains the first 5000 blocks from carthagenet at least.

Warning, local nodes will keep only a limited number of levels preceding the `head` (around 7-8k), so having a node which goes too far ahead will not work, since oldest blocks will not be available.

I need to put some wiki entry to record such info and how to workaround that...

ivanopagano:
I see a significant decrease in coverage, do you think it's because we've duplicated some parts and now they're less covered, or maybe did we lose some ignore in the coverage conf, that should be resumed, for code which is known to be not relevant for testing?

ajuszczak:
@ivanopagano I'm pretty sure it's about duplicates, because I've solved the issue with wrong packages being skipped during coverage check.

PR #768, opened 2020-04-20
724 - added operation_group_id to the big_map_contents table

### LINKED ISSUE #724: big_map_contents operation reference
It would be useful to include the operation group id responsible for the value update.

piotrkosecki:
resolves #724


src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -686,22 +686,25 @@ object DatabaseConversions extends LazyLogging {
...
-          val internalOperationData = internalResults.map { oop =>
-            val op = oop.convertTo[Operation]
-            (from, group.hash, op)
-              .convertTo[Tables.OperationsRow]
-              .copy(internal = true, nonce = Some(oop.nonce.toString)) -> op
-              .convertToA[List, Tables.BalanceUpdatesRow]
-          }
-          mainOperationData ++ internalOperationData
-      }.toList
+      TezosOptics.Blocks
```
ivanopagano, line 689: Any change here?

piotrkosecki: no, I don't think so, just `scalafmt` things 😄

ivanopagano: Ok, I've found the reformat was already on previous PRs, too.
Possibly rebasing on master might remove it?

In any case it seems quite irrelevant apart from being irritating 😁

PR #771, opened 2020-04-22
Tests: update the broken regression tests

ivanopagano:
Update the regression tests, available in the `it` sub-folder, which were currently broken.

The tests are slightly re-designed to work with fewer blocks to sync from the node, to run faster, and allows to customize more by passing the configuration file as an argument.

This will also remove the hard-coded apiKeys, that obviously won't be portable if trying to change networks.

The current tests work against the carthage network and requires 5k blocks to be downloaded.
This is tailored to allow a local node to be spun and resynced in a short time to speed up the indexing phase.
There's a lot of work to be done if we want to allow switching tests to multiple networks via arguments.

This PR is relevant make basic checks for the modular split of #764 

I additionally provided instructions on how to run such smoke tests in the wiki section on development.


src/it/scala/tech/cryptonomic/conseil/client/EndpointClient.scala
```diff
@@ -17,44 +17,76 @@ import io.circe.{parser, Json}
 import cats.effect.IO
 import cats.syntax.all._
 import scala.io.Source
```
piotrkosecki, line 19: Source is no longer used

src/it/scala/tech/cryptonomic/conseil/client/RegressionRun.scala
```diff
@@ -5,25 +5,45 @@ import cats.effect.{ExitCode, IO}
...
     * be empty.
     * Otherwise a previously filled instance from this same application can be re-used to
     * dramatically speed up the process.
+    * This last optimization makes sense only if you're testing regressions against the Api changes.
+    * Changing anything in the Lorre indexer would imply you need to re-run the indexer to check nothing
+    * got broken!
     */
   override def run(args: List[String]): IO[ExitCode] = {
-    val network = args.headOption
+    val defaultConfigFileName = "conseil-regression-tests.conf"
```
piotrkosecki, line 28: is this file provided anywhere?

src/it/scala/tech/cryptonomic/conseil/client/EndpointClient.scala
ivanopagano: Thanks, will remove

src/it/scala/tech/cryptonomic/conseil/client/RegressionRun.scala
ivanopagano: No, it's simply a default filename given as default if you don't provide it as command argument

ivanopagano: I've essentially decided to require such file, so that it can be used to read the apiKey, which was previously a hardcoded constant (didn't work very well of course)

PR #774, opened 2020-04-24
762 - Governance table should have records for proposal periods

### LINKED ISSUE #762: Governance table should have records for proposal periods
The `governance` table is not populated for blocks with governance period `proposal`. We should add such records so we have a complete dataset. 

These records will not have any data for any of the rolls or count columns. `proposal_hash` can be calculated in the following manner:
- Get the cycle for the current block
- For the given cycle, fetch all proposal operations
- For these operations, find all unique proposal hashes
- Covert these hashes into a comma delimited list.

piotrkosecki:
resolves #762

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -464,7 +464,7 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
         ballots: Map[Block, List[Voting.Ballot]],
         ballotCountsPerCycle: Map[Block, Voting.BallotCounts],
         ballotCountsPerLevel: Map[Block, Voting.BallotCounts],
-        proposalHashes: Map[Block, List[String]]
+        proposalHashes: Map[Block, Map[String, Int]]
     ): List[
       (
           BlockData,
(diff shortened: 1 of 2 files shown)


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -469,15 +475,28 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
       val listing = listings.get(block.data.header.level).toList.flatten
       val prevListings = listings.get(block.data.header.level - 1).toList.flatten
       val listingByBlock = listing.diff(prevListings)
       val ballot = ballots.get(block).toList.flatten
       val ballotCountPerCycle = ballotCountsPerCycle.get(block)
       val ballotCountPerLevel = ballotCountsPerLevel.get(block)
-      (block.data, proposal, listing, listingByBlock, ballot, ballotCountPerCycle, ballotCountPerLevel)
+      val proposalHashesForBlock = proposalHashes.get(block).toList.flatten
+
+      List((block.data, proposal, listing, listingByBlock, ballot, ballotCountPerCycle, ballotCountPerLevel)) ++
```
ivanopagano, line 488: Instead of building a singleton List with a single value and concatenate that with the proposal's list data, It might be easier to prepend the fist element to the rest with the cons `::` operator?

`(block.data, proposal, ...) :: proposalHashesFroBlock.map {...}` ?

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -626,6 +626,19 @@ object TezosDatabaseOperations extends LazyLogging {
...
+    Tables.Operations
+      .filter(op => op.kind === "proposals" && op.cycle === cycle)
+      .map(_.proposal)
+      .distinct
+      .result
+      .map(_.toList.flatMap(_.toList.flatMap(expandStringArray)))
+
+  /** Helper method for expanding array storred as a string to list */
+  private def expandStringArray(arr: String): List[String] =
+    arr.trim.stripPrefix("[").stripSuffix("]").split(',').toList
```
ivanopagano, line 640: What if this fails?

piotrkosecki: which part could fail here?

ivanopagano: ok, sorry, you're right, it'll spit a list of strings anyhow 👍

PR #779, opened 2020-04-29
766 - make balance updates easier to use

### LINKED ISSUE #766: Make balance_updates easier to use
- [ ] The `contract` column of `balance_updates` contains nulls. It should always be populated so we can get all balance updates for a given account ID. Where the column has null values, the value of the `delegate` column should be copied over. 

- [ ] `contract` should be renamed to `account_id` to make the terminology consistent with the other tables.

- [ ] The `delegate` column should be removed if it does not add any information beyond what `account_id` does.

- [ ] `block_id`, `block_level`, `cycle` and `period` should be added to the table.

- [ ] It's possible to view balance updates data in Arronax because there are no indexes to help with the display. We should index the following columns in the table:
- `account_id`
- `block_level`

Reference: https://github.com/Cryptonomic/Conseil/wiki/The-balance_updates-table

piotrkosecki:
resolves #766

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
@@ -549,7 +549,7 @@ object DatabaseConversions extends LazyLogging {
     */
   implicit def anyToBalanceUpdates[T, Label](
       implicit
-      balances: Getter[T, Map[Label, List[BlockTagged[OperationMetadata.BalanceUpdate]]]],
+      balances: Getter[T, Map[BlockTagged[Label], List[OperationMetadata.BalanceUpdate]]],
       hashing: Getter[T, Option[String]],
       showing: Show[Label]
   ) = new Conversion[List, T, Tables.BalanceUpdatesRow] {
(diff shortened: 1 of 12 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -561,24 +561,34 @@ object DatabaseConversions extends LazyLogging {
...
+                    level
+                  )
                   ) =>
                 Tables.BalanceUpdatesRow(
                   id = 0,
                   source = tag.show,
                   sourceHash = hashing.get(from),
                   kind = kind,
-                  contract = contract.map(_.id),
+                  accountId = contract.map(_.id).orElse(delegate.map(_.value)).get, // either contract or delegate should be always set
```
ivanopagano, line 584: I suggest we program defensively, given that the balance updates is data coming from external sources that we can't control.
If both those values are missing, possibly for a breaking version update, the whole process will fail and stop Lorre until it's fixed.

We might choose to log an error if the value is missing and put a placeholder (empty string or "N/A"?) here.

It depends on how critical to business is the fact that this reference is absolutely filled correctly
@vishakh ?

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -543,13 +543,13 @@ object DatabaseConversions extends LazyLogging {
...
     * @tparam T the source type that contains the balance updates values
-    * @tparam Label the source Label type, which has a `Show` contraint, to be representable as a string
+    * @tparam Label the source Label type, which has a `Show` constraint, to be representable as a string
     * @param balances an optic instance that lets extract the Map of tagged updates from the source
     * @param hashing an optic instance to extract an optional reference hash value
     */
   implicit def anyToBalanceUpdates[T, Label](
       implicit
-      balances: Getter[T, Map[Label, List[OperationMetadata.BalanceUpdate]]],
+      balances: Getter[T, Map[Label, List[BlockTagged[OperationMetadata.BalanceUpdate]]]],
```
ivanopagano, line 552: Wouldn't it be more efficient to "tag" the `Label` here?
That would be made once per block, for operations and header, instead of replicating the same data for each update.

```scala
balances: Getter[T, Map[BlockTagged[Label], List[OperationMetadata.BalanceUpdate]]],
```
and then we pattern match directly on the key.

It should simplify also the definition of instances [here](https://github.com/Cryptonomic/Conseil/pull/779/files#diff-9a63ae02048772847f4ed11b44a80206R37)
WDYT?

PR #782, opened 2020-04-30
Refactor/split tezos into sub modules

ajuszczak:
This pull-request provides mechanism to run multiple Lorre process for different platforms, which is preparation step for extending current integrations.

Lorre requires new argument `platform` to be provided. Based on that specific configuration can be picked from `application.conf` file.

In addition this pull-request also contains steps to make API more generic and prepare it for more integrations as well.


conseil-api/src/main/scala/tech/cryptonomic/conseil/api/Conseil.scala
```diff
@@ -81,11 +82,11 @@ object Conseil
     * @return a metadata services object or a failed result
     */
   def initServices(
-      conseilOperations: ConseilOperations,
-      server: ServerConfiguration,
-      platforms: PlatformsConfiguration,
-      metadataOverrides: MetadataConfiguration,
-      nautilusCloud: Option[NautilusCloudConfiguration]
+                    conseilOperations: ConseilOperations,
```
ivanopagano, line 85: please run scalafmt

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/Conseil.scala
```diff
@@ -130,18 +135,18 @@ object Conseil
...
+                 platforms: PlatformsConfiguration,
+                 metadataOverrides: MetadataConfiguration,
+                 securityApi: SecurityApi,
+                 verbose: VerboseOutput
   )(implicit executionContext: ExecutionContext, system: ActorSystem, mat: ActorMaterializer) = {
     val tezosDispatcher = system.dispatchers.lookup("akka.tezos-dispatcher")
 
     lazy val platformDiscovery = PlatformDiscovery(metadataService)
-    lazy val data = Data(metadataService, server, metadataOverrides, conseilOperations)(tezosDispatcher)
+    lazy val data = Data(metadataService, metadataOverrides, conseilOperations, server.maxQueryResultSize)(tezosDispatcher)
```
ivanopagano, line 149: I think that you can rename the dispatcher even now, if you prefer. It's not fundamentally tied to the platform.
It was named so mostly because it was tuned to tezos latencies/throughput.
But we can decide to rename later on when/if we need to introduce different dispatchers.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/config/ConseilAppConfig.scala
```diff
@@ -0,0 +1,181 @@
...
+  protected def loadApplicationConfiguration(commandLineArgs: Array[String]): ConfigReader.Result[Configurations] = {
+    import ConseilAppConfig.Implicits._
+
+    implicit def hint[T]: ProductHint[T] = ProductHint[T](ConfigFieldMapping(CamelCase, CamelCase))
+
+    /** Use the pureconfig convention to handle configuration from the command line */
+    def readArgs(args: Array[String]): ConfigReader.Result[ArgumentsConfig] =
+      argsParser.parse(args, ArgumentsConfig()).toRight[ConfigReaderFailures](sys.exit(1))
+
+    //this extra configuration might be needed as we add send operations to the conseil API
```
ivanopagano, line 58: I would get rid of this, not sure if
a. the write api would pass through conseil-api
b. we'll need to rely on lib-sodium

@vishakh ?

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/config/ConseilAppConfig.scala
```diff
@@ -0,0 +1,181 @@
...
+import com.typesafe.config.{ConfigObject, ConfigValue}
+import com.typesafe.scalalogging.LazyLogging
+import pureconfig.error.{ConfigReaderFailures, ExceptionThrown, FailureReason}
+import pureconfig.generic.ProductHint
+import pureconfig.generic.auto._
+import pureconfig.{ConfigReader, _}
+import scopt.OptionParser
+import tech.cryptonomic.conseil.api.security.Security
+import tech.cryptonomic.conseil.common.config.Platforms._
+import tech.cryptonomic.conseil.api.security.Security._
```
ivanopagano, line 13: Can I ask you to kindly put imports sorted to a minimum, to make it easier to figure out?
Just putting similar things together, no need for anything too strict.
Thanks

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/generic/chain/DataTypes.scala
```diff
@@ -255,7 +259,9 @@ object DataTypes {
       aggregation: List[Aggregation] = List.empty,
       temporalPartition: Option[String] = None,
       snapshot: Option[Snapshot] = None
-  )
+  ) {
+    def adjustLimit(value: Int): Query = copy(limit = Math.min(limit, value))
```
ivanopagano, line 263: I'd suggest a more explicit name like
`capResultsLimit` `withLimitCap`, or similar

"adjust" is very vague

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
```diff
@@ -38,7 +38,22 @@ object Platforms extends LazyLogging {
     * Should associates each platform to a list of its internally defined configuration type, matching the
     * inner `platform.ConfigurationType`
     */
-  case class PlatformsConfiguration(platforms: Map[BlockchainPlatform, List[PlatformConfiguration]])
+  case class PlatformsConfiguration(platforms: Map[BlockchainPlatform, List[PlatformConfiguration]]) {
+
+    /*** Extracts platforms from configuration */
+    def getPlatforms: List[Platform] =
```
ivanopagano, line 44: This could be a `Set`

ivanopagano: but ignore this, if you're only moving a piece of code around

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/metadata/MetadataService.scala
```diff
@@ -17,22 +16,24 @@ class MetadataService(
...
+  private val platforms = transformation.overridePlatforms(config.getPlatforms)
 
   private val networks = platforms.map { platform =>
-    platform.path -> transformation.overrideNetworks(platform.path, ConfigUtil.getNetworks(config, platform.name))
+    platform.path -> transformation.overrideNetworks(platform.path, config.getNetworks(platform.name))
   }.toMap
 
   private val entities = {
-    val allEntities = Await.result(platformDiscoveryOperations.getEntities, 2 second)
+    val allEntities = platforms
```
ivanopagano, line 29: Would be `traverse` the platforms and then `Await` for the whole list more clear?

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/metadata/MetadataService.scala
```diff
@@ -44,7 +45,7 @@ class MetadataService(
     }.toSet
 
     val result = Future.traverse(networkPaths) { path =>
```
ivanopagano, line 47: please rename if you don't mind
`networkPaths` -> `entityPaths`

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/metadata/MetadataService.scala
```diff
@@ -17,22 +16,24 @@ class MetadataService(
...
-    val allEntities = Await.result(platformDiscoveryOperations.getEntities, 2 second)
+    val allEntities = platforms
+      .map(platform => platform.path -> Await.result(platformDiscoveryOperations(platform.name).getEntities, 2 second))
+      .toMap
 
     networks.values.flatten
       .map(_.path)
       .map(
-        networkPath => networkPath -> transformation.overrideEntities(networkPath, allEntities)
+        networkPath => networkPath -> transformation.overrideEntities(networkPath, allEntities(networkPath.up))
```
ivanopagano, line 36: What would you say of having local methods that extracts the platform from different paths, using explicit names?
```scala
private def platformPath(path: NetworkPath): PlatformPath = path.up
private def platformPath(path: EntityPath): PlatformPath = path.up.up
...
```

ivanopagano: at call site we would have, e.g.

```scala
if (exists(path)) {
  platformDiscoveryOperations(platformPath(path).platform).getEntities.map {...}
}
```

ivanopagano: or something like `upToPlatform(path: ...): PlatformPath`, which might be more clear?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/Lorre.scala
```diff
@@ -51,960 +25,45 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
...
-  private[this] def logProcessingProgress(entityName: String, totalToProcess: Int, processStartNanos: Long)(
-      processed: Int
-  ): Unit = {
-    val elapsed = System.nanoTime() - processStartNanos
-    val progress = processed.toDouble / totalToProcess
-    logger.info("================================== Progress Report ==================================")
-    logger.info("Completed processing {}% of total requested {}s", "%.2f".format(progress * 100), entityName)
+    //actual start of the indexer
+    indexer.start()
+  } finally shutdown()
```
ivanopagano, line 61: Why no shutdownHook?

What if the process is killed via OS signal, like a command line interrupt?
Will it shutdown cleanly?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/LorreLogging.scala
```diff
@@ -2,33 +2,56 @@ package tech.cryptonomic.conseil.indexer
...
 import tech.cryptonomic.conseil.common.io.MainOutputs.{showDatabaseConfiguration, showPlatformConfiguration}
+import tech.cryptonomic.conseil.indexer.config.LorreConfiguration
 
-/** Defines what to print when starting Lorre */
-trait LorreMainOutput {
+import scala.concurrent.duration.{Duration, NANOSECONDS}
 
-  /** we need to have a logger */
-  protected[this] def logger: Logger
+trait LorreLogging {
```
ivanopagano, line 11: Personally I'd choose a less generic term.

*Logging says nothing about what the trait is about, and anything else related to logging might reuse the term in the future and create confusion on each module's responsibility

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/Conseil.scala
ajuszczak: Right, I've just done it.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/Conseil.scala
ajuszczak: Hm... do you mean, to make it more generic and also rename the section from the configuration?

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/config/ConseilAppConfig.scala
ajuszczak: Sure, I guess `scalafmt` does not optimize them, cause I just did it and still issue is there. I did it manually here, and will try to find the places where it's unoptimized or find out more automatic approach.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/generic/chain/DataTypes.scala
ajuszczak: Renamed to `withLimitCap`. Does makes more sense that what I suggested

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
ajuszczak: Yea, I'm making some changes if they are needed. This change does not apply to much changes, so I can do that :)

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/Conseil.scala
ivanopagano: yes, if you think it makes sense, to start removing references to the "tezos" name when not strictly related.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/config/ConseilAppConfig.scala
ivanopagano: Ok, yeah, scalafmt doesn't help, and I'm sorry to say that I generally clean up things manually...
Let's say that it might help the reader, to easily find imports, but it's not fundamental in any way.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
ajuszczak: I was trying to do that and funny enough this change makes the endpoint broken, mostly because rest of the methods that uses it, expects List to be passed. While changing to Set I would need to change it everywhere. Just thinking that this change might be to big and I don't know how to solve missing implicit in endpoints for `Response[Set[Entity]]`. I could cut this off at some point and convert back to list at the "closest" possible level. What do you think?

ivanopagano: Refrain from doing too much refactor work, at this point is not that relevant.

Thanks for trying, don't worry about that

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/metadata/MetadataService.scala
ajuszczak: I'm thinking more about making this: 
```
private def toPlatformName(path: NetworkPath): String = path.up.platform
private def toPlatformName(path: EntityPath): String = toPlatformName(path.up)
```
mostly because there is only 1 (out of 6-7) usages that calls `.platform` directly after it, so I guess that would work better, to call `.platform` directly in this method.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/metadata/MetadataService.scala
ajuszczak:
```
val futureEntities = platforms.map { platform =>
   platformDiscoveryOperations(platform.name).getEntities.map(v => platform.name -> v)
}
val allEntities = Await.result(Future.sequence(futureEntities), 5 second).toMap
```

That's the idea I managed to came up with. Would that work better for you or maybe you have better one?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/Lorre.scala
ajuszczak: This is the same `try` we had: https://github.com/Cryptonomic/Conseil/blob/master/src/main/scala/tech/cryptonomic/conseil/Lorre.scala#L298 and this is what you are talking about: https://github.com/Cryptonomic/Conseil/pull/782/files/c8ab2b61d6f38a0789de939671fa13f03901fde4#diff-c7731ebc3651fb64cafd10746808698eR36 this is still there. I did not changed that :)

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/Conseil.scala
ajuszczak: I just put it under `http` section, since it's related more to API/HTTP than tezos itself.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/config/ConseilAppConfig.scala
ajuszczak: Yea, I'm mostly doing it too, but there are couple of places where imports are marked as 'unused' but they are actually used, so IntelliJ goes crazy, removes them and I'm back with the weird compilation issues. I'm marking these imports as used, just for me, but clearly forgot to do it in this class. Will pay attention in the future - this one is done.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/metadata/MetadataService.scala
ivanopagano: yeah, it's fine.

FYI `traverse` is exactly map+sequence, and it's most often what's actually needed

```scala
val futureEntities = Future.traverse(platforms) { platform =>
  platformDiscoveryOperations(platform.name)
    .getEntities.map(v => platform.name -> v)
}
val allEnties = Await.result(futureEntities, 5.seconds).toMap
```

ajuszczak: Awesome. I've changed that.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/metadata/MetadataService.scala
ivanopagano: Yes, fine for me.
Let me check if I got it: since the actual PlatformPath is seldom used without calling the platform name immediately after, you want to make that operation easier, and do things explicitly for the only case needing the actual path?

If that's what you meant, it's perfectly ok with me

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/Lorre.scala
ivanopagano: You're right, I missed it hiding somewhere in the diffs 👍

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/metadata/MetadataService.scala
ajuszczak: Yew, that is correct.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/sql/postgres/PostgresDataOperations.scala
```diff
@@ -0,0 +1,100 @@
...
+import slick.jdbc.PostgresProfile.api._
+import tech.cryptonomic.conseil.common.generic.chain.DataOperations
+import tech.cryptonomic.conseil.common.generic.chain.DataTypes.OutputType.OutputType
+import tech.cryptonomic.conseil.common.generic.chain.DataTypes.{Field, Predicate, Query, QueryResponse, _}
+import tech.cryptonomic.conseil.common.sql.Sanitizer._
+import tech.cryptonomic.conseil.common.util.DatabaseUtil.QueryBuilder.{makeQuery, makeTemporalQuery, _}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object PostgresDataOperations {
```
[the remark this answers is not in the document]
ajuszczak: Well... I'm not either, but I guess that importing Postgres profile all over the places is not the best solution either. What I would suggest is to simply create trait like that:
```
trait DatabaseAware[T <: JdbcProfile] {
  val profile: T
}
```
and it's implementation like that:
```
trait DefaultDatabaseAware extends DatabaseAware[PostgresProfile] {
  override val profile: PostgresProfile = PostgresProfile
}
```

then you can I think simply import `profile.api._` inside the class that requires that and that should do the trick. Once we will decide to change database we are only going to change the profile in `DefaultDatabaseAware` without jumping all over the places.

Do you have any suggestion how I can rename that? There is so many classes prefixed with `Database` and I believe that there are currently two of them having the same name.

conseil-lorre/src/main/resources/application.conf
```diff
@@ -44,6 +44,8 @@ lorre {
     maxConnections: 10
   }
 
+  //TODO Should we wrap configuration from below into 'Tezos' secion? The same in the *Configuration class?
```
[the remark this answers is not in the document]
ajuszczak: It should, but question is whether or not, rest of the blockchains we are planning to integrate will have for example delegates, accounts stuff like that. Meaning if we can actually map that to all of the blockchains? I don't think so, thus this `//TODO`.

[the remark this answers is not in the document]
ajuszczak: Same about `blockRightsFetching` section

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/LorreLogging.scala
ajuszczak: I've split that into two different traits (cause they are also used separately), so method `display*` are in `LorreInfoLogging` and the first one is in `LorreProgressLogging`. Would that work?

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/config/ConseilAppConfig.scala
```diff
@@ -49,8 +49,6 @@ trait ConseilAppConfig {
   protected def loadApplicationConfiguration(commandLineArgs: Array[String]): ConfigReader.Result[Configurations] = {
     import ConseilAppConfig.Implicits._
 
-    implicit def hint[T]: ProductHint[T] = ProductHint[T](ConfigFieldMapping(CamelCase, CamelCase))
```
ivanopagano: Are you sure we can remove it?

The compiler flagged this as unused in the past, but in fact it might break the actual runtime behaviour.
Do the tests run correctly even without this hint?

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/tezos/TezosTupleFlattenHelper.scala
```diff
@@ -1,10 +1,10 @@
-package tech.cryptonomic.conseil.api.routes.openapi
+package tech.cryptonomic.conseil.api.routes.platform.data.tezos
 
 /** Helper object which implements flatten for the tuples using shapeless library
   * Taken from:
   * https://gist.github.com/travisbrown/3945529
   * */
-object TupleFlattenHelper {
+private[tezos] object TezosTupleFlattenHelper {
```
ivanopagano, line 7: This shouldn't be tezos-specific, am I missing something?
I would expect this to be a common helper

conseil-api/src/main/resources/application.conf
```diff
@@ -34,20 +34,19 @@ conseil {
...
-    type: "Dispatcher"
-    executor: "thread-pool-executor"
-    throughput: 1
-
-    thread-pool-executor {
-      fixed-pool-size: 16
+  http {
+    # this is essentially available to enable composition of database operations
+    # current configuration is based upon the default-blocking-io-dispatcher in akka
+    dispatcher {
```
ivanopagano, line 40: will this conflict and override a default akka-http value that might be used elsewhere from the toolkit?

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/sql/postgres/PostgresDataOperations.scala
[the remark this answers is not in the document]
ivanopagano: I think your current naming might work.

About wrapping the dependency in a trait it needs be done sooner or later.

Not sure this is the right moment, maybe on the next project phase?

conseil-lorre/src/main/resources/application.conf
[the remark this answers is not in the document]
ivanopagano: Sorry, my bad, clearly they are tezos specific, at least until we know more about other chains

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/tezos/TezosTupleFlattenHelper.scala
ajuszczak: You are actually right.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/config/ConseilAppConfig.scala
ajuszczak: Yes, tests are passing. I'm quite positive about that, because only `tech.cryptonomic.conseil.api.config.ConseilAppConfig.Implicits#tezosConfigurationReader`requires that, according to tests and what I've noticed in configuration itself.

Yes, compiler flags that, but when I removed that from class mentioned above, tests are failing, here, nothing happens.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/tezos/TezosTupleFlattenHelper.scala
ajuszczak: I've moved that to `util` package in `common`

conseil-api/src/main/resources/application.conf
ajuszczak: In akka-http configuration, section like `dispatcher` does not exist, but I can rename that to feel a bit safer or move under different section. Which would you prefer? I guess renaming that simply to `api-dispatcher` or `data-api-dispatcher` would do the trick.

ivanopagano: no need for me, I was just double-checking with you

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/config/ConseilAppConfig.scala
ivanopagano: I just hope we're not breaking things at runtime, that we didn't manage to unit-test 😅

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
```diff
@@ -0,0 +1,136 @@
...
+    lazy val cachedDataEndpoints: MetadataService => Map[String, ApiDataEndpoints] = service =>
+      cache.map {
+        case (key, value) => key.name -> value.dataEndpoint(service)
+      }
+
+    private def forVisiblePlatforms(f: BlockchainPlatform => Api): Map[BlockchainPlatform, Api] =
+      transformation
+        .overridePlatforms(config.platforms.getPlatforms)
+        .map(toBlockchainPlatform)
+        .map(platform => platform -> f(platform))
```
ivanopagano, line 130: why two splitted `.map` calls and not a single

```scala
.map {
  p =>
    val platform = BlockchainPlatform.fromString(p.name)
    platform -> f(platform)
}
```

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
```diff
@@ -0,0 +1,136 @@
...
+                  }
+                }
+            }
+          }
+        }
+      }
+    )
+  }
+
+  private class ApiCache {
```
ivanopagano, line 112: Wouldn't an `object` serve the same purpose more clearly?

ivanopagano: Here we need more docs to be able to easily make out what is what, especially with String map keys.
Can't existing wrappers be used for keys? Especially Value classes (if there is any...)

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/Api.scala
```diff
@@ -0,0 +1,61 @@
...
+
+import scala.concurrent.ExecutionContext
+import scala.util.{Failure, Success}
+
+trait Api {
+  def dataEndpoint(metadataService: MetadataService): ApiDataEndpoints
+  def discoveryOperations: PlatformDiscoveryOperations
+}
+
+class TezosApi(metadataOverrides: MetadataConfiguration, server: ConseilConfiguration)(
```
ivanopagano, line 23: Why isn't this splitted to a separate file, like the data endpoints?

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/tezos/TezosDataEndpoint.scala
```diff
@@ -16,22 +17,23 @@ import scala.concurrent.{ExecutionContext, Future}
   * Platform discovery routes.
   * @param apiExecutionContext   is used to call the async operations exposed by the api service
   */
-case class TezosData(
+case class TezosDataEndpoint(
```
ivanopagano, line 20: trivial change: I'd expect this to be plural: TezosDataEndpoints

ivanopagano: Changed my mind: we need a different name for this, since TezosDataEnpoints is already the trait describing the abstract endpoints algebra.
This would generate a lot of confusion.

`TezosDataRoutes` might work?

This would imply also renaming the parent interface to `ApiDataRoutes`...

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/tezos/TezosDataEndpoints.scala
```diff
@@ -1,16 +1,18 @@
...
-trait DataEndpoints
+trait TezosDataEndpoints
     extends algebra.JsonSchemaEntities
-    with DataJsonSchemas
-    with ApiFilterFromQueryString
+    with TezosDataJsonSchemas
+    with TezosApiFilterFromQueryString
     with Validation {
 
   /** Common path among endpoints */
```
ivanopagano, line 18: I think we have an issue here.

Differently from the metadata endpoints paths, where "platform" is a path segment, here we should have it fixed to the actual `"tezos"` value, because this routes will be combined with all those from any platform found in config, as I seem to understand from ConseilApi.scala L#90+

piotrkosecki: at the moment(meaning before the big refactor) each query worked for each network found in config(TBH I've never tested with different platforms, but I suspect it would work/require small change). 
How I see it is that we should be able to use thee same routes(as the API is mostly generic) to use with all the platforms/networks and be overridden only with the metadata configuration.
There is also another thing here - those endpoints always use `platform` so renaming it to start with `Tezos` is misguiding.

piotrkosecki:
In general I think that refactoring Conseil API is not really what it was supposed to be. 
I think, the idea behind the metadata files was for them to introduce platform/network specific descriptions.
If you look at the code(beside tests and compatibility endpoints, which were migrated from the old API to use new functions - @anonymoussprocket are compatibility endpoints still used?) Conseil does not have any platform/network specific implementations - therefore renaming Conseil related classes with `Tezos` prefix should not be necessary.
@ajuszczak can you see the way of adjusting query/metadata implementations to use metadata for platforms/networks instead of dedicated implementations?

@vishakh @anonymoussprocket Did I get the idea correctly?

vishakh:
> In general I think that refactoring Conseil API is not really what it was supposed to be.
> I think, the idea behind the metadata files was for them to introduce platform/network specific descriptions.
> If you look at the code(beside tests and compatibility endpoints, which were migrated from the old API to use new functions - @anonymoussprocket are compatibility endpoints still used?) Conseil does not have any platform/network specific implementations - therefore renaming Conseil related classes with `Tezos` prefix should not be necessary.
> @ajuszczak can you see the way of adjusting query/metadata implementations to use metadata for platforms/networks instead of dedicated implementations?
> 
> @vishakh @anonymoussprocket Did I get the idea correctly?

@piotrkosecki Yes, you are right.

I think we do use the compatibility endpoints though. We can decommission them if it's too much of a pain to maintain.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/tezos/TezosDataEndpoints.scala
ivanopagano: > There is also another thing here - those endpoints always use `platform` so renaming it to start with `Tezos` is misguiding.

Not sure I'm following you there, when asking for metadata I'm ok that the path would end on `platform` but then the user will swap that with `/tezos/networks` when digging deeper...

Same thing I expect for the data endpoints?

Like `/v2/data/tezos/mainnet/operations` ... 

The need here should be to swap the actual implementation based on the platform, because other blockchains might find data differently, AFAICT, so we build "custom routes" for each platform and mash them together under the same path ("/v2/data/") to get the specific implementations.
But each route implementation should provide a different segment for the platform. This is what I'm trying to tell.

Is there any error in what I'm saying?

ivanopagano:
> Conseil does not have any platform/network specific implementations - therefore renaming Conseil related classes with `Tezos` prefix should not be necessary.

If I understood correctly, and @ajuszczak can confirm, the issue was just that the custom caching and other parts of the code actually mixed tezos specific parts so much that re-using the code as-is was not possible anymore. Hence a different solution was needed, to cleanly separate tezos-specific code, that would not block this whole process for more than needed.

I think we can, and will need to, do further cleanup when time allows or necessity requires. Yet right now we need to start working on integration, which will probably pull more details to the surface, which might be useful to actual understand the direction we'll want to take.

YMMV

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/tezos/TezosDataEndpoints.scala
piotrkosecki: > Is there any error in what I'm saying?

no, thats more or less that I meant - create different segments for the platforms if we are going to get through with this redesign, and we should remove platform parameters from those endpoints.

ajuszczak: Keep in mind that `queryEndpoint` and everything what comes with that is "generic" so `platform` can be actually extracted and passed there, but only for this resource. Rest of the resources should be platform-specific, meaning `platform` segment should be removed from `path` and put as constant. 

I decided not to do that here, mostly because this changes the behavior of API. Currently, when platform is not supported, the lower layer will respond with such information and endpoint will convert this to 404. Once I'll hardcode "tezos" in a path name, the request with unsupported platform will be rejected (different error code or even exception). This requires dedicated error handling mechanism. I decided to limit the scope and leave it as it is. I can add comment, or even take care of such mechanism if you want to, but I don't know what changes it requires to be done. Need to investigate.

ajuszczak:
I agree with @ivanopagano that this was mostly to "unlock" new integrations. 
I'm not so sure tho, how easy it might be (or if that is even possible) to have entire API blockchain agnostic, mostly because, when you are asking for blocks from Tezos (at least at code level), we expect specific type to be returned. Of course we can always return `Map[String, Any]` but is that what we really want to do? I'm not sure.

My long-run idea was, as Ivano mentioned to have some parts to be generic (like this `post` route) and some to be platform and then network specific (but that level is not needed I guess). That being said, we cannot generalize entire API. Not, when we are asking for `data`. Different case when we are asking for `Metadata` and I believe, what we can do there is to use similar mechanism that we have for `selectWithPredicate`. Methods that we are using directly from `TezosDatabaseOperations` (which blocks us from making that generic) are done, because we need to join two or more tables. if we could add generic method for doing that, that would help.

Keep in mind that, since different blockchains, will return different data, Query/Filter needs to be implemented per blockchain as well.

About the caching. I'm not sure how it blocks us from making code generic. I'm pretty sure that caching mechanism makes entire project much complicated that it would be with optimizing database itself.

To sum it up, there is more work to be done to make it generic that you probably imagine. Since this phase was mostly about splitting modules to increase visibility of classes "with problems", I'm not going to do that here. We may think about that for the future. Treat this changes more like initial step, there is more work to be done, which I hope I'll be able to do in upcoming days, to give entire phase a better shape that it is for now and identify potential classes/mechanisms, that we could re-implement to make entire code clean, faster, better.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/tezos/TezosDataEndpoint.scala
ajuszczak: That make sense. There are too many, very similar names for classes.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/Api.scala
ajuszczak: No particular reason, besides the fact that endpoints are in specific packages and I couldn't find better place for it. I extracted that to `TezosApi` file.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
ajuszczak: So, here we are using `BlockchainPlatform` to fill up the map. Mostly because it's sealed trait, which provides all possible cases (and warning is displayed when any of it is not implemented in `forVisiblePlatforms` method, but... the `MetadataService` actually uses `Platform` case class, which does not have such option. 

That being said, I can use the same type here (which is `Platform`) but loose this warning and compile-time, or use `BlockchainPlatform` but when, `MetadataService` would need to convert `Platform` -> `BlockchainPlatfrom` every time.

Probably both works, question is which one is better option?

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
ajuszczak: Visibility mostly, I believe that this double `.map` would be optimized anyway by compiler (?)

ivanopagano: In this case I think so, but it's not always guaranteed. Especially with future composition...

I personally tend to avoid multiple map / flatmap calls that could be manually fused, but it's not a general rule, so I'd say it's ok.

Thanks for explaining

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
ivanopagano: I'm not sure I see what you mean here, was I unclear, possibly? I'll restate my observations:

1. why `private class ApiCache` vs. `private object ApiCache`?
2. in `cachedDiscoveryOperations` and `cachedDataEndpoints` the maps have both Strings as keys. Is there any more explicitly typed wrapper to use, to understand what should be expected there? The latter is because I needed to check both here and where those maps are used, to understand the logic, and it wasn't immediately obvious to me.

Maybe the original code was lacking in terms of comments, it's personally something I try to recommend always, to explain any piece of code which might not be obvious, YMMV.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/tezos/TezosDataEndpoints.scala
ivanopagano: Ok for me, I didn't get that you planned to enable different platforms in a later step.

This was just preliminary preparation, right?

ajuszczak: Yes

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
ajuszczak: Point 1. is clear to me. Point 2 is a bit more complicated. I just explained above why `Map[String, ...]` not `Map[anything-else, ...]`. Mostly because it is hard to find the best option to pick, because we have two different types describing kind of the same (or even 3 very similar): `BlockchainPlatform`, `Platform` and `PlatformPath`.

I just said, that the best idea would be to change `Map[String, ...]` into `Map[BlockchainPlatform, ...]`, but then, since we are accessing these data from `MetadataService` and this service has `PlatformPath`, I would need to convert `PlatformPath` -> `BlockchainPlatform` every time someone would like to access these services. The common ground for `BlockchainPlatform` and `PlatformPath` is Platform's name. That's why `String` :)

I was just asking, taking into what I just wrote, if you see, which type from what we have, would work better, because for me, `String` is such type, at least for the time being.

ivanopagano: I trust your decision on this, I see that it's more complex than it appeared at first sight

ajuszczak:
Due to issues with commits history made during conflicts resolving. I'm closing this PR in favor of #799 New pull-request contains exactly the same changes and history is clean. Sorry for confusion.

PR #798, opened 2020-05-11
Adding dependency-check plugin

piotrkosecki:
From what I've found this plugin checks only dependencies and does not perform static code analysis. I think most optimal way to use it in our case is with `sbt dependencyCheckAggregate`.
It shows known vulnerabilities in the dependencies in our new modular setup. Generated report can be found in project path `target/scala-2.12/dependency-check-report.html`

For static code analysis there are some tools mentioned here:
https://owasp.org/www-community/Free_for_Open_Source_Application_Security_Tools
but they require separate setup and some of them, integration with CI.


ivanopagano:
@piotrkosecki Do you have a resulting report from running on the current conseil repo?
Is it possible to have a link to the report source, or a snapshot, just to get an idea?

Thank you

piotrkosecki:
> @piotrkosecki Do you have a resulting report from running on the current conseil repo?
> Is it possible to have a link to the report source, or a snapshot, just to get an idea?
> 
> Thank you

So basically the report is being generated as html file(but it can be generated in other formats, but none of them look good).
Github does not support attaching html files so I put here some screen shots. 
Here you can see list of vulnerabilities:
Screenshot 2020-05-12 at 09 59 53
Then each vulnerability is explained and gives links to for reference.
Screenshot 2020-05-12 at 10 02 16

PR #799, opened 2020-05-12
Refactor/add support for integrating multiple chain versions

ajuszczak:
This pull-request is continuation of https://github.com/Cryptonomic/Conseil/pull/782 but since I made a mistake during rebase and history is rewritten there I decided to create clean copy without this issue.

Below you can find changes, which:
- makes Lorre module easily extendable. Blockchain Integration is controlled from the argument parameter.
- makes API module easily extendable and exposing endpoints is controlled via configuration file. Is uses metadata configuration to see which endpoint (blockchain) should be exposed and which one not.


conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilMainOutput.scala
```diff
@@ -50,4 +50,14 @@ trait ConseilMainOutput {
       showDatabaseConfiguration("conseil")
     )
 
+  /* prepare output to display existing platforms and networks */
```
ivanopagano, line 53: 👍

project/Dependencies.scala
```diff
@@ -2,116 +2,189 @@ import sbt._
...
+
+  private val bitcoin = Seq("fr.acinq" %% "bitcoin-lib" % "0.9.18-SNAPSHOT")
+
+  private val scorex = Seq(
+    "org.scorexfoundation" %% "scrypto"     % "2.1.7",
+    "org.scorexfoundation" %% "scorex-util" % "0.1.6"
   )
 
+  //List of dependencies that are not really used (?) - to be removed
+  ////      "com.chuusai"             %% "shapeless"   % "2.3.3",
```
ivanopagano, line 131: I'm not totally sure if it's not at least marginally (like 1 line of code) used anywhere.

The point here is that shapeless is used by many of our dependencies anyway, and I prefer to have a direct specific dependency if we use it, rather than by transitivity, if you get my meaning

project/Dependencies.scala
```diff
@@ -2,116 +2,189 @@ import sbt._
...
+
+  private val scopt = Seq("com.github.scopt" %% "scopt" % "4.0.0-RC2")
+
+  private val logging = Seq(
+    "ch.qos.logback"             % "logback-classic"          % "1.2.3",
+    "net.logstash.logback"       % "logstash-logback-encoder" % "5.3",
+    "com.typesafe.scala-logging" %% "scala-logging"           % "3.7.2"
+  )
+
   private val akkaVersion = "2.5.21"
```
ivanopagano, line 17: I personally preferred all external versions to be collected together, instead of spread near to their actual dependency list.
That's because usually you want to easily scan the currently used versions of those libraries, and now it's not so easy.

Some use the convention to put the versions inside a local `Version` object, to keep 'em bundled, and use them as `Version.akka` for example.

ajuszczak: Yea, that will serve the purpose better I guess. Just moved versions to `Versions` object and lets see.

project/Dependencies.scala
ajuszczak: I checked if packages from this class are imported anywhere and I believe they are not, thus I decided to comment them out and eventually remove.

PR #803, opened 2020-05-14
chores - put assembled fat-jar into a local project sub-folder

ivanopagano:
This PR aims to make it easier to find the assembled jars, related to the wiki instructions on how to run the app from those.

* the jars will be found inside a local project sub-folder: namely `dist`
* the jar names will have a version in the naming [1]

[1] @vishakh @anonymoussprocket There was a particular reason not to produce versioned jar-names? If that's the case I'll revert to the previous behaviour.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ COPY --chown=builduser:builduser ./conseil-lorre/src /src/conseil-lorre/src
 COPY --chown=builduser:builduser ./build.sbt /src
 COPY --chown=builduser:builduser ./publishing.sbt /src
 WORKDIR /src
-RUN sbt clean assembly -J-Xss32m
+RUN sbt 'set api / assembly / assemblyOutputPath := file(s"/tmp/conseil-api.jar")' 'set lorre / assembly / assemblyOutputPath := file(s"/tmp/conseil-lorre.jar")' clean assembly -J-Xss32m
 
 FROM openjdk:13-alpine
 RUN apk --no-cache add ca-certificates
--- a/project/Assembly.scala
+++ b/project/Assembly.scala
@@ -12,12 +12,14 @@ object Assembly {
       project
         .settings(
           mainClass in assembly := mainClass.value,
-          assemblyOutputPath in assembly := file(s"/tmp/${name.value}.jar"),
+          assemblyOutputPath in assembly := file(
+                s"${(baseDirectory in ThisBuild).value}/dist/${name.value}-${version.value}.jar"
+              ),
           test in assembly := {},
           assemblyMergeStrategy in assembly := {
-            case "application.conf" => MergeStrategy.concat
-            case x => (assemblyMergeStrategy in assembly).value(x)
-          }
+              case "application.conf" => MergeStrategy.concat
+              case x => (assemblyMergeStrategy in assembly).value(x)
+            }
         )
         .enablePlugins(AssemblyPlugin)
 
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -20,7 +20,7 @@ COPY --chown=builduser:builduser ./conseil-lorre/src /src/conseil-lorre/src
 COPY --chown=builduser:builduser ./build.sbt /src
 COPY --chown=builduser:builduser ./publishing.sbt /src
 WORKDIR /src
-RUN sbt clean assembly -J-Xss32m
+RUN sbt 'set api / assembly / assemblyOutputPath := file(s"/tmp/conseil-api.jar")' 'set lorre / assembly / assemblyOutputPath := file(s"/tmp/conseil-lorre.jar")' clean assembly -J-Xss32m
```
ajuszczak, line 23: Since this setup depends heavily on `build.sbt` (each change of module name will brake this command for example) and basically `build.sbt` is a "place to go" when it comes to building the project. I would rather suggest to create command alias in the `build.sbt` rather than putting setting properties in here. 

Maybe we could have command: 
- `assembly` which would work as previously, to the `/tmp/` and
- `assemblyDist` which would put the jars into specific `/dist/` folders?

I mean, I'm not sure if it's safe for the community to change the original location, but I'll leave that up to you.

project/Assembly.scala
```diff
@@ -12,12 +12,14 @@ object Assembly {
       project
         .settings(
           mainClass in assembly := mainClass.value,
-          assemblyOutputPath in assembly := file(s"/tmp/${name.value}.jar"),
+          assemblyOutputPath in assembly := file(
+                s"${(baseDirectory in ThisBuild).value}/dist/${name.value}-${version.value}.jar"
```
ajuszczak, line 16: That is funny, because I would expect the same behavior for below paths:
- `s"${(baseDirectory in ThisBuild).value}/dist/${name.value}-${version.value}.jar"`
- `s"./dist/${name.value}-${version.value}.jar"`

but the first one creates files with version: `...-0.2020.0022` while in the same time the latter creates: `...-0.2020.0022-SNAPSHOT`.

Which one is correct, because I've started thinking that we should build it with `-SNAPSHOT`.

ivanopagano: Interesting, but it might depend on where the .git structure is? The snapshot suffix depends on changes seen by git wrt the clean commit status

Dockerfile
ivanopagano: I need your opinions on this, I made the change because one of our devs didn't found the assembled jars where expected, somehow....
I thought that putting things in a global tmp dir without any output clearly declaring so could be confusing. My take on that was to move the generated artifacts to a locally distinguished directory, as a default for people or ci doing that locally.

Though, even if the idea of separate tasks is fine, I'm not sure if it's worth it to use /tmp/ as the default choice. In that case I'd simply ignore the PR and leave things as is.

piotrkosecki: I think most misleading part was that `conseil-xyz.jar` was generated in the `target` directory inside of the project. I don't think the output directory of the two other jars is an issue.

ivanopagano: if that's the case, I can close this issue, with the mental reminder to document where the jars are created in the wiki

vishakh:
Yes, please update the wiki so Pratham can validate all the documentation
after the refactor.

ivanopagano:
Closing and leaving the situation as-is, with updated Wiki entry providing the missing information to locate the output.

PR #804, opened 2020-05-14
791 - make the token handling compatible with tzBTC format

### LINKED ISSUE #791: Support tzBTC token format
The big maps diffs for such tokens won't be correctly handled with the current token handling design.

We should support such token too

ivanopagano:
Closes #791 

This changes handles in a uniform way multiple token formats to read both addresses and balances from the big-map diffs for three common formats: USDtz, StakerDAO, tzBTC

It includes additional efforts to document the logic used to decode the specific formats of data included in those contracts

### details of the changes

The first release of the token-handling code would make simple assumptions on the format of balance results found in the big-map diffs.

From additional information collected on different FA1.2 token actual implementations, we now handle more formats to express the same balance changes in diffs, given that they can be

 - an explicit numeric value that defines the new token balance for the address identified by the map key (e.g. StakerDao)
 - a micheline instruction encoding a numeric int value of the token balance for the address identified by the map key (e.g. USDtz)
 - a byte-encoded micheline instruction that, once decoded, defines a pair holding a ZArith-encoded token balance for the address identified by the map key (e.g. tzBTC)

In addition to that, the key used to identify the balance owner in the big map can have different encodings, depending on the contract implementation, and now we handle

 - binary packed addresses (StakerDao and USDtz)
 - micheline binary encoded to an instruction that provides a "ledger" value as a packed address (tzBTC)
 - total supply reference for the tzBTC tokens, which is ignored by the indexer

All in all the information is essentially the same but can be stored in different formats.


vishakh:
Does this need to be applied to the release branch as well?

Could you add some details about the nature of the changes? What was changed and why? It will help me review the PR.

ivanopagano:
> Does this need to be applied to the release branch as well?
> 
> Could you add some details about the nature of the changes? What was changed and why? It will help me review the PR.

I've added additional information on the overall changes in the PR description.
Let me know if it's clear enough and if I might add a "design entry" for this in the wiki, too.

ivanopagano:
> > Does this need to be applied to the release branch as well?
> > Could you add some details about the nature of the changes? What was changed and why? It will help me review the PR.
> 
> @vishakh
> I've added additional information on the overall changes in the PR description.
> Let me know if it's clear enough and if I might add a "design entry" for this in the wiki, too.

I will prepare a backport for release today

ivanopagano:
@vishakh @anonymoussprocket 
I can't find a note with the addresses to fill the token-registries (mainnet, carthagenet) for tzBTC.

Could you kindly tell me how to fill the entry to put in this very PR to handle it?

PR #805, opened 2020-05-15
783 - capture additional operation cost details

### LINKED ISSUE #783: Capture additional operation cost details
Here's a [sample operation](https://tezos-dev.cryptonomic-infra.tech/chains/main/blocks/383401/operations/3/2) and [before](https://tezos-dev.cryptonomic-infra.tech/chains/main/blocks/383400/context/contracts/tz1LS1QM6EcRpNtNXxWPAWWscDyuQAmKv8Nj/balance) and [after](https://tezos-dev.cryptonomic-infra.tech/chains/main/blocks/383401/context/contracts/tz1LS1QM6EcRpNtNXxWPAWWscDyuQAmKv8Nj/balance) blocks. Note that we're not recording the burn (25700), and storage fee (4650000). We can back out the storage fee because we capture `paid_storage_delta`

piotrkosecki:
resolves #783

### DIFF
--- a/conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/DatabaseConversions.scala
+++ b/conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/DatabaseConversions.scala
@@ -418,6 +418,13 @@ object DatabaseConversions extends LazyLogging {
         )
         ) =>
       val (year, month, day, time) = extractDateTime(toSql(block.data.header.timestamp))
+      val paidStorageDiff = metadata.operation_result.paid_storage_size_diff.flatMap(extractBigDecimal)
+      val burn = paidStorageDiff.flatMap { psd =>
+        metadata.operation_result.balance_updates.toList.flatten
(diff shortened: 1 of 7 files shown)


conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/DatabaseConversions.scala
```diff
@@ -418,6 +418,13 @@ object DatabaseConversions extends LazyLogging {
         )
         ) =>
       val (year, month, day, time) = extractDateTime(toSql(block.data.header.timestamp))
+      val paidStorageDiff = metadata.operation_result.paid_storage_size_diff.flatMap(extractBigDecimal)
+      val burn = paidStorageDiff.flatMap { psd =>
```
ivanopagano, line 422: I think we might need to document what's going on in this block.
I have problems figuring out what is actually computed for burn

piotrkosecki:
It needs to be redone when I'll get more information.

PR #808, opened 2020-05-19
784 - Registered tokens table update

### LINKED ISSUE #784: Registered tokens table update
Token contracts should store scale. That is the power of 10 that should be used to interpret the values in the balance map.

If account balance is 1 and scale is 0, then token balance is 1. If account balance is 1 and scale is 6, the token balance is 0.000001.

piotrkosecki:
resolves #784


conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/Tables.scala
```diff
@@ -2493,25 +2493,26 @@ trait Tables {
     *  @param id Database column id SqlType(int4), PrimaryKey
     *  @param name Database column name SqlType(text)
     *  @param contractType Database column contract_type SqlType(text)
-    *  @param accountId Database column account_id SqlType(text) */
-  case class RegisteredTokensRow(id: Int, name: String, contractType: String, accountId: String)
+    *  @param accountId Database column account_id SqlType(text)
```
vishakh, line 2496: How come Tables.scala changed when conseil.sql did not?

vishakh: Ah, never mind. Need to get used to the new package structure!

PR #822, opened 2020-05-27
791 - add tzBTC to token registry [release backport]

ivanopagano:
relates to #791 

Register the tzBTC contract on mainnet in the token registry.


ivanopagano:
Configuration change Included in #820

PR #824, opened 2020-05-28
763 - Conseil GraphQL API

### LINKED ISSUE #763: Conseil should also expose a GraphQL API
As discussed previously, Conseil should also expose its data through GraphQL as many front end devs prefer GraphQL over REST. Let us identify a good library and add provisional support for our own investigations.

piotrkosecki:
possibly resolves #763 

Reviewing multiple GraphQL addition possibilities I came to propose using Hasura GraphQL engine.

### Why not Sangria or Caliban?
Both of those solutions require hardcoding elements of the DB we want to share with the user - in my opinion it would conflict with the idea of Conseil, which was to be independent of the DB structure and only use metadata configuration files. Also using any of those libraries would be harder to maintain.

### How this would work
I've added Hasura container in the `docker-compose` file. Configuration required me to add different user with limited capabilities to the Postgres so he won't be able to modify the DB. I did it by creating separate `.sql` file for Postgres container to initialise with.
For the configuration I've prepared basic metadata file and its being mounted in the `docker-compose` file.

### Running and using Hasura
Running Hasura is simple, just run `docker-compose up graphql-engine` and it will start, reqding metadata which I've prepared.

After it starts the UI will be available on `http://localhost:8080/console`. There you can see DB configuration and run queries.

Exposed GraphQL endpoint will be available here `http://localhost:8080/v1/graphql`.

### Deployment
I think this would be most tricky part because we would need to secure GraphQL endpoint with the metering API. This will surely need some DevOps attention.

To sum up, I think this is the best solution for adding GraphQL to Conseil. It has it's drawbacks which will require certain workarounds, but still does not require hardcoding table representations, and it will be easier to maintain. 

I'm leaving it for discussion.


piotrkosecki:
> I've easily set up using @piotrkosecki compose file.
> I had to manually select which tables and FKs to import from the GUI, before navigating any data.
> 
> Seems promising to me, we need to verify how to configure it in more depth

Did my basic metadata configuration work? I didn't add there any foreign keys, just tables except `_checkpoint` ones.

ivanopagano:
> > I've easily set up using @piotrkosecki compose file.
> > I had to manually select which tables and FKs to import from the GUI, before navigating any data.
> > Seems promising to me, we need to verify how to configure it in more depth
> 
> Did my basic metadata configuration work? I didn't add there any foreign keys, just tables except `_checkpoint` ones.

I started the container with no additional steps, when I opened the UI I couldn't find no "imported" table. I did go to the appropriate tab to enable which tables should be visible by hand.

Should I expect most tables to be visible out-of-the-box? If so it wasn't working for me.

piotrkosecki:
> > > I've easily set up using @piotrkosecki compose file.
> > > I had to manually select which tables and FKs to import from the GUI, before navigating any data.
> > > Seems promising to me, we need to verify how to configure it in more depth
> > 
> > 
> > Did my basic metadata configuration work? I didn't add there any foreign keys, just tables except `_checkpoint` ones.
> 
> I started the container with no additional steps, when I opened the UI I couldn't find no "imported" table. I did go to the appropriate tab to enable which tables should be visible by hand.
> 
> Should I expect most tables to be visible out-of-the-box? If so it wasn't working for me.

Ok, I must have made a mistake before and did not remove the container when I was testing it. 

Fortunately I've found a fix for this. It wasn't straightforward. The problem was that metadata exported by the UI(in json format) is not the same metadata with which I can initialise container.
I had to prepare additional yaml files for that, and believe me every one of them is necessary...

@ivanopagano can you try again running it with my latest changes?

PR #827, opened 2020-05-29
bugfix - trying to fix AppInfoRouteTest

piotrkosecki:
Added git tag before and removed it after running travis build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,8 @@ jdk:
   - openjdk8
 scala:
   - 2.12.10
-script: sbt clean coverage test coverageAggregate
+script:
+  - git tag 0.1234.5678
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -3,5 +3,8 @@ jdk:
   - openjdk8
 scala:
   - 2.12.10
-script: sbt clean coverage test coverageAggregate
+script:
+  - git tag 0.1234.5678
+  - sbt clean coverage test coverageAggregate
```
vishakh, line 8: What happens if this step fails for reason and the tag is never removed?

piotrkosecki: tested and it works, I think because it runs build on new VM each time and does not push the tag

PR #830, opened 2020-06-02
828 - wrong operation group hash fix[backport]

piotrkosecki:
fixing the bug for the release branch

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/TezosOptics.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/TezosOptics.scala
@@ -105,7 +105,7 @@ object TezosOptics {
       */
     def extractOperationsSequence(
         block: Block
-    ): List[Either[Operation, InternalOperationResults.InternalOperationResult]] =
+    ): List[(Either[Operation, InternalOperationResults.InternalOperationResult], OperationHash)] =
       for {
         group <- block.operationGroups
         op <- group.contents
@@ -120,7 +120,7 @@ object TezosOptics {
             Left(op) :: d.metadata.internal_operation_results.toList.flatten.map(Right(_))
           case _ => List(Left(op))
         }
-      } yield all
+      } yield all -> group.hash
 
     //Note, cycle 0 starts at the level 2 block
     def extractCycle(block: Block): Option[Int] =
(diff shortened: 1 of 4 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/TezosOptics.scala
```diff
@@ -105,7 +105,7 @@ object TezosOptics {
       */
     def extractOperationsSequence(
         block: Block
-    ): List[Either[Operation, InternalOperationResults.InternalOperationResult]] =
+    ): List[(Either[Operation, InternalOperationResults.InternalOperationResult], OperationHash)] =
```
ivanopagano, line 108: Same consideration as the parallel PR.

Maybe using the opHash as first element in the tuple and possibly using a wrapping case class could improve the reading experience

src/main/scala/tech/cryptonomic/conseil/tezos/app/TezosNamesOperations.scala
```diff
@@ -81,9 +81,9 @@ class TezosNamesOperations(tnsContracts: TNSContract, node: TezosNodeOperator) e
     blocks =>
       blocks.map { b =>
         val refs = extractAppliedTransactions(b).filter {
-          case Left(op) => tnsContracts.isKnownRegistrar(op.destination)
-          case Right(op) => tnsContracts.isKnownRegistrar(op.destination)
-        }.map(tnsContracts.readLookupMapReference).flattenOption
+          case (Left(op), _) => tnsContracts.isKnownRegistrar(op.destination)
```
ivanopagano, line 84: Here and everywhere else the group hash is not used from code downstream, I would suggest to simply remove it from the list as a first step, to simplify all the code coming after

src/main/scala/tech/cryptonomic/conseil/tezos/TezosOptics.scala
piotrkosecki: sure, changed order in the tuple

src/main/scala/tech/cryptonomic/conseil/tezos/app/TezosNamesOperations.scala
piotrkosecki: changed

PR #831, opened 2020-06-02
bugfix - bakers table with empty cycle and period[backport]

piotrkosecki:
fixed the issue where in bakers table could contain empty cycle/period for non genesis block

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
@@ -704,9 +704,13 @@ object DatabaseConversions extends LazyLogging {
   }
 
   implicit val blockAccountsAssociationToCheckpointRow =
-    new Conversion[List, (BlockHash, Int, Option[Instant], Option[Int], List[AccountId]), Tables.AccountsCheckpointRow] {
-      override def convert(from: (BlockHash, Int, Option[Instant], Option[Int], List[AccountId])) = {
-        val (blockHash, blockLevel, timestamp, cycle, ids) = from
+    new Conversion[
+      List,
+      (BlockHash, Int, Option[Instant], Option[Int], Option[Int], List[AccountId]),
+      Tables.AccountsCheckpointRow
+    ] {
+      override def convert(from: (BlockHash, Int, Option[Instant], Option[Int], Option[Int], List[AccountId])) = {
+        val (blockHash, blockLevel, timestamp, cycle, _, ids) = from
         ids.map(
           accountId =>
             Tables.AccountsCheckpointRow(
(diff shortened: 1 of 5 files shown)


src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
```diff
@@ -704,9 +704,13 @@ object DatabaseConversions extends LazyLogging {
...
   implicit val blockAccountsAssociationToCheckpointRow =
-    new Conversion[List, (BlockHash, Int, Option[Instant], Option[Int], List[AccountId]), Tables.AccountsCheckpointRow] {
-      override def convert(from: (BlockHash, Int, Option[Instant], Option[Int], List[AccountId])) = {
-        val (blockHash, blockLevel, timestamp, cycle, ids) = from
+    new Conversion[
+      List,
+      (BlockHash, Int, Option[Instant], Option[Int], Option[Int], List[AccountId]),
+      Tables.AccountsCheckpointRow
+    ] {
+      override def convert(from: (BlockHash, Int, Option[Instant], Option[Int], Option[Int], List[AccountId])) = {
```
ivanopagano, line 712: I won't block the PR for this, but I honestly start to find it hard to make sense of so many opaque parameters, with no structure and totally common types.
I personally feel it gets hard to follow and prone to easily make mistake.

I leave it to you if there's anything to improve or not.

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
```diff
@@ -350,7 +350,7 @@ object TezosDatabaseOperations extends LazyLogging {
         ids =>
           writeAccountsCheckpoint(
             List(
-              (hash, level, Some(timestamp), cycle, ids.map(AccountId(_)).toList)
+              (hash, level, Some(timestamp), cycle, None, ids.map(AccountId(_)).toList)
```
ivanopagano, line 353: aren't we losing info here?

src/main/scala/tech/cryptonomic/conseil/tezos/DatabaseConversions.scala
piotrkosecki: I think next time when I'll be out of issues I'll improving this, I don't think it's necessary for the bugfix PR

src/main/scala/tech/cryptonomic/conseil/tezos/TezosDatabaseOperations.scala
ivanopagano: ok,but I can't clearly make out, why we need to pass None if that parameter is not needed anyway? Can't it be removed completely from the expected values at this point in the code?

piotrkosecki: but in one place we use `.asTuple` to extract this data for passing to the `writeAccountsCheckpoint`, and here we're not.

ivanopagano: ok, now my eyes are opened... I can see clearly :eye:

PR #834, opened 2020-06-04
bugfix - fixed governance table to correctly store the proposals[backport]

piotrkosecki:
Fixes the issue that governance table did not store proposals - issue was that proposals had the same block_hash, and proposal_hash and it crashed the Lorre when it tried to insert it into the database(it was composite key). Fixed that with adding kind to the key.

### DIFF
--- a/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
+++ b/src/main/scala/tech/cryptonomic/conseil/Lorre.scala
@@ -465,7 +465,7 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
     /** Groups data needed for generating GovernanceRow
      *  Warning! It works on assumption that we're not processing Genesis block
      */
-    def groupGovernanceDataByBlock(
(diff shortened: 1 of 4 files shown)


src/main/scala/tech/cryptonomic/conseil/Lorre.scala
```diff
@@ -462,7 +462,9 @@ object Lorre extends App with TezosErrors with LazyLogging with LorreAppConfig w
       } yield ()
     }
 
-    /** Groups data needed for generating Governance */
+    /** Groups data needed for generating GovernanceRow
+     *  Warning! It works on assumption that we're not processing Genesis block
+     */
```
ivanopagano, line 467: can be marked private?

piotrkosecki: nope, it can't

piotrkosecki: just to be clear, compiler does not allow me to add private to method within method

PR #835, opened 2020-06-04
bugfix - fixed governance table to correctly store the proposals

piotrkosecki:
Fixes the issue that governance table did not store proposals - issue was that proposals had the same block_hash, and proposal_hash and it crashed the Lorre when it tried to insert it into the database(it was composite key). Fixed that with adding kind to the key.

### DIFF
--- a/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
+++ b/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
@@ -465,7 +465,9 @@ class TezosIndexer(
       } yield ()
     }
 
-    /** Groups data needed for generating Governance */
+    /** Groups data needed for generating GovernanceRow
+     *  Warning! It works on assumption that we're not processing Genesis block
+     */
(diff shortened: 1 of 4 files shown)


conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
```diff
@@ -460,12 +460,14 @@ class TezosIndexer(
...
-        ).flatMap(_.convertToA[Option, Tables.GovernanceRow])
+        ).map(_.convertTo[Tables.GovernanceRow])
         _ <- db.run(TezosDb.insertGovernance(governanceRows))
       } yield ()
     }
 
-    /** Groups data needed for generating Governance */
+    /** Groups data needed for generating GovernanceRow
+     *  Warning! It works on assumption that we're not processing Genesis block
+     */
```
ivanopagano, line 470: can be marked private?

piotrkosecki: nope, it can't

PR #836, opened 2020-06-05
Make platform discovery generic

ajuszczak:
This pull request prepares metadata endpoint for integrating more blockchains into the code and makes caching "blockchain" aware. 

Please keep in mind, that currently, we can keep multiple blockchains in the same database, but not multiple networks for the same blockchain.

### DIFF
--- a/conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
+++ b/conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
@@ -9,26 +9,31 @@ import akka.http.scaladsl.server.Route
 import akka.stream.ActorMaterializer
 import cats.effect.{ContextShift, IO}
 import ch.megard.akka.http.cors.scaladsl.CorsDirectives.cors
+import com.typesafe.scalalogging.LazyLogging
 import tech.cryptonomic.conseil.api.config.ConseilAppConfig.CombinedConfiguration
 import tech.cryptonomic.conseil.api.directives.{EnableCORSDirectives, RecordingDirectives, ValidatingDirectives}
 import tech.cryptonomic.conseil.api.metadata.{AttributeValuesCacheConfiguration, MetadataService, UnitTransformation}
 import tech.cryptonomic.conseil.api.routes.Docs
 import tech.cryptonomic.conseil.api.routes.info.AppInfo
 import tech.cryptonomic.conseil.api.routes.platform.data.ApiDataRoutes
-import tech.cryptonomic.conseil.api.routes.platform.discovery.PlatformDiscovery
+import tech.cryptonomic.conseil.api.routes.platform.discovery.{GenericPlatformDiscoveryOperations, PlatformDiscovery}
 import tech.cryptonomic.conseil.api.routes.platform.{Api, TezosApi}
 import tech.cryptonomic.conseil.api.security.Security
+import tech.cryptonomic.conseil.common.cache.MetadataCaching
 import tech.cryptonomic.conseil.common.config.Platforms
 import tech.cryptonomic.conseil.common.config.Platforms.BlockchainPlatform
-import tech.cryptonomic.conseil.common.generic.chain.PlatformDiscoveryOperations
+import tech.cryptonomic.conseil.common.sql.DatabaseMetadataOperations
+import tech.cryptonomic.conseil.common.util.DatabaseUtil
 
 import scala.concurrent.ExecutionContext
+import scala.util.{Failure, Success}
 
 object ConseilApi {
   def create(config: CombinedConfiguration)(implicit system: ActorSystem): ConseilApi = new ConseilApi(config)
 }
 
-class ConseilApi(config: CombinedConfiguration)(implicit system: ActorSystem) extends EnableCORSDirectives {
+class ConseilApi(config: CombinedConfiguration)(implicit system: ActorSystem)
+  extends EnableCORSDirectives with LazyLogging {
 
   private val transformation = new UnitTransformation(config.metadata)
   private val cacheOverrides = new AttributeValuesCacheConfiguration(config.metadata)
@@ -117,13 +122,21 @@ class ConseilApi(config: CombinedConfiguration)(implicit system: ActorSystem) ex
       case Platforms.Tezos => new TezosApi(config.metadata, config.server)
     }
 
-    /**
-      * Map, that contains list of available `PlatformDiscoveryOperations` accessed by platform name.
-      * @see `tech.cryptonomic.conseil.common.config.Platforms` to get list of possible platforms.
-      */
-    lazy val cachedDiscoveryOperations: Map[String, PlatformDiscoveryOperations] = cache.map {
-      case (key, value) => key.name -> value.discoveryOperations
+
+    private val cacheOverrides = new AttributeValuesCacheConfiguration(config.metadata)
+    private val metadataCaching = MetadataCaching.empty[IO].unsafeRunSync()
+    private val metadataOperations: DatabaseMetadataOperations = new DatabaseMetadataOperations {
+      override val dbReadHandle = DatabaseUtil.conseilDb
     }
+
+    lazy val cachedDiscoveryOperations: GenericPlatformDiscoveryOperations =
+      new GenericPlatformDiscoveryOperations(
+        metadataOperations,
+        metadataCaching,
+        cacheOverrides,
+        config.server.cacheTTL,
+        config.server.highCardinalityLimit
+      )
 
     /**
       * Map, that contains list of available `ApiDataRoutes` accessed by platform name.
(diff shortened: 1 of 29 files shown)


conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
```diff
@@ -134,14 +147,29 @@ class ConseilApi(config: CombinedConfiguration)(implicit system: ActorSystem) ex
...
-      transformation
-        .overridePlatforms(config.platforms.getPlatforms)
-        .map { platform =>
-          val blockchainPlatform = BlockchainPlatform.fromString(platform.name)
-          blockchainPlatform -> f(blockchainPlatform)
-        }
-        .toMap
+      visiblePlatforms.map { platform =>
+        val blockchainPlatform = BlockchainPlatform.fromString(platform.name)
+        blockchainPlatform -> f(blockchainPlatform)
```
ivanopagano, line 156: I don't remember, might've asked before... can we find a meaning for this `f` function?

Like, it's not abstract or generic, so I'd personally appreciate if it could have a proper name explaining what the function's supposed to get you.

Thanks

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/cache/MetadataCaching.scala
```diff
@@ -68,32 +76,36 @@ class MetadataCaching[F[_]](
...
-    cache.read.map(_.get(CacheKey(key)))
+  private def getFromCache[K <: CacheKey, A](cache: MVar[F, Cache[K, A]])(key: K): F[Option[CacheEntry[A]]] =
+    cache.read.map(_.get(key))
 
   /** Reads all attributes from cache */
   def getAllAttributes: F[AttributesCache] =
     attributesCache.read
 
+  /** Reads all attributes from cache for given seq of keys */
+  def getAllAttributesByKeys: Seq[AttributeCacheKey] => F[AttributesCache] =
```
ivanopagano, line 95: scala `Set` can be used "intentionally" as a function that verifies if an element belongs.

This means that we could use it as input `Set[AttributeCacheKey] => F[AttfributesCache]`
to use it like `attributesCache.read.map(_.filterKeys(keys))`

as `keys <: AttributeCacheKey => Boolean`

Would you think it's too obscure a use?

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/discovery/TestPlatformDiscoveryOperations.scala
```diff
@@ -12,23 +12,23 @@ import scala.concurrent.Future
  **/
 class TestPlatformDiscoveryOperations extends PlatformDiscoveryOperations {
 
-  var entities: List[Entity] = List.empty
-  var attributes: List[Attribute] = List.empty
+  private var entities: Map[String, List[Entity]] = Map.empty.withDefaultValue(List.empty[Entity])
+  private var attributes: Map[String, List[Attribute]] = Map.empty.withDefaultValue(List.empty[Attribute])
```
ivanopagano, line 16:
```suggestion
  private var entities: Map[NetworkPath, List[Entity]] = Map.empty.withDefaultValue(List.empty[Entity])
  private var attributes: Map[EntityPath, List[Attribute]] = Map.empty.withDefaultValue(List.empty[Attribute])
```

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/cache/MetadataCaching.scala
```diff
@@ -15,12 +15,23 @@ object MetadataCaching {
...
 
   /** Class representing key in cache */
-  final case class CacheKey(key: String)
-
-  type Cache[A] = Map[CacheKey, CacheEntry[A]]
-  type AttributesCache = Cache[List[Attribute]]
-  type AttributeValuesCache = Cache[RadixTree[String, String]]
-  type EntitiesCache = Cache[List[Entity]]
+  sealed trait CacheKey {
+    val value: String
```
ivanopagano, line 19: I find this confusing... a key which contains a value...

Is it possible to think of something else if `key` won't work? Like `get`, `text`... don't know

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
ajuszczak: I've changed that to `init` but I've added Scala doc for entire function, which should clarify that a bit.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/cache/MetadataCaching.scala
ajuszczak: I don't think so. I believe we can use that actually and that should not be a big deal. Remember that we need to transform list to set before passing it here, since originally, we are dealing with list. That is the reason of the `keys.contains` function here :)

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/discovery/TestPlatformDiscoveryOperations.scala
ajuszczak: Awesome suggestion!

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/cache/MetadataCaching.scala
ajuszczak: `get` works better, but previously there was `key` inside `CacheKey` which was even more confusing to me :D You could access this like `key.key`.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
```diff
@@ -150,10 +150,17 @@ class ConseilApi(config: CombinedConfiguration)(implicit system: ActorSystem)
     private val visiblePlatforms =
       transformation.overridePlatforms(config.platforms.getPlatforms)
 
-    private def forVisiblePlatforms(f: BlockchainPlatform => Api): Map[BlockchainPlatform, Api] =
+    /**
+      * Function, that while used is going to execute function `f` for every platform found (and visible) in configuration.
```
ivanopagano, line 154: no more `f` function is there 😄

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/cache/MetadataCaching.scala
```diff
@@ -16,22 +16,22 @@ object MetadataCaching {
 
   /** Class representing key in cache */
   sealed trait CacheKey {
-    val value: String
+    def get: String
   }
-  case class EntityCacheKey(platform: String, network: String) extends CacheKey {
-    val value: String = s"$platform.$network"
+  case class EntitiesCacheKey(platform: String, network: String) extends CacheKey {
+    def get: String = s"$platform.$network"
```
ivanopagano, line 22: why `def` here?

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/cache/MetadataCaching.scala
```diff
@@ -16,22 +16,22 @@ object MetadataCaching {
...
   }
-  case class AttributeCacheKey(platform: String, table: String) extends CacheKey {
-    val value: String = s"$platform.$table"
+  case class AttributesCacheKey(platform: String, table: String) extends CacheKey {
+    def get: String = s"$platform.$table"
```
ivanopagano, line 25: why `def` here?

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/cache/MetadataCaching.scala
```diff
@@ -16,22 +16,22 @@ object MetadataCaching {
...
   }
-  case class AttributeValueCacheKey(platform: String, table: String, column: String) extends CacheKey {
-    val value: String = s"$platform.$table.$column"
+  case class AttributeValuesCacheKey(platform: String, table: String, column: String) extends CacheKey {
+    def get: String = s"$platform.$table.$column"
```
ivanopagano, line 28: why `def` here?

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/cache/MetadataCaching.scala
ajuszczak: Mostly, to do not cache yet another field, which can be easily re-computed, but can be `val` if you want :)

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
ajuszczak: Yea... done :D

ajuszczak:
Now, this pull-request depends on https://github.com/Cryptonomic/Conseil/pull/849 Currently there are some tests commented out due to the issue with supporting arrays. Before merging this, above pull-request needs to be merged as well. 

I've decided to add it, just to make sure that this change actually provides more generic approach for platform discovery operations.

Issue #832, opened 2020-06-03
Conseil needs new indices


leenpaws:
In order to fix the previously slow queries being performed due to sequential scan, new indices need to be added. 

![image](https://user-images.githubusercontent.com/18505064/83584906-73ddca00-a516-11ea-8d6c-3eecbbe9892e.png)

![image](https://user-images.githubusercontent.com/18505064/83584955-966fe300-a516-11ea-9f5e-ce5a28409fe6.png)

the following indices were added to prod2 and a remarkable increase in performance was seen
```	CREATE INDEX ix_account_id ON tezos.accounts_history USING btree (account_id);
       	CREATE INDEX ix_manager_pubkey ON tezos.operations USING btree (manager_pubkey);
       	CREATE INDEX ix_operation_group_hash ON tezos.operations USING btree (operation_group_hash);
	CREATE INDEX ix_originated_contracts ON tezos.operations USING btree (originated_contracts);
```

After the change, the slowest queries were:
![image](https://user-images.githubusercontent.com/18505064/83585022-c28b6400-a516-11ea-92fb-6fe859a7fc74.png)

 
![image](https://user-images.githubusercontent.com/18505064/83585033-c919db80-a516-11ea-97b9-85e42b5f57b4.png)
[pgbadger-output.zip](https://github.com/Cryptonomic/Conseil/files/4720481/pgbadger-output.zip)

leenpaws:
Attached is pgbadger output before and after the addition of the indices, ending in main3.html is before the addition.

ivanopagano:
Is this ticket done, @leenpaws ?

PR #837, opened 2020-06-06
add indices to speed up conseil

leenpaws:
Addresses issue #832 

Added the following indices to operations table
manager_pubkey
operation_group_hash
origination_contracts

and to account history table
account_id

This was done in response to the query taking more than 100ms and being the most frequent to be queried for the tezblock block explorer.


sql/conseil.sql
```diff
@@ -585,6 +590,7 @@ CREATE INDEX endorsing_rights_level_idx ON tezos.endorsing_rights USING btree (l
 
 CREATE INDEX endorsing_rights_delegate_idx ON tezos.endorsing_rights USING btree (delegate);
 
+CREATE INDEX ix_delegate_slot ON tezos.endorsing_rights USING btree (delegate);
```
vishakh, line 593: There is no actual slot in the definition of the index.

leenpaws: Fixed.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/Tables.scala
```diff
@@ -1751,6 +1757,9 @@ trait Tables {
 
     /** Index over (level) (database name endorsing_rights_level_idx) */
     val index2 = index("endorsing_rights_level_idx", level)
+
+    /** Index over (delegate) (database name ix_delegate_slot) */
+    val index3 = index("ix_delegate_slot", delegate)
```
ivanopagano, line 1762: the generated index name referred in the val doesn't match with the db-schema which is `ix_delegate`

sql/conseil.sql
```diff
@@ -448,7 +450,9 @@ CREATE TABLE tezos.operations (
     utc_time character varying NOT NULL
 );
 
-
+ CREATE INDEX ix_manager_pubkey ON tezos.operations USING btree (manager_pubkey);
+ CREATE INDEX ix_operation_group_hash ON tezos.operations USING btree (operation_group_hash);
+ CREATE INDEX ix_originated_contracts ON tezos.operations USING btree (originated_contracts);
```
ivanopagano, line 455: I'm honestly wondering how the new indexes don't reflects on the PlatformDiscoveryOperations, that should mention those fields as UniqueKeys... just the same as with the other similar PR targeting the release tag branch...?

Yet the tests passes...

ivanopagano: 🤷‍♂️

ivanopagano: Man! I think I've commented the wrong PR, that's why I got inconsistent findings... 🤦

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/Tables.scala
ivanopagano: the comment in the previous line should be amended accordingly, too

Issue #838, opened 2020-06-08
Follow-up incorrect constraints on proposal governance data


ivanopagano:
As per fix in #835 now the governance rows would be unique per voting_period_kind, which is not correct.

Revert the incorrect situation

ivanopagano:
This should be covered by the latest governance-related code updates

PR #839, opened 2020-06-08
838 - refactor: simplify governance-related code

ivanopagano:
In view of fixing logic issues related to #838, 
this PR's goal is to first clean up the code logic to better rearrange the eventual bugfix.
The latter will be made on top of this changes.

The same kind of changes will be then back-ported to the current tentative release branch.

### DIFF
--- a/conseil-common/src/main/scala/tech/cryptonomic/conseil/common/util/CollectionOps.scala
+++ b/conseil-common/src/main/scala/tech/cryptonomic/conseil/common/util/CollectionOps.scala
@@ -27,6 +27,16 @@ object CollectionOps {
     def byKey(): Map[K, Seq[V]] = groupByKey(seq)
   }
 
+  /** allows counting the occurrences of values in
+    * a generic sequence that might contain duplicates
+    *
+    * @param seq the original sequence
+    */
+  implicit class SeqCounter[T](seq: Seq[T]) {
+    def countValues(): Map[T, Int] =
+      seq.groupBy(identity).mapValues(_.size)
+  }
+
   /**
     * allows to apply a function on collection boundaries, if they're available
     */
--- a/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
+++ b/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
@@ -523,7 +523,7 @@ object TezosDatabaseOperations extends LazyLogging {
   ): DBIO[Seq[Tables.EndorsingRightsRow]] =
     Tables.EndorsingRights.filter(_.level === blockLevel).result
 
-  /** Stores teh governance statistic aggregates in the database
+  /** Stores the governance statistic aggregates in the database
     *
     * @param governance aggregates
     * @return the number of rows added, if available from the driver
--- a/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
+++ b/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
@@ -4,6 +4,7 @@ import java.sql.Timestamp
 import java.time.Instant
 
 import cats.effect.Async
+import cats.implicits._
 import com.typesafe.scalalogging.LazyLogging
 import org.slf4j.LoggerFactory
 import slick.jdbc.PostgresProfile.api._
@@ -19,6 +20,7 @@ import tech.cryptonomic.conseil.indexer.tezos.bigmaps.BigMapsOperations
 import tech.cryptonomic.conseil.indexer.tezos.michelson.contracts.{TNSContract, TokenContracts}
 import tech.cryptonomic.conseil.common.tezos.Tables
 import tech.cryptonomic.conseil.common.util.ConfigUtil
+import tech.cryptonomic.conseil.common.util.CollectionOps._
 import tech.cryptonomic.conseil.common.util.Conversion.Syntax._
 import tech.cryptonomic.conseil.common.util.MathUtil.{mean, stdev}
 import tech.cryptonomic.conseil.indexer.sql.DefaultDatabaseOperations._
@@ -26,6 +28,7 @@ import scala.collection.immutable.Queue
 import scala.concurrent.{ExecutionContext, Future}
 import scala.math.{ceil, max}
 import scala.util.{Failure, Success}
+import tech.cryptonomic.conseil.indexer.tezos.TezosGovernanceOperations.GovernanceAggregate
 
 /**
   * Functions for writing Tezos data to a database.
@@ -520,9 +523,14 @@ object TezosDatabaseOperations extends LazyLogging {
   ): DBIO[Seq[Tables.EndorsingRightsRow]] =
     Tables.EndorsingRights.filter(_.level === blockLevel).result
 
-  def writeGovernance(governance: List[GovernanceRow]): DBIO[Option[Int]] = {
+  /** Stores teh governance statistic aggregates in the database
+    *
+    * @param governance aggregates
+    * @return the number of rows added, if available from the driver
+    */
+  def writeGovernance(governance: List[GovernanceAggregate]): DBIO[Option[Int]] = {
     logger.info("Writing {} governance rows into database...", governance.size)
-    Tables.Governance ++= governance
+    Tables.Governance ++= governance.map(_.convertTo[Tables.GovernanceRow])
   }
 
   def upsertTezosNames(names: List[TNSContract.NameRecord]): DBIO[Option[Int]] = {
@@ -592,7 +600,10 @@ object TezosDatabaseOperations extends LazyLogging {
     }
   }
 
-  /** Gets ballot operations for given cycle */
+  /** Gets ballot operations for given cycle
+    * then sums together all distinct votes (yay, nay, pass) in
+    * a single result object.
+    */
   def getBallotOperationsForCycle(cycle: Int)(implicit ec: ExecutionContext): DBIO[Voting.BallotCounts] =
     Tables.Operations
       .filter(op => op.kind === "ballot" && op.cycle === cycle)
@@ -611,7 +622,10 @@ object TezosDatabaseOperations extends LazyLogging {
         Voting.BallotCounts(yaysCount, naysCount, passesCount)
       }
 
-  /** Gets ballot operations for given level */
+  /** Gets all ballot operations for given level
+    * then sums together all distinct votes (yay, nay, pass) in
+    * a single result object.
+    */
   def getBallotOperationsForLevel(level: Int)(implicit ec: ExecutionContext): DBIO[Voting.BallotCounts] =
     Tables.Operations
       .filter(op => op.kind === "ballot" && op.level === level)
@@ -631,23 +645,18 @@ object TezosDatabaseOperations extends LazyLogging {
       }
 
   /** Gets proposal hashes from operations table for given cycle */
-  def getProposalOperationHashesByCycle(cycle: Int)(implicit ec: ExecutionContext): DBIO[Map[String, Int]] =
+  def getProposalOperationHashesByCycle(cycle: Int)(implicit ec: ExecutionContext): DBIO[Map[ProtocolId, Int]] =
     Tables.Operations
-      .filter(op => op.kind === "proposals" && op.cycle === cycle)
-      .map(_.proposal)
+      .filter(op => op.kind === "proposals" && op.cycle === cycle && op.proposal.nonEmpty)
+      .map(_.proposal.getOrElse("[]"))
       .result
       .map {
-        _.toList
-          .flatMap(_.toList.flatMap(expandStringArray))
-          .groupBy(hash => hash)
-          .map {
-            case (hash, list) => hash -> list.length
-          }
+        _.map(expandStringArray).flatten.countValues
       }
 
   /** Helper method for expanding array storred as a string to list */
-  private def expandStringArray(arr: String): List[String] =
-    arr.trim.stripPrefix("[").stripSuffix("]").split(',').toList
+  private def expandStringArray(arr: String): List[ProtocolId] =
+    arr.trim.stripPrefix("[").stripSuffix("]").split(',').map(ProtocolId).toList
 
   /** Fetch the latest block level available for each account id stored */
   def getLevelsForAccounts(ids: Set[AccountId]): DBIO[Seq[(String, BigDecimal)]] =
--- a/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
+++ b/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
@@ -493,7 +493,7 @@ object TezosDatabaseOperations extends LazyLogging {
     * Writes baking rights to the database
     * @param bakingRights mapping of hash to endorsingRights list
     */
-  def writeBakingRights(bakingRights: List[BakingRights]): DBIO[Option[Int]] = {
+  def insertBakingRights(bakingRights: List[BakingRights]): DBIO[Option[Int]] = {
     berLogger.info("Inserting baking rights to the DB...")
     Tables.BakingRights ++= bakingRights.map(_.convertTo[Tables.BakingRightsRow])
   }
@@ -502,7 +502,7 @@ object TezosDatabaseOperations extends LazyLogging {
     * Writes endorsing rights to the database
     * @param endorsingRights mapping of hash to endorsingRights list
     */
-  def writeEndorsingRights(endorsingRights: List[EndorsingRights]): DBIO[Option[Int]] = {
+  def insertEndorsingRights(endorsingRights: List[EndorsingRights]): DBIO[Option[Int]] = {
     berLogger.info("Inserting endorsing rights to the DB...")
     Tables.EndorsingRights ++= endorsingRights.flatMap(_.convertToA[List, Tables.EndorsingRightsRow])
   }
@@ -528,7 +528,7 @@ object TezosDatabaseOperations extends LazyLogging {
     * @param governance aggregates
     * @return the number of rows added, if available from the driver
     */
-  def writeGovernance(governance: List[GovernanceAggregate]): DBIO[Option[Int]] = {
+  def insertGovernance(governance: List[GovernanceAggregate]): DBIO[Option[Int]] = {
     logger.info("Writing {} governance rows into database...", governance.size)
     Tables.Governance ++= governance.map(_.convertTo[Tables.GovernanceRow])
   }
--- a/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
+++ b/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
@@ -490,7 +490,7 @@ object TezosDatabaseOperations extends LazyLogging {
     * Writes baking rights to the database
     * @param bakingRights mapping of hash to endorsingRights list
     */
-  def insertBakingRights(bakingRights: List[BakingRights]): DBIO[Option[Int]] = {
+  def writeBakingRights(bakingRights: List[BakingRights]): DBIO[Option[Int]] = {
     berLogger.info("Inserting baking rights to the DB...")
     Tables.BakingRights ++= bakingRights.map(_.convertTo[Tables.BakingRightsRow])
   }
@@ -499,7 +499,7 @@ object TezosDatabaseOperations extends LazyLogging {
     * Writes endorsing rights to the database
     * @param endorsingRights mapping of hash to endorsingRights list
     */
-  def insertEndorsingRights(endorsingRights: List[EndorsingRights]): DBIO[Option[Int]] = {
+  def writeEndorsingRights(endorsingRights: List[EndorsingRights]): DBIO[Option[Int]] = {
     berLogger.info("Inserting endorsing rights to the DB...")
     Tables.EndorsingRights ++= endorsingRights.flatMap(_.convertToA[List, Tables.EndorsingRightsRow])
   }
@@ -520,7 +520,7 @@ object TezosDatabaseOperations extends LazyLogging {
   ): DBIO[Seq[Tables.EndorsingRightsRow]] =
     Tables.EndorsingRights.filter(_.level === blockLevel).result
 
-  def insertGovernance(governance: List[GovernanceRow]): DBIO[Option[Int]] = {
+  def writeGovernance(governance: List[GovernanceRow]): DBIO[Option[Int]] = {
     logger.info("Writing {} governance rows into database...", governance.size)
     Tables.Governance ++= governance
   }
--- a/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
+++ b/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
@@ -1,9 +1,10 @@
 package tech.cryptonomic.conseil.indexer.tezos
 
 import tech.cryptonomic.conseil.common.tezos.TezosTypes._
+import tech.cryptonomic.conseil.common.tezos.VotingOperations._
+import scala.concurrent.{ExecutionContext, Future}
 import com.typesafe.scalalogging.LazyLogging
 import slick.jdbc.PostgresProfile.api.Database
-import scala.concurrent.{ExecutionContext, Future}
 import cats.implicits._
 
 /** Process blocks and voting data to compute details for
@@ -44,6 +45,60 @@ object TezosGovernanceOperations extends LazyLogging {
     */
   case class VoteRollsCounts(yay: Int, nay: Int, pass: Int)
 
+  /** Extracts and aggregates the information relative to different
+    * governance periods, restricted to a selection of blocks.
+    * We use the input voting rolls and additional data retrieved both
+    * from the tezos node, for the currently examined blocks, and from
+    * previously indexed voting data entries, extracted from db.
+    * The input keys identifies all blocks we're interested in, independently
+    * from having any actual rolls data in the associated map value.
+    *
+    * @param db the reference database
+    * @param node the operator to get information from the remore tezos node
+    * @param bakerRollsByBlock blocks of interest, with any rolls data available
+    * @return the computed aggregate data
+    */
+  def extractGovernanceAggregations(
+      db: Database,
+      nodeOperator: TezosNodeOperator
+  )(
+      bakerRollsByBlock: Map[Block, List[Voting.BakerRolls]]
+  )(implicit ec: ExecutionContext): Future[List[GovernanceAggregate]] = {
+    // DANGER Will Robinson!
+    // We assume in the rest of the code that we won't be handling the genesis metadata.
+    // The original arguments must not be re-used directly to ensure such guarantee
+    val (blocks, bakerRollsByLevel) = {
+      val nonGenesis = bakerRollsByBlock.filterNot { case (block, _) => block.data.metadata == GenesisMetadata }
+      (nonGenesis.keys.toList, nonGenesis.map { case (block, rolls) => block.data.header.level -> rolls })
+    }
+
+    /* Optimize the remote query for proposals by pre-selecting the blocks */
+    val blocksInActiveVotingPeriod = blocks.filter(votingPeriodIn(activePeriods)).map(_.data)
+
+    /* Here we retrieve node data needed to collect any voting period's interesting numbers.
+     * We want proposals to identify blocks during a voting period, and votes cast in the blocks.
+     * From those we create data aggregates with break-down information per level and
+     * convert those into database rows to be stored.
+     */
+    for {
+      activeProposals <- nodeOperator.getProposals(blocksInActiveVotingPeriod)
+      proposalsMap = activeProposals.collect { case (hash, Some(protocol)) => hash -> protocol }.toMap
+      //collect only those blocks that have a current voting proposal
+      activeProposalsBlocks = blocks.collect {
+        case block if (proposalsMap.contains(block.data.hash)) =>
+          //we know for sure that the value's there: the fallback value is actually never called
+          block -> proposalsMap.getOrElse(block.data.hash, ProtocolId(""))
+      }.toMap
+      ballots <- nodeOperator.getVotes(activeProposalsBlocks.keys.toList)
+      aggregates <- aggregateData(db)(
+        activeProposalsBlocks,
+        ballots.toMap,
+        bakerRollsByLevel
+      )
+    } yield aggregates
+
+  }
+
   /** We take basic governance data and the blocks storing that
     * with the goal of deriving more detailed aggregate data on
     * the voting process.
@@ -64,7 +119,7 @@ object TezosGovernanceOperations extends LazyLogging {
     * @param ec needed to compose concurrent operations
     * @return aggregated data
     */
-  def aggregateData(db: Database)(
+  private def aggregateData(db: Database)(
       proposalsBlocks: Map[Block, ProtocolId],
       proposalsBallots: Map[Block, List[Voting.Ballot]],
       levelsRolls: Map[Int, List[Voting.BakerRolls]]
@@ -76,11 +131,8 @@ object TezosGovernanceOperations extends LazyLogging {
     def countBallotsPerLevel(block: Block) =
       TezosDatabaseOperations.getBallotOperationsForLevel(block.data.header.level)
 
-    // as stated, this is a runtime failure if the block contains the wrong metadata
-    def countBallotsPerCycle(block: Block) = block.data.metadata match {
-      case BlockHeaderMetadata(_, _, _, _, _, level) =>
-        TezosDatabaseOperations.getBallotOperationsForCycle(level.cycle)
-    }
+    def countBallotsPerCycle(cycle: Int) =
+      TezosDatabaseOperations.getBallotOperationsForCycle(cycle)
 
     // as stated, this is a runtime failure if the block contains the wrong metadata
     def proposalOperationsHashes(block: Block) = block.data.metadata match {
@@ -91,30 +143,48 @@ object TezosGovernanceOperations extends LazyLogging {
     logger.info("Searching for governance data in voting period...")
 
     //main algorithm
-    val ballotCountsResult =
+    val cycles = proposalsBlocks.keys
+      .filter(votingPeriodIn(ballotPeriods))
+      .map(_.data.metadata)
+      .collect {
+        case md: BlockHeaderMetadata => md.level.cycle
+      }
+      .toList
+      .distinct
+
+    val cycleBallotCountsResult =
       Future
-        .traverse(proposalsBlocks.keys) { block =>
-          val cycleQuery = db.run(countBallotsPerCycle(block))
-          val levelQuery = db.run(countBallotsPerLevel(block))
-          (cycleQuery, levelQuery).mapN(
-            (cycleCounts, levelCounts) => block -> ((BallotsPerCycle(cycleCounts), BallotsPerLevel(levelCounts)))
-          )
+        .traverse(cycles) { cycle =>
+          db.run(countBallotsPerCycle(cycle)).map(cycle -> BallotsPerCycle(_))
         }
         .map(_.toMap)
 
-    val proposalCountsResult = Future.traverse(proposalsBlocks.keys) { block =>
+    val levelBallotCountsResult =
+      Future
+        .traverse(
+          proposalsBlocks.keys.filter(votingPeriodIn(ballotPeriods))
+        ) { block =>
+          db.run(countBallotsPerLevel(block)).map(block -> BallotsPerLevel(_))
+        }
+        .map(_.toMap)
+
+    val proposalCountsResult = Future.traverse(
+      proposalsBlocks.keys.filter(votingPeriodIs(VotingPeriod.proposal))
+    ) { block =>
       db.run(proposalOperationsHashes(block).map(block -> _))
     }
 
     for {
-      countsMap <- ballotCountsResult
+      levelCountsMap <- levelBallotCountsResult
+      cycleCountsMap <- cycleBallotCountsResult
       proposalCounts <- proposalCountsResult
     } yield
       fillAggregates(
         proposalsBlocks,
         levelsRolls,
         proposalsBallots,
-        countsMap,
+        levelCountsMap,
+        cycleCountsMap,
         proposalCounts.toMap
       )
   }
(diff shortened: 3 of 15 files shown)


conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -0,0 +1,203 @@
...
+    // as stated, this is a runtime failure if the block contains the wrong metadata
+    def proposalOperationsHashes(block: Block) = block.data.metadata match {
+      case BlockHeaderMetadata(_, _, _, _, _, level) =>
+        TezosDatabaseOperations.getProposalOperationHashesByCycle(level.cycle)
+    }
+
+    logger.info("Searching for governance data in voting period...")
+
+    //main algorithm
+    val ballotCountsResult =
```
vishakh, line 94: We can save time here as well. `proposalsBlocks` contains data from three governance periods. However, ballots are only cast during the testing_vote and promotion_vote periods. If we filter out the blocks for the testing period, we will save many database calls.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -0,0 +1,203 @@
...
+        .traverse(proposalsBlocks.keys) { block =>
+          val cycleQuery = db.run(countBallotsPerCycle(block))
+          val levelQuery = db.run(countBallotsPerLevel(block))
+          (cycleQuery, levelQuery).mapN(
+            (cycleCounts, levelCounts) => block -> ((BallotsPerCycle(cycleCounts), BallotsPerLevel(levelCounts)))
+          )
+        }
+        .map(_.toMap)
+
+    val proposalCountsResult = Future.traverse(proposalsBlocks.keys) { block =>
```
vishakh, line 105: Proposal operations are only made during the proposal period. Therefore, we should only run these counts for blocks for which the governance period is 'proposal'.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -0,0 +1,203 @@
...
+   * extract the numbers returning a collector object.
+   */
+  private def fillAggregates(
+      proposalsBlocks: Map[Block, ProtocolId],
+      listingsPerLevel: Map[Int, List[Voting.BakerRolls]],
+      ballots: Map[Block, List[Voting.Ballot]],
+      ballotCounts: Map[Block, (BallotsPerCycle, BallotsPerLevel)],
+      proposalProtocolCounts: Map[Block, Map[ProtocolId, Int]] //comes from individual operations on the block
+  ): List[GovernanceAggregate] =
+    proposalsBlocks.toList.flatMap {
```
vishakh, line 132: I think we have a bug here because we will be missing blocks from the proposal period.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -0,0 +1,203 @@
...
+        TezosDatabaseOperations.getProposalOperationHashesByCycle(level.cycle)
+    }
+
+    logger.info("Searching for governance data in voting period...")
+
+    //main algorithm
+    val ballotCountsResult =
+      Future
+        .traverse(proposalsBlocks.keys) { block =>
+          val cycleQuery = db.run(countBallotsPerCycle(block))
```
vishakh, line 97: Each cycle consists of 4096 blocks, so we will usually not have more than one cycle in any block page. Therefore, it is very redundant to get the same cycle count over and over. Instead, we should get the unique cycle numbers in the block page and fetch this number only once for each cycle number.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -0,0 +1,203 @@
...
+        val ballotCountPerCycle = ballotCounts.get(block).map(_._1.counts)
+        val ballotCountPerLevel = ballotCounts.get(block).map(_._2.counts)
+        val proposalCounts = proposalProtocolCounts.getOrElse(block, Map.empty).toList
+
+        val allRolls = countRolls(listing, ballot)
+        val levelRolls = countRolls(listingByBlock, ballot)
+
+        /* Here we collect a row for the block being considered
+         * to get voting data for the periods with a specific proposal
+         * under scrutinee: testing vote, testing, promotion.
```
vishakh, line 147: Typo: "scrutiny"

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -0,0 +1,203 @@
...
+                  case (proposalProtocol, count) =>
+                    GovernanceAggregate(
+                      block.data.hash,
+                      metadata.copy(voting_period_kind = VotingPeriod.proposal), //we know these are from operations
+                      Some(proposalProtocol),
+                      allRolls,
+                      levelRolls,
+                      Some(Voting.BallotCounts(count, 0, 0)),
+                      ballotCountPerLevel
+                    )
```
vishakh, line 178: Is there a way to express the cons logic a bit more elegantly, e.g. cons two named variables?

vishakh:
Codacy has some non-stupid comments on this PR as well.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/util/CollectionOps.scala
```diff
@@ -27,6 +27,16 @@ object CollectionOps {
     def byKey(): Map[K, Seq[V]] = groupByKey(seq)
   }
 
+  /** allows counting the occurrences of values in
+    * a generic sequence that might contain duplicates
+    *
+    * @param seq the original sequence
+    */
+  implicit class SeqCounter[T](seq: Seq[T]) {
```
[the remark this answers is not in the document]
ivanopagano: Because it's super-verbose and noisy on the call site, and in my experience with scala I've found this pattern to show up many times.

It's actually more general, but I didn't generalise here.

The pattern is to count grouped values in a collection, using a projection function. In this case we use identity as projection since we only want to count values based on equality.

But you can count grouping on a specific field, e.g. you have operations and you want to count how many for each kind, or multiple ballots and you want to count by vote.

I think that having a specific name makes it clear on the usage site, much more than this groupBy mapValues

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosNodeFetchers.scala
```diff
@@ -545,37 +545,6 @@ private[tezos] trait TezosBlocksDataFetchers {
     }
   }
 
-  /** a fetcher of baker rolls for blocks */
-  implicit val bakersFetcher = new FutureFetcher {
```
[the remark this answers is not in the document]
ivanopagano: I merged the usage into a single fetcher.

Previously we fetched bakers both using blocks and block hashes as input. Now we re-use the more specific one with an adapter for the other case.
After all if you have a block, you can always extract the hash and call the fetcher based on the hash.

The corollary is that you need to adapt the output, because a fetcher gives you batches of results indexed on the input value (you get back a sequence of input paired with the output). So you might need to find the original block value associated with each hash, to have a block/bakers pair, instead of hash/bakers.

That's what I've done with `adaptResults`, here: https://github.com/Cryptonomic/Conseil/pull/839/files/76109ec1a5cb373bdbd9010608871ced6e9ce477#diff-a6b43db1952e44cb604adee47f25e528R364

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosNodeOperator.scala
```diff
@@ -343,19 +343,43 @@ private[tezos] class TezosNodeOperator(
...
     import tech.cryptonomic.conseil.common.generic.chain.DataFetcher.fetch
-    fetch[BlockHash, Option[ProtocolId], Future, List, Throwable].run(blocks.filterNot(b => isGenesis(b)).map(_.hash))
+
+    val nonGenesisHashes = blocks.filterNot(isGenesis).map(_.hash)
+    fetch[BlockHash, Option[ProtocolId], Future, List, Throwable].run(nonGenesisHashes)
   }
 
   /** Fetches detailed data for voting associated to the passed-in blocks */
-  def getVotingDetails(blocks: List[Block]): Future[List[BakerBlock]] = {
+  def getRolls(blockHashes: List[BlockHash]): Future[List[(BlockHash, List[Voting.BakerRolls])]] = {
```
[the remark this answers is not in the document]
ivanopagano: Surely make sense, thanks for the feedback

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosNodeOperator.scala
```diff
@@ -343,19 +343,43 @@ private[tezos] class TezosNodeOperator(
...
     import cats.instances.list._
     import tech.cryptonomic.conseil.common.generic.chain.DataFetcher.fetch
 
-    val fetchBakers =
-      fetch[Block, List[Voting.BakerRolls], Future, List, Throwable]
+    fetch[BlockHash, List[Voting.BakerRolls], Future, List, Throwable].run(blockHashes)
+  }
+
+  /** Fetches detailed data for voting associated to the passed-in blocks */
+  def getBlockBakers(blocks: List[Block]): Future[List[BakerBlock]] = {
```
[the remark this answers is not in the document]
ivanopagano: Sure, will do

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
```diff
@@ -352,177 +352,119 @@ class TezosIndexer(
...
-      val safeBlocks = blocks.filterNot(block => block.data.metadata == GenesisMetadata)
-      val safeListings = listings.filterNot { case (block, rolls) => block.data.metadata == GenesisMetadata }
+      val aggregateGovernanceData =
+        TezosGovernanceOperations.aggregateData(db) _
+
+      // DANGER Will Robinson!
+      // We assume in the rest of the code that we won't be handling the genesis metadata.
+      // The original arguments must not be re-used directly to ensure such guarantee
+      val (blocks, levelListings) = {
+        val nonGenesis = rollsPerBlock.filterNot { case (block, _) => block.data.metadata == GenesisMetadata }
```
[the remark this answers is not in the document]
ivanopagano: ok, I re-used the "listings" term as taken from the tezos api endpoint... (which I've found confusing at times, but I thought was conventionally accepted as a term)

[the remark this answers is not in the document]
ivanopagano: > Should we rename `blocks` to `nonGenesisBlocks` for clarity?

I'm not sure, you know, because I see it not as a defining characteristic, as much as a accidental necessity.
We can't extract the values we need for the genesis block, but it's not that it's intrinsically meaningful to discard the genesis from processing.
To me it looks mildly misleading to identify those as nonGenesisBlocks, it's just blocks we need to process, and we can skip the genesis because there's no data there to process.
WDYT?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
```diff
@@ -352,177 +352,119 @@ class TezosIndexer(
...
-              TezosDb.getProposalOperationHashesByCycle(TezosTypes.discardGenesis(block.data.metadata).get.level.cycle)
-            )
-            .map(block -> _)
-        }
-        ballots <- nodeOperator.getVotes(blocksWithProposals)
-        governanceRows = groupGovernanceDataByBlock(
-          blocksWithProposals,
-          proposals.toMap,
-          safeListings.map { case (block, listing) => block.data.header.level -> listing }.toMap,
+        proposals <- nodeOperator.getProposals(blocks.map(_.data))
```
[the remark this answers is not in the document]
ivanopagano: Both sounds like good ideas. I will get the voting period from the block voting_period_kind, if I understand correctly.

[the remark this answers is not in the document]
ivanopagano: the rpc json-schema called it protocol id. Is that the same?
I find confusing when the domain language conflicts with the specs.

Considering that I would prefer to keep the domain objects strongly consistent with the schema definition (for the sake of easy reference with the official documentation), what would you advice for here?
We can use proposal here and in every use site, to be consistent with the logic, and accept that there will be naming clashes.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
ivanopagano: Thanks

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
ivanopagano: Will take care of that, I agree that it's ugly

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
```diff
@@ -352,177 +352,119 @@ class TezosIndexer(
...
     }
 
-    /** Processes blocks and fetches needed data to produce Governance rows*/
+    /** Processes blocks and fetches needed data to produce Governance rows.
+      * We assume that the input mapping will include all block entries relevant
+      * to a batch processing, that is, there could be block keys with a corresponding
+      * empty list as value.
+      * This way we can collapse the arguments to a single one, instead of requiring
+      * a separate one just to list the blocks of interest, and expecting
+      * rolls to only have non-empty values.
```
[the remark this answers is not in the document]
ivanopagano: I will. I probably shouldn't mention alternative design choices that are discarded. It obscures the meaning of the sentence.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
```diff
@@ -352,177 +352,119 @@ class TezosIndexer(
...
+      *
+      * @param rollsPerBlock rolls listings of delegates for the blocks to be processed
+      */
     def processBlocksForGovernance(
-        blocks: List[Block],
-        listings: List[(Block, List[Voting.BakerRolls])]
+        rollsPerBlock: Map[Block, List[Voting.BakerRolls]]
```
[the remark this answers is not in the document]
ivanopagano: makes sense

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
```diff
@@ -352,177 +352,119 @@ class TezosIndexer(
...
     ): Future[Unit] = {
-      import TezosDatabaseConversions._
-      import tech.cryptonomic.conseil.common.util.Conversion.Syntax._
-
-      //DANGER Will Robinson! we assume in the rest of the code that we're not handling the genesis block
-      val safeBlocks = blocks.filterNot(block => block.data.metadata == GenesisMetadata)
-      val safeListings = listings.filterNot { case (block, rolls) => block.data.metadata == GenesisMetadata }
+      val aggregateGovernanceData =
```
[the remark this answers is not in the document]
ivanopagano: Yes, you're right. I think I originally thought something different but eventually it remained like this, which doesn't make much sense.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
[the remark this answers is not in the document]
vishakh: > > Should we rename `blocks` to `nonGenesisBlocks` for clarity?
> 
> I'm not sure, you know, because I see it not as a defining characteristic, as much as a accidental necessity.
> We can't extract the values we need for the genesis block, but it's not that it's intrinsically meaningful to discard the genesis from processing.
> To me it looks mildly misleading to identify those as nonGenesisBlocks, it's just blocks we need to process, and we can skip the genesis because there's no data there to process.
> WDYT?

Interestingly, `rollsPerBlock` will not have an entry for the genesis block since the corresponding RPC query comes back with no value! Not that I am saying it's not good to always validate your data. 

I agree with your comment otherwise.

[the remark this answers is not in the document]
vishakh: > ok, I re-used the "listings" term as taken from the tezos api endpoint... (which I've found confusing at times, but I thought was conventionally accepted as a term)

I think the Tezos node uses listings within some context so it's clear what the listings are referring to. I think the RPC route is something like '../governance/listings'. That context is not explicitly present here.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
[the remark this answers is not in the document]
vishakh: Yep, you understand correctly!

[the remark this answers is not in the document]
vishakh: Yes, you are right. It's good to be aligned with the domain object even when it's inaccurate. Perhaps leave a comment noting this in the data fetcher?

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/util/CollectionOps.scala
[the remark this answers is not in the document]
vishakh: Ok, thanks.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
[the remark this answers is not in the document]
ivanopagano: > Interestingly, `rollsPerBlock` will not have an entry for the genesis block since the corresponding RPC query comes back with no value! Not that I am saying it's not good to always validate your data.
> 

I think the same, but as you say, I would not risk, considering we already had troubles with those genesis metadata values

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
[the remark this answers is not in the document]
ivanopagano: I've put the comment in the node operator method

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -0,0 +1,203 @@
...
+    * @param levelsRolls listings of baker rolls involved, indexed per level, we expect all levels for the
+    *                         `proposalsBlocks` to be available, plus the level immediately before this batch.
+    *                          ideally this would be a `Level => List[Voting.BakerRolls]` (i.e. a total function)
+    * @param ec needed to compose concurrent operations
+    * @return aggregated data
+    */
+  def aggregateData(db: Database)(
+      proposalsBlocks: Map[Block, ProtocolId],
+      proposalsBallots: Map[Block, List[Voting.Ballot]],
+      levelsRolls: Map[Int, List[Voting.BakerRolls]]
```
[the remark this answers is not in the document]
ivanopagano: I had the same thoughts. I originally left it like this to avoid another dependency for the governance operator, both on database and on the node operator.

But it's probably cleaner like you suggest.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -0,0 +1,203 @@
...
+    * with the goal of deriving more detailed aggregate data on
+    * the voting process.
+    * The computation needs to access the database to fetch essential
+    * ballot data
+    *
+    * SAFETY NOTE: make sure that the blocks passed to this method have
+    *              have a proper metadata value, different from the genesis one
+    *
+    * @param db needed to fetch previously stored voting data
+    * @param proposalsBlocks blocks we want to have governance data for, associated to the protocol of the proposal
```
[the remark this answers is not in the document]
ivanopagano: > This is not quite accurate. IIUC, `proposalBlocks` will only contain blocks for three of the four governance periods, i.e. not for the proposal period (as there is no active proposal during the proposal period). We actually want to have governance data for all four periods.

I've been a bit nasty with this, I knew I would've fixed that on the following PR, so I wrote the docs as they should've been eventually.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -0,0 +1,203 @@
...
+    * @param proposalsBallots votes cast on the blocks of interest,
```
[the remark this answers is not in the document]
ivanopagano: Just like the name for blocks, this is inaccurate because of the ephemeral nature of this PR, whose goal is only to separate refactors in preparation of the actual fix, such that the changes will be easier to validate

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -0,0 +1,203 @@
...
+    *                         the keys should be a superset of the `proposalsBlocks` keys,
+    *                         so that the value is available for each of those blocks.
+    * @param levelsRolls listings of baker rolls involved, indexed per level, we expect all levels for the
```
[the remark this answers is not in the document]
ivanopagano: Same considerations as for the name of other parameters

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
ivanopagano: will do

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
ivanopagano: will do

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
ivanopagano: That's the subject of the following PR, which is already waiting

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
ivanopagano: Sounds like a smart thing to do

ivanopagano:
> Codacy has some non-stupid comments on this PR as well.

The 50+ long methods are
 - aggregateData which includes local methods that could be bringing the LoC count over the threshold. I tend to use local methods to encapsulate code relevant only to a single task, as opposed to class private support methods that won't be re-used nowhere else, polluting the class scope
 - fillAggregate logic is simple, the LoC without comments are actually less than 50.

In general I tend to see no terrible value in LoC metrics.
 * Scala is a "vertically stacked" language from commonly used styles, and I like it because it better matches our brain capabilities at scanning writings than horizontally spanned style, like java. YMMV
 * Measuring method quality on LoC is not universally accepted. Lots of tiny functions might not necessarily make more sense than a single cohesive one, reducing the mental burden of jumping back and forth between the functions, unless they're well defined and improve overall understanding of the semantic of the operation. I'm trying to say that, as with many things, just splitting for the sake of splitting is not necessarily good. Once again YMMV

This is to say that I'm not convinced to spend an effort on this, but I'm open to discussion, as always.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
```diff
@@ -490,7 +493,7 @@ object TezosDatabaseOperations extends DefaultDatabaseOperations("tezos") with L
     * Writes baking rights to the database
     * @param bakingRights mapping of hash to endorsingRights list
     */
-  def insertBakingRights(bakingRights: List[BakingRights]): DBIO[Option[Int]] = {
+  def writeBakingRights(bakingRights: List[BakingRights]): DBIO[Option[Int]] = {
```
piotrkosecki, line 496: I think we should consider consistent naming - in one of my previous PRs you asked me to change names to begin with `write` when there were upsert operations, here were doing inserts

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
```diff
@@ -499,7 +502,7 @@ object TezosDatabaseOperations extends DefaultDatabaseOperations("tezos") with L
     * Writes endorsing rights to the database
     * @param endorsingRights mapping of hash to endorsingRights list
     */
-  def insertEndorsingRights(endorsingRights: List[EndorsingRights]): DBIO[Option[Int]] = {
+  def writeEndorsingRights(endorsingRights: List[EndorsingRights]): DBIO[Option[Int]] = {
```
piotrkosecki, line 505: the same as above

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -0,0 +1,279 @@
...
+object TezosGovernanceOperations extends LazyLogging {
+
+  /** Collects all relevant data for the governance
+    * voting process
+    *
+    * @param hash identifies a specific block
+    * @param metadata the block metadata
+    * @param proposalId a specific proposal under evaluation
+    * @param allRolls ballot rolls up to this block for the proposal
+    * @param singleBlockRolls ballot rolls included in this specific block
```
piotrkosecki, line 22: please change name in the doc

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -0,0 +1,279 @@
...
+      proposalsBlocks: Map[Block, ProtocolId],
+      listingsPerLevel: Map[Int, List[Voting.BakerRolls]],
+      ballots: Map[Block, List[Voting.Ballot]],
+      ballotCountsPerLevel: Map[Block, BallotsPerLevel],
+      ballotCountsPerCycle: Map[Int, BallotsPerCycle],
+      proposalProtocolCounts: Map[Block, Map[ProtocolId, Int]] //comes from individual operations on the block
+  ): List[GovernanceAggregate] =
+    proposalsBlocks.toList.flatMap {
+      case (block, proposal) =>
+        val listing = listingsPerLevel.get(block.data.header.level).getOrElse(List.empty)
```
piotrkosecki, line 205: you can change this to:
`listingsPerLevel.getOrElse(block.data.header.level, List.empty)`

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -0,0 +1,279 @@
...
+        val prevListings = listingsPerLevel.get(block.data.header.level - 1).getOrElse(List.empty)
```
piotrkosecki, line 206: Similarly here
`listingsPerLevel.getOrElse(block.data.header.level - 1, List.empty)`

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
ivanopagano: I'm not sure I follow.
I was trying to be consistent with what I thought was a pattern:
all insert operations were initially called `writeX`.

When some upsert operation was beginning to be used I think I decided to highlight them as such to make it clear that existing records would be overwritten.
I didn't chose upsert to reflect the db naming, it was just the idea of "update when present, or insert a new one", as a single word.
I checked the db operations and no insert was named as such... I opted to align these methods with the rest.

If I'm missing something please point it out, and overall we can decide to define together an explicit naming policy and put it somewhere written as a reference

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
ivanopagano: done

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
ivanopagano: So right!

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
ivanopagano: Yea!

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
piotrkosecki: `writeAccounts`, `writeBakers`, `writeBakersAndCopyContracts` use `insertOrUpdateAll` - I think you requested here to change update to write, and that's what got me confused:
https://github.com/Cryptonomic/Conseil/pull/797#discussion_r429145776

ivanopagano: I must be more confused than you are...
I'm fine as long as we got a consistent naming scheme. My memory is probably failing me in this.

piotrkosecki: I agree, we should get consistent naming there 👍

ivanopagano: I'm reading the comment you mentioned once more...
and it seems consistent with my request to keep naming all methods that writes data as `writeX`, be it insert or upsert alike.
I see that of the existing `upsertX` methods, I did slip once and used `upsertTezosNames`... I should probably change that too.

As I said, I'm ok to differentiate, as long as
 1. we try to keep it consistent
 2. the differentiation is meaningful at the domain level, not at the database level (that would be an implementation leak)

What do you propose?

piotrkosecki: For now we can leave it as it is, but we can mark it as an improvement for the future.

ivanopagano: I'm opening a separate issue to address the point, which is more involved and probably is not so relevant to this PR anyway

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosNodeOperator.scala
```diff
@@ -338,24 +338,60 @@ private[tezos] class TezosNodeOperator(
       (fetchCurrentQuorum, fetchCurrentProposal).mapN(CurrentVotes.apply)
     }
 
-  /** Fetches proposals for given blocks */
+  /** Fetches proposals for given blocks
+    * The returned value, when available for active voting periods, is the ID of
+    * a proposal, though for the sake of naming consistency with tezos schema
+    * we refer to that as a protocol ID.
+    */
```
vishakh, line 345: The word 'proposals' is ambiguous. It could mean proposal operations, suggested proposal hashes or active proposals. Therefore, I recommend renaming this function to `getActiveProposals`.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosNodeFetchers.scala
```diff
@@ -502,15 +502,15 @@ private[tezos] trait TezosBlocksDataFetchers {
   }
 
   /** a fetcher of baker rolls for blocks */
-  implicit val bakersFetcherByHash = new FutureFetcher {
+  implicit val bakersFetcher = new FutureFetcher {
```
vishakh, line 505: Related to a previous comment of mine..here we are fetching baker rolls. `bakersFetcher` has ambiguous meaning. Could we rename this to something like `bakerRollsFetcher`?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosNodeOperator.scala
```diff
@@ -338,24 +338,60 @@ private[tezos] class TezosNodeOperator(
...
-      fetch[Block, List[Voting.BakerRolls], Future, List, Throwable]
+    fetch[BlockHash, List[Voting.BakerRolls], Future, List, Throwable].run(blockHashes)
+  }
+
+  /** Fetches rolls for all bakers of individual blocks
+    *
+    * @param blocks the blocks we want the rolls for
+    * @return the lists of rolls, for each individual requested block
+    */
+  def getBakerRollsForBlocks(blocks: List[Block]): Future[List[BakerBlock]] = {
```
vishakh, line 373: `BakerBlock` is ambiguous as well. Can we rename the type to something like `BakerRollsByBlock`?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
```diff
@@ -352,176 +352,86 @@ class TezosIndexer(
...
         _ <- tnsOperations.processNamesRegistrations(blocks).flatMap(db.run)
-        votingData <- processVotesForBlocks(results.map { case (block, _) => block })
-        rollsData = votingData.map {
-          case (block, rolls) => block.data.hash -> rolls
-        }
-        delegateCheckpoints <- processAccountsForBlocks(accountUpdates, rollsData) // should this fail, we still recover data from the checkpoint
-        _ <- processBakersForBlocks(delegateCheckpoints) // same as above
-        _ <- processBlocksForGovernance(blocks, votingData)
-        _ <- updateBakers(blocks)
+        bakersData <- nodeOperator.getBakerRollsForBlocks(blocks)
```
vishakh, line 361: Perhaps this should be `bakerRollsData`?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
```diff
@@ -352,176 +352,86 @@ class TezosIndexer(
...
+        rollsByHash = bakersData.map { case (block, rolls) => block.data.hash -> rolls }.toMap
```
vishakh, line 362: `rollsByBlockHash`?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -0,0 +1,279 @@
...
+    * governance periods, restricted to a selection of blocks.
+    * We use the input voting rolls and additional data retrieved both
+    * from the tezos node, for the currently examined blocks, and from
+    * previously indexed voting data entries, extracted from db.
+    * The input keys identifies all blocks we're interested in, independently
+    * from having any actual rolls data in the associated map value.
+    *
+    * @param db the reference database
+    * @param node the operator to get information from the remore tezos node
+    * @param bakerRollsByBlock blocks of interest, with any rolls data available
```
vishakh, line 58: This is out of date.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -0,0 +1,279 @@
...
+    * @return the computed aggregate data
+    */
+  def extractGovernanceAggregations(
+      db: Database,
+      nodeOperator: TezosNodeOperator
+  )(
+      bakerRollsByBlock: Map[Block, List[Voting.BakerRolls]]
+  )(implicit ec: ExecutionContext): Future[List[GovernanceAggregate]] = {
```
vishakh, line 66: Passing around the execution context is annoying but I think the code structure is much better now! :+1:

vishakh:
@ivanopagano I reviewed all three governance PRs. Besides the comments I left, I don't believe a performance bug I previously described has been fixed yet. If you recall, the ballots per cycle is being fetched from the database for each block in a given block page. This is redundant because there are probably no more than two cycles for the entire page. We are therefore making many, many redundant queries against the database. We should fetch all unique cycles in a given block page, pre-fetch the ballot counts and then store them in memory for later usage.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
ivanopagano: Is `rollsData` enough? it's a local name, after all

ivanopagano:
> @ivanopagano I reviewed all three governance PRs. Besides the comments I left, I don't believe a performance bug I previously described has been fixed yet. If you recall, the ballots per cycle is being fetched from the database for each block in a given block page. This is redundant because there are probably no more than two cycles for the entire page. We are therefore making many, many redundant queries against the database. We should fetch all unique cycles in a given block page, pre-fetch the ballot counts and then store them in memory for later usage.

I'm quite sure of having addressed that, I'ill double check and eventually fix the error or refer here to the changes.

ivanopagano:
> > @ivanopagano I reviewed all three governance PRs. Besides the comments I left, I don't believe a performance bug I previously described has been fixed yet. If you recall, the ballots per cycle is being fetched from the database for each block in a given block page. This is redundant because there are probably no more than two cycles for the entire page. We are therefore making many, many redundant queries against the database. We should fetch all unique cycles in a given block page, pre-fetch the ballot counts and then store them in memory for later usage.
> 
> I'm quite sure of having addressed that, I'ill double check and eventually fix the error or refer here to the changes.

@vishakh isn't this piece of code handling your concerns?
https://github.com/Cryptonomic/Conseil/pull/845/files#diff-1270eeb68d55da036dc8de466f32a7d1R159
Here first cycles are collected for all blocks involved (which should turn out to be 1-2 at most, as you said) and then we collect ballots counts only for each cycle value

PR #841, opened 2020-06-09
Add indices to speed up conseil

leenpaws:
Mirrors PR 837 in master branch, addresses issue #832


leenpaws:
Requires reopening to add more indices

PR #845, opened 2020-06-11
843 - 844 - Missing proposals data from governance and fix incorrect rolls counts per level

### LINKED ISSUE #843: Incorrect rolls counts in governance data for batch lead block
Incorrect roll counts can be seen for some block levels on the governance data.

relates to #762 

Reported:
>when running the following query [...] with release branch r3:
>bug 1: Running the following query comes back with bad data for the yay, nay and total count and rolls `select * from tezos.governance where level between 360499 and 360505 order by level desc;`
>the block_yay_rolls column is wrong for level 360500.

Diagnosis: the rolls listing loaded during processing of block batches for the governance data storage will not include levels outside of the batch itself, which leads to the wrong values for the "lead block" in the batch, which misses previous data needed for the computation

Note: the plan is to create a fix PR for both the master branch and a back-port to the release branch

### LINKED ISSUE #844: Missing data for proposals in governance entries
Blocks with proposals seems to be missing from the governance data collected.

relates to #762 

Reported:
>when running the following query [...] with release branch r3:
>Bug #2: In the db the earliest known proposal operation is at block 332624, and the repo has the conf file stating `batchedFetches.governanceStartLevel: 350000`, the governance table only has records starting at level 360500

Diagnosis: the code collects proposal counts for all relevant blocks in a processing batch but generates data only for those having a current proposal, i.e. the proposal operations are not included, since they are from a different voting period

Note: the plan is to create a fix PR for both the master branch and a back-port to the release branch

ivanopagano:
Fixes #843 and #844 

The code to extract governance rows now takes all blocks into considerations, and differentiates between those that have a current proposal or not.
Hence we now have also proposals data that was not collected before.

Use stored data for rolls from the previous batch to account for the missing entry to compute per-level rolls counts.

### DIFF
--- a/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
+++ b/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
@@ -3,8 +3,12 @@ package tech.cryptonomic.conseil.indexer.tezos
 import tech.cryptonomic.conseil.common.tezos.TezosTypes._
 import tech.cryptonomic.conseil.common.tezos.VotingOperations._
 import scala.concurrent.{ExecutionContext, Future}
+import scala.math.max
+import scala.util.Try
 import com.typesafe.scalalogging.LazyLogging
+import slick.dbio.DBIO
 import slick.jdbc.PostgresProfile.api.Database
+import cats.implicits._
 
 /** Process blocks and voting data to compute details for
   * the governance-related cycles
@@ -43,6 +47,39 @@ object TezosGovernanceOperations extends LazyLogging {
     * @param pass how many passes
     */
   case class VoteRollsCounts(yay: Int, nay: Int, pass: Int)
+
+  object VoteRollsCounts {
+
+    /** a wrapper with all zeroes */
+    val zero = VoteRollsCounts(0, 0, 0)
+
+    /** defines substraction between counts, with the caveat
+      * that no result of the substraction can have negative counts,
+      * which would have no meaning
+      *
+      * @param subtrahend
+      * @param minuend
+      * @return the "zero-min" difference of the counts, element by element
+      */
+    def subtract(subtrahend: VoteRollsCounts, minuend: VoteRollsCounts): VoteRollsCounts = (subtrahend, minuend) match {
+      case (VoteRollsCounts(y1, n1, p1), VoteRollsCounts(y2, n2, p2)) =>
+        /* We make use of cats implicit instances which defines general operations that applies to Ints too:
+         * - |+| sums two integers
+         * - i.inverse() changes the sign, i.e. i.inverse() === -i
+         * Therefore we do substraction by adding the negated number.
+         *
+         * Cats provides the additional facilities to do that for a whole triplet of Ints
+         * in a single call, such that corresponding elements of the tuple are subtracted
+         * as you might naturally want to do.
+         *
+         * We subtract the three values of the rolls counts each by each as 2 triplets.
+         * Then we make sure to avoid results of negative rolls counts.
+         */
+        val (y, n, p) = (y1, n1, p1) |+| ((y2, n2, p2).inverse())
+        VoteRollsCounts(max(y, 0), max(n, 0), max(p, 0))
+    }
+
+  }
 
   /** Extracts and aggregates the information relative to different
     * governance periods, restricted to a selection of blocks.
@@ -84,7 +121,7 @@ object TezosGovernanceOperations extends LazyLogging {
       proposalsMap = activeProposals.toMap
       //pair blocks and proposals
       blocksAndProposals = blocks.map(block => block -> proposalsMap.getOrElse(block.data.hash, None)).toMap
-      ballots <- nodeOperator.getVotes(blocksAndProposals.keys.toList)
+      ballots <- nodeOperator.getVotes(blocks)
       aggregates <- aggregateData(db)(
         blocksAndProposals,
         ballots.toMap,
@@ -177,10 +214,16 @@ object TezosGovernanceOperations extends LazyLogging {
       db.run(proposalHashesPerCycle(block).map(block -> _))
     }
 
+    val previousRollsResult = db
+      .run(
+        queryPreviousBatchStats(activeProposalsBlocks.keySet)
+      )
+
     for {
       levelCountsMap <- levelBallotCountsResult
       cycleCountsMap <- cycleBallotCountsResult
       proposalCounts <- proposalCountsResult
+      previousBatchRolls <- previousRollsResult
     } yield
       fillAggregates(
         activeProposalsBlocks,
@@ -188,7 +231,8 @@ object TezosGovernanceOperations extends LazyLogging {
         activeProposalsBallots,
         levelCountsMap,
         cycleCountsMap,
-        proposalCounts.toMap
+        proposalCounts.toMap,
+        previousBatchRolls.getOrElse(VoteRollsCounts.zero)
       )
   }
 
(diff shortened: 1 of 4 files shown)


conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -83,17 +83,25 @@ object TezosGovernanceOperations extends LazyLogging {
...
         TezosDatabaseOperations.getProposalOperationHashesByCycle(level.cycle)
     }
 
     logger.info("Searching for governance data in voting period...")
 
+    //find blocks for a specific proposal under scrutiny, relevant for counting ballots
+    val votingBlocks = proposalsBlocks.collect { case (block, Some(protocol)) => block }.toList
+
+    logger.info(
+      "There are {} blocks related to some voting - excluding proposals.",
```
vishakh, line 97: Let's make this a bit clearer: "There are {} blocks related to testing vote and proposal vote periods..".

ivanopagano: done, thanks

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -230,17 +234,26 @@ object TezosGovernanceOperations extends LazyLogging {
...
+                )
             )
             val proposalOperationsAggregates =
-              proposalCounts.map { //these come from all individual proposal operations during the proposal period
+              proposalCounts //these come from all individual proposal operations during the proposal period
+              .filterNot {
+                /* this should never be the case: the proposal currently under evaluation
+                 * should not be proposed in operations of the same block
+                 */
+                case (proposal, _) => currentProposal.exists(_ == proposal)
```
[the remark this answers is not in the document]
ivanopagano: nice one

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -119,8 +115,8 @@ object TezosGovernanceOperations extends LazyLogging {
     * @param ec needed to compose concurrent operations
     * @return aggregated data
     */
-  private def aggregateData(db: Database)(
-      proposalsBlocks: Map[Block, ProtocolId],
+  def aggregateData(db: Database)(
```
vishakh, line 118: `proposalsBlocks` and `proposalsBallots` are still ambiguous. They specifically leave out data for the proposal governance period. I would renamed them to `activeProposalsBlocks` and `activeProposalsBallots`.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -193,15 +197,15 @@ object TezosGovernanceOperations extends LazyLogging {
    * extract the numbers returning a collector object.
    */
   private def fillAggregates(
-      proposalsBlocks: Map[Block, ProtocolId],
+      proposalsBlocks: Map[Block, Option[ProtocolId]],
```
vishakh, line 200: `listingsPerLevel` is still too generic a name.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -193,15 +197,15 @@ object TezosGovernanceOperations extends LazyLogging {
...
       listingsPerLevel: Map[Int, List[Voting.BakerRolls]],
       ballots: Map[Block, List[Voting.Ballot]],
       ballotCountsPerLevel: Map[Block, BallotsPerLevel],
       ballotCountsPerCycle: Map[Int, BallotsPerCycle],
       proposalProtocolCounts: Map[Block, Map[ProtocolId, Int]] //comes from individual operations on the block
```
vishakh, line 205: `proposalProtocolCounts` is misleading. It implies we are counting the proposal operation counts by protocol ID. A better name might be something like `proposalCountsByBlock`.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -230,17 +234,26 @@ object TezosGovernanceOperations extends LazyLogging {
...
-              ballotCountPerCycle,
-              ballotCountPerLevel
+            val activeProposalAggregate = currentProposal.map(
+              proposal =>
+                GovernanceAggregate(
+                  block.data.hash,
+                  metadata,
+                  Some(proposal),
+                  allRolls,
+                  levelRolls,
```
vishakh, line 244: We have a bug here. The current logic is correct for the testing_vote and promotion_vote periods where we want to show the number of rolls corresponding to `ballot` operations. However, for the proposal period, we also want to show the number of rolls corresponding to `proposal` operations. We are missing the rolls counts for proposal operations here. IIUC, the column `yay_rolls` for proposal records will always be zero.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
```diff
@@ -119,8 +115,8 @@ object TezosGovernanceOperations extends LazyLogging {
     * @param ec needed to compose concurrent operations
     * @return aggregated data
     */
-  private def aggregateData(db: Database)(
-      proposalsBlocks: Map[Block, ProtocolId],
+  def aggregateData(db: Database)(
+      proposalsBlocks: Map[Block, Option[ProtocolId]],
```
vishakh, line 119: We probably have a bug here too. `proposalsBlocks` is ultimately derived from `TezosGovernanceOperations`, line 84:

```scala
activeProposals <- nodeOperator.getProposals(blocksInActiveVotingPeriod)
```

This only picks up blocks with active proposals. This excludes blocks from the proposal period where there is no one active proposal. When you run your local testing, you will not see any `governance` records from the proposals periods. Once you fix the bug, you will start seeing valid records for the proposals periods. Please use this as a validation of your code.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
ivanopagano: renamed

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
ivanopagano: I've renamed those as `rollsByLevel` both here and in the code that pass them through

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
ivanopagano: I have to disagree here: `activeProposals` collects by block any active proposal id, yet immediately after we associate those _optionally_ , with the original block list, at line 87.
That is, we associate each block with a proposal id, if there's one, otherwise the optional value is empty.

It's a sparse associative array, there are holes in the map values, corresponding to the proposal periods

ivanopagano: In fact I did test locally and saw those entries being saved, where previously there were none.

I can check once more, but I have logs that still shows

```
17:19:52.557 [lorre-tezos-indexer-akka.actor.default-dispatcher-40] INFO  t.c.c.i.t.TezosGovernanceOperations$ - Searching for governance data in voting period...
17:19:52.558 [lorre-tezos-indexer-akka.actor.default-dispatcher-40] INFO  t.c.c.i.t.TezosGovernanceOperations$ - There are no blocks related to some voting - excluding proposals.
17:19:53.502 [lorre-tezos-indexer-akka.actor.default-dispatcher-40] INFO  t.c.c.i.t.TezosDatabaseOperations$   - Writing 1000 governance rows into database...
```

Which attest that 1000 rows were written, even if no active voting period blocks were found

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosGovernanceOperations.scala
ivanopagano: I'm asking because that's where all our rolls counts descend from.
We collect the responses of such endpoint for each block, without consider the voting period the block's been baked in.

vishakh: No, you are right. The logic is, in fact, correct!

PR #853, opened 2020-06-16
852 - verify correct parameters conversion in transactions

### LINKED ISSUE #852: Inconsistent values for parameters and parameters_micheline
Sending the following query to the `/v2/data/tezos/mainnet/operations` endpoint:

```json
{
    "fields": ["parameters", "parameters_micheline"],
    "predicates": [
        {
            "field": "operation_group_hash",
            "operation": "eq",
            "set": [
                "opKYnbone62mtx6tNhkbPRbawmHzXZXwuAmHoSZKtGjhtUjpSaM"
            ],
            "inverse": false
        }
    ]
}
```

results in:

```json
[
    {
        "parameters": "Pair \"tz1XVRuKfcbCGWK4pvJaxyFAN7ayhoFtdWEt\" (Pair \"tz1SbC5uEwQ2114PhHooPhLr8T5NLEghbToz\" 1)",
        "parameters_micheline": "[{\"prim\":\"DROP\"},{\"prim\":\"NIL\",\"args\":[{\"prim\":\"operation\"}]},{\"prim\":\"PUSH\",\"args\":[{\"prim\":\"key_hash\"},{\"string\":\"tz1TG2LGuJYpRPa3AAW6bugGgy2LiyxrUN7w\"}]},{\"prim\":\"IMPLICIT_ACCOUNT\"},{\"prim\":\"PUSH\",\"args\":[{\"prim\":\"mutez\"},{\"int\":\"60000000\"}]},{\"prim\":\"UNIT\"},{\"prim\":\"TRANSFER_TOKENS\"},{\"prim\":\"CONS\"}]"
    }
]
```

The `parameters_micheline` value is clearly wrong since it should be just the micheline version of the `parameters` value. Not sure if this is happening only for this specific operation or if there are other inconsistencies, I only found this one so far.

ivanopagano:
Fix #852 

Actually no fix is included in this PR, as the code is already correct on this branch.
The issue was opened referring to the production deployment which is still missing the latest code updates.
This PR adds missing unit tests that verify the situation described in the issue shouldn't manifest itself.

Since the issue was related to some "Optics" that we're using, I took the chance to clean-up/re-organize the optics code we had across the codebase, such that:
 * all optics related code is collected in a single code module
 * introductory documentation on how to use monocle (optics library) is available
 * there is a consistent naming convention throughout the code
 * any existing test and the new ones are all collected in a single consistently named unit test


project/Dependencies.scala
```diff
@@ -22,9 +22,9 @@ object Dependencies {
     val postgres = "42.1.4"
 
     val endpoints = "0.9.0"
-    val cats = "1.6.0"
+    val cats = "2.1.1"
```
piotrkosecki, line 25: what's included in new cats that we use here?

ivanopagano: The Monocle latest is based on that

ivanopagano: Or, to be precise, it's based on cats 2.x

PR #857, opened 2020-06-18
Revert "Add indices to speed up conseil"

leenpaws:
Reverts Cryptonomic/Conseil#841, need to add more indices

### DIFF
--- a/sql/conseil.sql
+++ b/sql/conseil.sql
@@ -186,7 +186,7 @@ CREATE TABLE tezos.accounts_history (
     is_activated boolean NOT NULL DEFAULT false,
     is_active_baker boolean
 );
-CREATE INDEX ix_account_id ON tezos.accounts_history USING btree (account_id);
+
 --
 -- Name: accounts_checkpoint; Type: TABLE; Schema: tezos; Owner: -
 --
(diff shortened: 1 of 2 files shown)


sql/conseil.sql
```diff
@@ -555,13 +552,14 @@ ALTER TABLE ONLY tezos.operations
...
 CREATE INDEX baking_rights_level_idx ON tezos.baking_rights USING btree (level);
-
+CREATE INDEX ix_delegate_priority ON tezos.baking_rights USING btree (delegate, priority);
 
 --
 -- Name: endorsing_rights_level_idx; Type: INDEX; Schema: tezos; Owner: -
 --
 
 CREATE INDEX endorsing_rights_level_idx ON tezos.endorsing_rights USING btree (level);
+CREATE INDEX ix_delegate_slot ON tezos.endorsing_rights USING btree (delegate);
```
vishakh, line 562: Is this actually correct? The name of the index contains `slot` but there is no slot in the actual definition.

src/main/scala/tech/cryptonomic/conseil/tezos/Tables.scala
```diff
@@ -1592,6 +1592,9 @@ trait Tables {
 
     /** Index over (level) (database name endorsing_rights_level_idx) */
     val index1 = index("endorsing_rights_level_idx", level)
+
+    /** Index over (delegate) (database name ix_delegate_slot) */
+    val index2 = index("ix_delegate_slot", delegate)
```
ivanopagano, line 1597: The index reference here is not up-to-date, the Tables file needs be re-generated.
Alternatively remove the `_slot` appendix here and in the comment in the previous line.

PR #859, opened 2020-06-23
581 - modular splitting of tezos indexer

### LINKED ISSUE #581: Refactor main body of Lorre

ivanopagano:
closes #581 

The change split the main indexer code into multiple modules, each handling a specific entity:
* blocks
* accounts
* bakers
* accounts events
* baking/endorsing rights

Additionally, the code to create all necessary resources (db, actor system, ...) and auxiliary processors (the modules listed before) is now separated into the constructor code, to further remove code inside the main indexer body.

### DIFF
--- a/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
+++ b/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
@@ -43,7 +43,7 @@ import scala.util.{Failure, Success, Try}
   * @param accountsProcessor module providing entity-related operations
   * @param bakersProcessor module providing entity-related operations
   * @param rightsProcessor module providing entity-related operations
-  * @param accountsEventsProcessor module handling global events that could trigger global accounts reprocessing
+  * @param accountsResetHandler module handling global events that could trigger global accounts reprocessing
   * @param terminationSequence a function to clean up any pending resource upon shutdown of the indexer
   */
 class TezosIndexer private (
--- a/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
+++ b/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
@@ -28,6 +28,7 @@ import tech.cryptonomic.conseil.indexer.config.{
 import tech.cryptonomic.conseil.indexer.logging.LorreProgressLogging
 import tech.cryptonomic.conseil.indexer.tezos.TezosErrors._
 import tech.cryptonomic.conseil.indexer.tezos.processing._
+import tech.cryptonomic.conseil.indexer.tezos.processing.AccountsResetHandler.{AccountResetEvents, UnhandledResetEvents}
 
 import scala.annotation.tailrec
 import scala.concurrent.duration.{Duration, _}
@@ -53,7 +54,7 @@ class TezosIndexer private (
     accountsProcessor: AccountsProcessor,
     bakersProcessor: BakersProcessor,
     rightsProcessor: BakingAndEndorsingRightsProcessor,
-    accountsEventsProcessor: AccountsEventsProcessor,
+    accountsResetHandler: AccountsResetHandler,
     terminationSequence: () => Future[ShutdownComplete]
 )(
     implicit
@@ -87,18 +88,18 @@ class TezosIndexer private (
   @tailrec
   private def mainLoop(
       iteration: Int,
-      accountsRefreshLevels: AccountsEventsProcessor.AccountUpdatesEvents
+      accountResetEvents: AccountResetEvents
   ): Unit = {
     val noOp = Future.successful(())
     val processing = for {
-      nextRefreshes <- accountsEventsProcessor.processAccountRefreshes(accountsRefreshLevels)
+      unhandled <- accountsResetHandler.applyUnhandledAccountsResets(accountResetEvents)
       _ <- processTezosBlocks()
       _ <- if (iteration % lorreConf.feeUpdateInterval == 0)
         TezosFeeOperations.processTezosAverageFees(lorreConf.numberOfFeesAveraged)
       else
         noOp
       _ <- rightsProcessor.updateRightsTimestamps()
-    } yield Some(nextRefreshes)
+    } yield Some(unhandled)
 
     /* Won't stop Lorre on failure from processing the chain, unless overridden by the environment to halt.
      * Can be used to investigate issues on consistently failing block or account processing.
@@ -110,17 +111,22 @@ class TezosIndexer private (
           //swallow the error and proceed with the default behaviour
           case f @ (AccountsProcessingFailed(_, _) | BlocksProcessingFailed(_, _) | BakersProcessingFailed(_, _)) =>
             logger.error("Failed processing but will keep on going next cycle", f)
-            None //we have no meaningful response to provide
+            None //we have no meaningful data as results, so we return nothing
         } else processing
 
     //if something went wrong and wasn't recovered, this will actually blow the app
-    val updatedLevels = Await.result(attemptedProcessing, atMost = Duration.Inf)
+    val unhandledResetEvents = Await.result(attemptedProcessing, atMost = Duration.Inf) match {
+      case None => //retry processing the same reset events for accounts
+        accountResetEvents
+      case Some(UnhandledResetEvents(events)) =>
+        events
+    }
 
     lorreConf.depth match {
       case Newest =>
         logger.info("Taking a nap")
         Thread.sleep(lorreConf.sleepInterval.toMillis)
-        mainLoop(iteration + 1, updatedLevels.getOrElse(accountsRefreshLevels))
+        mainLoop(iteration + 1, unhandledResetEvents)
       case _ =>
         logger.info("Synchronization is done")
     }
@@ -184,12 +190,12 @@ class TezosIndexer private (
 
   override def start(): Unit = {
     checkTezosConnection()
-    val accountRefreshesToRun =
+    val accountResetsToHandle =
       Await.result(
-        accountsEventsProcessor.unprocessedLevelsForRefreshingAccounts(lorreConf.chainEvents),
+        accountsResetHandler.unprocessedResetRequestLevels(lorreConf.chainEvents),
         atMost = 5.seconds
       )
-    mainLoop(0, accountRefreshesToRun)
+    mainLoop(0, accountResetsToHandle)
   }
 
   override def stop(): Future[ShutdownComplete] = terminationSequence()
@@ -239,8 +245,8 @@ object TezosIndexer extends LazyLogging {
     /* handles standard accounts data */
     val accountsProcessor = new AccountsProcessor(nodeOperator, db, batchingConf, lorreConf.blockRightsFetching)
 
-    /* handles occasional global accounts refreshes due to very specific events */
-    val accountsEventsProcessor = new AccountsEventsProcessor(db, indexedData)
+    /* handles wide-range accounts refresh due to occasional special events */
+    val accountsResetHandler = new AccountsResetHandler(db, indexedData)
 
     /* handles bakers data */
     val bakersProcessor = new BakersProcessor(nodeOperator, db, batchingConf, lorreConf.blockRightsFetching)
@@ -342,7 +348,7 @@ object TezosIndexer extends LazyLogging {
       accountsProcessor,
       bakersProcessor,
       rightsProcessor,
-      accountsEventsProcessor,
+      accountsResetHandler,
       gracefulTermination
     )
   }
--- a/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
+++ b/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
@@ -34,7 +34,17 @@ import scala.concurrent.duration.{Duration, _}
 import scala.concurrent.{Await, ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
-/** Class responsible for indexing data for Tezos BlockChain */
+/** Class responsible for indexing data for Tezos BlockChain
+  * @param ignoreProcessFailures `true` if non-critical errors while fetchign data should simply resume the indexer logic and retry
+  * @param lorreConf keeps necessary configuration values
+  * @param nodeOperator access to the remote node to read data
+  * @param blocksProcessor module providing entity-specific operations
+  * @param accountsProcessor module providing entity-related operations
+  * @param bakersProcessor module providing entity-related operations
+  * @param rightsProcessor module providing entity-related operations
+  * @param accountsEventsProcessor module handling global events that could trigger global accounts reprocessing
+  * @param terminationSequence a function to clean up any pending resource upon shutdown of the indexer
+  */
 class TezosIndexer private (
     ignoreProcessFailures: Boolean,
     lorreConf: LorreConfiguration,
@@ -220,10 +230,10 @@ object TezosIndexer extends LazyLogging {
 
     /* provides operations to handle rights to bake and endorse blocks */
     val rightsProcessor = new BakingAndEndorsingRightsProcessor(
-      db,
-      lorreConf.blockRightsFetching,
       nodeOperator,
-      indexedData
+      indexedData,
+      db,
+      lorreConf.blockRightsFetching
     )
 
     /* handles standard accounts data */
@@ -236,8 +246,27 @@ object TezosIndexer extends LazyLogging {
     val bakersProcessor = new BakersProcessor(nodeOperator, db, batchingConf, lorreConf.blockRightsFetching)
 
     /* Reads csv resources to initialize db tables and smart contracts objects */
-    def initAnyCsvConfig(): TokenContracts = {
+    def parseCSVConfigurations(): TokenContracts = {
+      /* This will derive automatically all needed implicit arguments to convert csv rows into a proper table row
+       * Using generic representations for case classes, provided by the `shapeless` library, it will create the
+       * appropriate `kantan.csv.Decoder` for
+       *  - headers, as extracted by the case class definition which, should match name and order of the csv header row
+       *  - table row types, which can be converted to shapeless HLists, which so happens to be of any case class.
+       *
+       * What shapeless does, is provide an automatic conversion, from a case class to a typed generic list of values,
+       * where each element has a type corresponding to a field in the case class, and a "label" that is
+       * a sort of string extracted at compile-time via macros, to figure out the field names.
+       * Such special list, is a `shapeless.HList`, or heterogeneous list, sort of a dynamic tuple,
+       * to be built by adding/removing individual elements in the list, recursively.
+       * This allows generic libraries like kantan to define codecs for generic HLists and,
+       * using shapeless, to adapt any case class to his specific HList, at compile-time.
+       *
+       * For additional information, refer to the project wiki, and
+       * - http://nrinaudo.github.io/kantan.csv/
+       * - http://www.shapeless.io/
+       */
       import kantan.csv.generic._
+      //we add any missing implicit decoder for column types, not provided by default from the library
       import tech.cryptonomic.conseil.common.util.ConfigUtil.Csv._
 
       /* Inits tables with values from CSV files */
@@ -268,10 +297,10 @@ object TezosIndexer extends LazyLogging {
       tokenContracts
     }
 
-    /* read known token smart contracts from configuration, which represents crypto-currencies internal to the chain
+    /* read known token smart contracts from configuration, which represents crypto-assets internal to the chain
      * along with other static definitions to save in registry tables
      */
-    implicit val tokens: TokenContracts = initAnyCsvConfig()
+    implicit val tokens: TokenContracts = parseCSVConfigurations()
 
     /* This is a smart contract acting as a Naming Service which associates accounts hashes to registered memorable names.
      * It's read from configuration and includes the possibility that none is actually defined
--- a/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
+++ b/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
@@ -9,7 +9,7 @@ import mouse.any._
 import tech.cryptonomic.conseil.common.config.Platforms.{BlockchainPlatform, TezosConfiguration}
 import tech.cryptonomic.conseil.common.config._
 import tech.cryptonomic.conseil.indexer.tezos.michelson.contracts.{TNSContract, TokenContracts}
-import tech.cryptonomic.conseil.common.tezos.{Tables, TezosTypes}
+import tech.cryptonomic.conseil.common.tezos.Tables
 import tech.cryptonomic.conseil.common.tezos.TezosTypes.ContractId
 import tech.cryptonomic.conseil.common.util.DatabaseUtil
 import tech.cryptonomic.conseil.indexer.config.LorreAppConfig.LORRE_FAILURE_IGNORE_VAR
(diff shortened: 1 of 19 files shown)


conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
```diff
@@ -28,166 +27,42 @@ import tech.cryptonomic.conseil.indexer.config.{
...
-    callsConf: NetworkCallsConfiguration,
-    streamingClientConf: HttpStreamingConfiguration,
-    batchingConf: BatchFetchConfiguration
+    nodeOperator: TezosNodeOperator,
+    blocksProcessing: BlocksProcessing,
+    accountsProcessing: AccountsProcessing,
+    bakersProcessing: BakersProcessing,
+    rightsProcessing: BakingAndEndorsingRightsProcessing,
+    accountsEventsProcessing: AccountsEventsProcessing,
+    terminationSequence: () => Future[ShutdownComplete]
```
vishakh, line 51: Why is this a member when it can simply live a function?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
```diff
@@ -518,446 +174,151 @@ class TezosIndexer(
...
+      //return the contracts definitions
+      tokenContracts
     }
-  }
-
-  private object BakersProcessor {
 
-    /* Fetches the data from the chain node and stores bakers into the data store.
-     * @param ids a pre-filtered map of delegate hashes with latest block referring to them
+    /* read known token smart contracts from configuration, which represents crypto-currencies internal to the chain
```
vishakh, line 275: "crypto-currencies" => "crypto-assets"

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
```diff
@@ -518,446 +174,151 @@ class TezosIndexer(
...
+    /* handles bakers data */
+    val bakersProcessing = new BakersProcessing(nodeOperator, db, batchingConf, lorreConf.blockRightsFetching)
+
+    /* Reads csv resources to initialize db tables and smart contracts objects */
+    def initAnyCsvConfig(): TokenContracts = {
+      import kantan.csv.generic._
+      import tech.cryptonomic.conseil.common.util.ConfigUtil.Csv._
+
+      /* Inits tables with values from CSV files */
+      TezosDb.initTableFromCsv(db, Tables.KnownAddresses, selectedNetwork)
```
vishakh, line 248: Our of curiosity, where is the implicit `HeaderDecoder` here coming from? Maybe it's possible to leave a comment?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
```diff
@@ -518,446 +174,151 @@ class TezosIndexer(
...
+    val accountsProcessing = new AccountsProcessing(nodeOperator, db, batchingConf, lorreConf.blockRightsFetching)
+
+    /* handles occasional global accounts refreshes due to very specific events */
+    val accountsEventsProcessing = new AccountsEventsProcessing(db, indexedData)
+
+    /* handles bakers data */
+    val bakersProcessing = new BakersProcessing(nodeOperator, db, batchingConf, lorreConf.blockRightsFetching)
+
+    /* Reads csv resources to initialize db tables and smart contracts objects */
+    def initAnyCsvConfig(): TokenContracts = {
```
vishakh, line 243: I feel this name is too vague. The function does not init *any* CSV config. It inits specific CSV configs.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
```diff
@@ -28,166 +27,42 @@ import tech.cryptonomic.conseil.indexer.config.{
...
 import scala.util.{Failure, Success, Try}
+import tech.cryptonomic.conseil.indexer.tezos.processing.BakingAndEndorsingRightsProcessing
+import tech.cryptonomic.conseil.indexer.tezos.processing.AccountsEventsProcessing
+import tech.cryptonomic.conseil.indexer.tezos.processing.BakersProcessing
+import tech.cryptonomic.conseil.indexer.tezos.processing.AccountsProcessing
+import tech.cryptonomic.conseil.indexer.tezos.processing.BlocksProcessing
 
 /** * Class responsible for indexing data for Tezos BlockChain */
-class TezosIndexer(
+class TezosIndexer private (
```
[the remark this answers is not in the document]
vishakh: BTW, I am fine with keeping it stateful if you think the functions will become too clumsy otherwise. However, I do think there's a elegance to piping the same state through multiple function compositions, with each function acting on the result of the previous computation. Maybe something like a State monad?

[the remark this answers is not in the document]
vishakh: I can't help but feel there is a good opportunity to describe the Lorre logic as a computation pipeline and have a single for-expression that captures the essence of that pipeline.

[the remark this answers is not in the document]
vishakh: I also feel the logic of each Lorre iteration can be captured as a very descriptive for-expression, something like:

```scala
for {
blocks <- fetchBlocks()
ops <- fetchOperations()
accounts <- fetchAccounts()
...
_ <- writeToDatabase(blocks, ops, accounts, ..)
}(yield ???)
```

I realize this is a simple refactor PR so, if my comments make any sense, I recommend only following them it it's not more than a couple of days' effort.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BlocksProcessing.scala
```diff
@@ -0,0 +1,80 @@
...
+    extends LazyLogging {
+
+  /* will store a single page of block results */
+  private[tezos] def processBlocksPage(results: nodeOperator.BlockFetchingResults)(
+      implicit ec: ExecutionContext
+  ): Future[Int] = {
+    def logBlockOutcome[A]: PartialFunction[Try[Option[A]], Unit] = {
+      case Success(accountsCount) =>
+        logger.info(
+          "Wrote {} blocks to the database, checkpoint stored for{} account updates",
```
ajuszczak, line 35: Small thing, missing whitespace between `for` and `{}`

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BlocksProcessing.scala
```diff
@@ -0,0 +1,80 @@
...
+          results.size,
+          accountsCount.fold("")(" " + _)
```
ajuszczak, line 37: `accountsCount` suggests we are talking about `Numeric`, but we are folding them to a bit different format. What would be an example outcome from this expression? Not sure how it would fit to the logging message.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BlocksProcessing.scala
```diff
@@ -0,0 +1,80 @@
...
+  /** Prepares and stores statistics for voting periods of the
+    * blocks passed in.
+    *
+    * @param bakerRollsByBlock blocks of interest, with any rolls data available
+    * @return the outcome of the operation, which may fail with an error or produce no result value
+    */
+  private def processBlocksForGovernance(bakerRollsByBlock: Map[Block, List[Voting.BakerRolls]])(
+      implicit ec: ExecutionContext
+  ): Future[Unit] =
+    for {
```
ajuszczak, line 73: Looks like this for-comprehension can be easily replace with `flatMap`, but this probably my personal preferences, so you can ignore it, if you don't like it :)

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BakingAndEndorsingRightsProcessing.scala
```diff
@@ -0,0 +1,158 @@
...
+    import cats.implicits._
+
+    berLogger.info("Fetching future baking and endorsing rights")
+    val blockHead = nodeOperator.getBareBlockHead()
+    val brLevelFut = indexedData.fetchMaxBakingRightsLevel()
+    val erLevelFut = indexedData.fetchMaxEndorsingRightsLevel()
+
+    (blockHead, brLevelFut, erLevelFut).mapN { (head, brLevel, erLevel) =>
+      val headLevel = head.header.level
+      val rightsStartLevel = math.max(brLevel, erLevel) + 1
```
ajuszczak, line 98: +1 always looks suspicious to me :D

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BakersProcessing.scala
```diff
@@ -0,0 +1,246 @@
...
+                val rollsSum = rolls.filter(_.pkh.value == key.value).map(_.rolls).sum
+                (key, baker.updateRolls(rollsSum))
+            })
+        )
+
+      db.run(TezosDb.writeBakersAndCopyContracts(enrichedBakers.toList))
+        .andThen(logWriteFailure)
+    }
+
+    def cleanup[T] = (_: T) => {
```
ajuszczak, line 84: This signature looks weird, do we need to have this `(_: T)` parameter?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/AccountsEventsProcessing.scala
```diff
@@ -0,0 +1,112 @@
...
+          }
+          SortedSet(unprocessedEvents: _*)
+        }
+    }.getOrElse(Future.successful(SortedSet.empty))
+
+  /* Possibly updates all accounts if the current block level is past any of the given ones
+   *
+   * @param events the relevant levels, each with its own selection pattern, that calls for a refresh
+   */
+  private[tezos] def processAccountRefreshes(events: AccountUpdatesEvents)(
```
ajuszczak, line 58: Hm... maybe it's worth to return different type, at the end, because when I first time checked this method, for me (from the signature perspective), it was doing nothing, but comment says something different.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/AccountsProcessing.scala
```diff
@@ -0,0 +1,301 @@
...
+
+    def logWriteFailure: PartialFunction[Try[_], Unit] = {
+      case Failure(e) =>
+        logger.error("Could not write accounts to the database")
+    }
+
+    def logOutcome: PartialFunction[Try[(Option[Int], Option[Int], _)], Unit] = {
+      case Success((accountsRows, delegateCheckpointRows, _)) =>
+        logger.info(
+          "{} accounts were touched on the database. Checkpoint stored for{} bakers.",
```
ajuszczak, line 70: Log message can be improved (missing whitespace)

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/AccountsProcessing.scala
```diff
@@ -0,0 +1,301 @@
...
+          accountsRows.fold("The")(String.valueOf),
```
ajuszczak, line 71: Are you sure, that whitespace is not needed here?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/AccountsProcessing.scala
```diff
@@ -0,0 +1,301 @@
...
+      db.run {
+        DBIO.sequence {
+          levels.map { level =>
+            TezosDb.fetchRecentOperationsHashByKind(Set("activate_account"), level).map(x => level -> x)
+          }
+        }.map(_.toMap)
+      }
+    }
+
+    def cleanup[T] = (_: T) => {
```
ajuszczak, line 173: Do we need this `(_: T)` here?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BlocksProcessing.scala
ivanopagano: Assume results.size = 10
1. accountsCount = None =>
  Wrote 10 blocks to the database, checkpoint stored for account updates
2. accountsCount = Some(2)
  Wrote 10 blocks to the database, checkpoint stored for 2 account updates

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BlocksProcessing.scala
ivanopagano: It's an optional number, and I need a result that will be empty as a string when the value is not defined.
Folding provides conversion of the number to a string (with the leading space), with a default of empty string.
See the previous comment for the example output

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BlocksProcessing.scala
ivanopagano: You're right, it's the result of pruning a longer expression, and I missed the opportunity

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/AccountsEventsProcessing.scala
ivanopagano: AccountUpdatesEvents is an alias to a `Set`.

Why would you expect that a method Set => Set would do nothing?
Even without considering side-effects... the signature being Set => F[Set]

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BakingAndEndorsingRightsProcessing.scala
ivanopagano: Will check once more, I think there's nothing wrong here though

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BakersProcessing.scala
ivanopagano: Not necessarily? I'll check it out, thanks

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/AccountsProcessing.scala
ivanopagano: I'll check it out, thanks

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/AccountsProcessing.scala
ivanopagano: Check the other similar comments to get the expected result

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/AccountsProcessing.scala
ivanopagano: Look at the previous reply

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
```diff
@@ -28,166 +27,42 @@ import tech.cryptonomic.conseil.indexer.config.{
...
-    tezosConf: TezosConfiguration,
-    callsConf: NetworkCallsConfiguration,
-    streamingClientConf: HttpStreamingConfiguration,
-    batchingConf: BatchFetchConfiguration
+    nodeOperator: TezosNodeOperator,
+    blocksProcessing: BlocksProcessing,
+    accountsProcessing: AccountsProcessing,
+    bakersProcessing: BakersProcessing,
+    rightsProcessing: BakingAndEndorsingRightsProcessing,
+    accountsEventsProcessing: AccountsEventsProcessing,
```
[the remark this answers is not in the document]
ivanopagano: yup, no problem

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
ivanopagano: It's an input parameter whose actual call argument will need to close over local variables (essentially the resources to be freed by the call)
Those variables can't be closed over by a separate function.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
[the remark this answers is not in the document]
ivanopagano: Makes sense overall, yet I'd need to think over it first a little more.
At first look I'd say that the indexer is indeed stateless, all it needs are input parameters representing configuration and unavoidable external resource (db, node calls, ...).
I don't know how this fits with your comment, please elaborate more if you will.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BlocksProcessing.scala
ajuszczak: Tricky, now I see :)

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BlocksProcessing.scala
ajuszczak: Yea, related to above. Now I understand that.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/AccountsEventsProcessing.scala
ajuszczak: I explain that wrongly. What I had in mind is that, it's quiet easy to forget to fire this method up. When you have "chain" of types (so next method accepts type from previously returned) it's harder (or even impossible) to forget some method. Even during refactoring, moving code, copying it etc. When you have 2-3 methods with Set -> Set, does the order or running them matter? I hope not, but it easier to mess this up than, when you have proper types (not even aliases, because for the point mentioned above) it's even worst to me. I thought this one is a case class, but probably I should find out this one is an alias, by looking at the methods used on it.

By "would do nothing" I expect that whatever this method is doing for the same in&out types, there is hidden condition inside the method which sometimes does something with the data, sometimes does not. Like `list.filter`, `list.collect`, there are condition defined while executing these methods. When we are dealing with case classes, you also don't want to have `case class Account(name: Option[String] = None)` and inside of some methods do `account.copy(name = "account" + i)` cause that creates more confusion than helps, thus different type would be helpful here. Otherwise, we can create case class with optionals only and be happy about that, but as you know it's not the best pattern.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BakingAndEndorsingRightsProcessing.scala
ajuszczak: Suspicious in a way, that sometimes it's hard do guess why you want to do +1. It can be totally valid tho.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
```diff
@@ -28,166 +27,42 @@ import tech.cryptonomic.conseil.indexer.config.{
...
+import tech.cryptonomic.conseil.indexer.tezos.processing.BakingAndEndorsingRightsProcessing
+import tech.cryptonomic.conseil.indexer.tezos.processing.AccountsEventsProcessing
+import tech.cryptonomic.conseil.indexer.tezos.processing.BakersProcessing
+import tech.cryptonomic.conseil.indexer.tezos.processing.AccountsProcessing
+import tech.cryptonomic.conseil.indexer.tezos.processing.BlocksProcessing
 
 /** * Class responsible for indexing data for Tezos BlockChain */
-class TezosIndexer(
+class TezosIndexer private (
+    ignoreProcessFailures: Boolean,
```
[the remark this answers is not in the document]
ivanopagano: done

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BlocksProcessing.scala
```diff
@@ -0,0 +1,80 @@
...
+  TezosNodeOperator,
+  TezosGovernanceOperations,
+  TezosDatabaseOperations => TezosDb
+}
+import tech.cryptonomic.conseil.common.tezos.TezosTypes.{Block, Voting}
+import tech.cryptonomic.conseil.common.tezos.TezosTypes.Syntax._
+import tech.cryptonomic.conseil.indexer.tezos.michelson.contracts.TokenContracts
+import tech.cryptonomic.conseil.indexer.tezos.michelson.contracts.TNSContract
+
+class BlocksProcessing(
```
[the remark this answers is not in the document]
ivanopagano: done

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/AccountsProcessing.scala
```diff
@@ -0,0 +1,301 @@
...
+  PublicKeyHash,
+  Voting
+}
+import tech.cryptonomic.conseil.indexer.tezos.TezosErrors.AccountsProcessingFailed
+import tech.cryptonomic.conseil.common.tezos.TezosTypes.Syntax._
+
+import com.typesafe.scalalogging.LazyLogging
+import slick.jdbc.PostgresProfile.api._
+
+class AccountsProcessing(
```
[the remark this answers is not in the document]
ivanopagano: done

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BakersProcessing.scala
```diff
@@ -0,0 +1,246 @@
...
+import slick.jdbc.PostgresProfile.api._
+import akka.stream.ActorMaterializer
+
+/** Takes care of fetching and processing any bakers' related data.
+  *
+  * @param nodeOperator
+  * @param db
+  * @param batchingConf
+  */
+class BakersProcessing(
```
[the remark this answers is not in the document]
ivanopagano: done

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BakingAndEndorsingRightsProcessing.scala
```diff
@@ -0,0 +1,158 @@
...
+
+import org.slf4j.LoggerFactory
+import scala.concurrent.ExecutionContext
+import akka.stream.scaladsl.Source
+import akka.stream.scaladsl.Sink
+import akka.stream.ActorMaterializer
+import slick.jdbc.PostgresProfile.api._
+import scala.concurrent.Future
+
+class BakingAndEndorsingRightsProcessing(
```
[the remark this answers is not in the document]
ivanopagano: done

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
ivanopagano: done

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
ivanopagano: ok, ok, I'll find a more fitting name

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosIndexer.scala
ivanopagano: will do

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BakersProcessor.scala
```diff
@@ -0,0 +1,247 @@
...
+import tech.cryptonomic.conseil.indexer.tezos.TezosErrors.BakersProcessingFailed
+import com.typesafe.scalalogging.LazyLogging
+import slick.jdbc.PostgresProfile.api._
+import akka.stream.ActorMaterializer
+
+/** Takes care of fetching and processing any bakers' related data.
+  *
+  * @param nodeOperator access to the remote node to read data
+  * @param db raw access to the slick database
+  * @param batchingConf defines values for batching of data processing
```
piotrkosecki, line 31: minor, missing `rightsConf` docs

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/AccountsEventsProcessing.scala
ivanopagano: I see now what you mean, specific types to address different states of a pipelined flow.
I'm not totally sold here as, foremost 'cause we're not deeply invested into types for reliability of the code. There will be gains in type-safety, but there's quite the cost in using types for everything.
I've found situations in this project where I could've done it, but I usually reject it because the additional complexity of the code needed wasn't worthwhile, all things considered.
I've usually made use of aliases and such, primarily to highlight the semantic of the logic, and to reduce code clutter, hence improving the reader experience and understanding. Obviously, I've found I have missed this goal not infrequently.

That said, I'll think about that part and try to see if something fits that would avoid the risks you outlined. I'm not promising I'll choose to do it, in the end.

ivanopagano: I'm mostly concerned by the fact that this set is passed recursively to the same method, hence I'm not so sure if I can really improve the situation.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/AccountsEventsProcessing.scala
```diff
@@ -0,0 +1,112 @@
...
+import com.typesafe.scalalogging.LazyLogging
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
+import scala.collection.immutable.SortedSet
+import tech.cryptonomic.conseil.common.config.ChainEvent
+import tech.cryptonomic.conseil.common.tezos.TezosTypes.BlockHash
+import tech.cryptonomic.conseil.indexer.tezos.{TezosIndexedDataOperations, TezosDatabaseOperations => TezosDb}
+import slick.jdbc.PostgresProfile.api._
+
+object AccountsEventsProcessing {
```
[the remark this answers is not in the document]
ivanopagano: renamed

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/AccountsEventsProcessing.scala
ivanopagano: Ok I thing I finally gotcha here @ajuszczak

ajuszczak: Sure, what you are saying with complexity also make sense to me. There is always a tradeoff in such cases.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BakingAndEndorsingRightsProcessing.scala
ivanopagano: I still have to check this

PR #861, opened 2020-06-26
Fix issue with tables leak from tezos schema in bitcoin's while prefetching metadata

ajuszczak:
This pull-request solves an issue at startup, where Conseil API has been reporting an issue with fetching metadata. An error has been noticed, while developing new integrations and results into stack-trace at startup and overall failure on metadata prefetching.

### DIFF
--- a/conseil-api/src/main/scala/tech/cryptonomic/conseil/api/metadata/AttributeValuesCacheConfiguration.scala
+++ b/conseil-api/src/main/scala/tech/cryptonomic/conseil/api/metadata/AttributeValuesCacheConfiguration.scala
@@ -1,7 +1,7 @@
 package tech.cryptonomic.conseil.api.metadata
 
 import tech.cryptonomic.conseil.common.config.MetadataConfiguration
-import tech.cryptonomic.conseil.common.config.Types.{AttributeName, EntityName}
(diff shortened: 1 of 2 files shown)


conseil-api/src/main/scala/tech/cryptonomic/conseil/api/metadata/AttributeValuesCacheConfiguration.scala
```diff
@@ -29,12 +29,12 @@ class AttributeValuesCacheConfiguration(metadataConfiguration: MetadataConfigura
   private def whenVisible[T](path: Path)(value: => Option[T]): Option[T] =
     when(metadataConfiguration.isVisible(path))(value).flatten
 
-  /** extracts pair (entity, attribute) which needs to be cached */
-  def getAttributesToCache: List[(EntityName, AttributeName)] =
+  /** extracts pair (platform, entity, attribute) which needs to be cached */
```
piotrkosecki, line 32: it's no longer a pair, but tuple😉

ajuszczak:
> Looks fine, care to add some docs on the methods, since we're here? 😁

Should be better now. I've added some docs and extended existing one :)

PR #867, opened 2020-06-29
868-4 - Lorre for Bitcoin

olownia:


conseil-common/src/main/resources/reference.conf
```diff
@@ -25,6 +25,26 @@ platforms {
...
   }
+
+  bitcoin {
+    mainnet {
+      node {
+        protocol: "http"
+        hostname: "localhost",
+        port: 8332,
+        username: "bitcoin",
+        password: "sP9PV88xtbGwLMAEi2rVlZ7jIFfJbpOmTUCsBBBRN9I="
```
ivanopagano, line 36: Is this somewhat sensitive information to be shared?

olownia: no, this is a password used in docker-compose

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinIndexer.scala
```diff
@@ -0,0 +1,128 @@
...
+
+      bitcoinOperations <- BitcoinOperations.resource(rpcClient, tx, bitcoinConf.batchingConf)
+    } yield bitcoinOperations
+
+  /**
+    * The method with all the computations for the Bitcoin. 
+    * Currently, it only contains the blocks. But it can be extended to 
+    * handle multiple computations.
+    */
+  def mainLoop(bitcoinOperations: BitcoinOperations[IO]): IO[Unit] =
```
ivanopagano, line 94: What about a helper function like

`def repeatAfter[A](interval: FiniteDuration)(code: IO[A]): IO[Unit]`

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperations.scala
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+    * Get Blocks from Bitcoin node through Bitcoin client and save them into the database using Slick.
+    * In the beginning, the current list of blocks is obtained from the database and removed from the computation.
+    *
+    * @param range Inclusive range of the block's height
+    */
+  def loadBlocksWithTransactions(range: Range.Inclusive): Stream[F, Unit] =
+    Stream
+      .eval(getExistingBlocks(range))
```
olownia, line 56: @ivanopagano 's comment:
> Isn't this quite expensive as things grow? Any data on the sql operational metrics?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinIndexer.scala
olownia: `mainLoop` is a very unpleasant name for the Scala project :)

ivanopagano: Any suggestions? We could change it

conseil-common/src/main/resources/reference.conf
ivanopagano: where does it come from?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperations.scala
```diff
@@ -0,0 +1,100 @@
...
+import tech.cryptonomic.conseil.indexer.config.{Custom, Depth, Everything, Newest}
+
+/**
+  * Bitcoin operations for Lorre.
+  *
+  * @param bitcoinClient Bitcoin client instance
+  * @param persistence Bitcoin persistence instance
+  * @param tx [[slickeffect.Transactor]] to perform a Slick operations on the database
+  */
+class BitcoinOperations[F[_]: Concurrent](
```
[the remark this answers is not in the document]
piotrkosecki: @ivanopagano what would be an advantage of that?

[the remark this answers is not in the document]
ivanopagano: What's the advantage of doing it generally, to be honest?

I've trodden this road before and I have the feeling that the constraint get so strict at a certain point, that keeping the abstraction makes sense only for the basic building blocks, but not at the top.
After all, once you get Concurrent/Effect/ConcurrentEffect, how would you think to implement that?

I mean, are you thinking of something different from IO, eventually?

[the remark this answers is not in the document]
piotrkosecki: I just asked because for me it makes no difference. It's like comparing colours of the same model of a car - engine is the same, but some people like it blue, and some people like it red.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinIndexer.scala
```diff
@@ -0,0 +1,128 @@
...
+  /**
+    * Dedicated executor for the http4s.
+    */
+  private val httpExecutor = Executors.newFixedThreadPool(bitcoinConf.batchingConf.httpFetchThreadsCount)
+
+  /**
+    * [[cats.ContextShift]] is the equivalent to [[ExecutionContext]], 
+    * it's used by the Cats Effect related methods.
+    */
+  implicit val contextShift = IO.contextShift(ExecutionContext.fromExecutor(indexerExecutor))
```
ajuszczak, line 51: Can we make private, what should be private? I guess here, there is only one method that should be public (or two, don't remember). I really like to keep it like that, because that can speed up future refactoring and tells instantly which method is actually used or not. Let's think about classes, changes in scopes, not in modules :)

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinIndexer.scala
ajuszczak: Maybe `runContinuously`, `runInfinitely` (or simply `infinitely`), we can always name it `run`, but that is too generic for me, but I also like Ivano's idea, about extracting interval and simply loop that with this parameter as well. 

Note that, in that case we could easily change the strategy for retries on errors.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperations.scala
```diff
@@ -0,0 +1,100 @@
...
+) extends LazyLogging {
+
+  /**
+    * Start Lorre with mode defined with [[Depth]].
+    *
+    * @param depth Can be: Newest, Everything or Custom
+    */
+  def loadBlocks(depth: Depth): Stream[F, Unit] =
+    Stream
+      .eval(tx.transact(Tables.Blocks.sortBy(_.height).take(1).result)) // get last block height
```
ajuszczak, line 36: Could we extract this operation to separate method? That might improve visibility. It sounds like "heavy" method and I think it's better to extract this than to make hide it in the middle. Although it will do exactly the same, we could also test this out, while having in separate method (just to test this specific method).

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperations.scala
```diff
@@ -0,0 +1,100 @@
...
+            .drain
+      )
+
+  /**
+    * Get sequence of existing blocks from the database.
+    *
+    * @param range Inclusive range of the block's height
+    * @return
+    */
+  def getExistingBlocks(range: Range.Inclusive): F[Seq[Int]] =
```
ajuszczak, line 77: I believe we need this method (instead of fetching the latest) due to concurrent nature of RPC Client? (Just to confirm). Asking, cause this list might be pretty long, one we will have million of blocks, right? Which specific implementation is hidden under `Seq` here - it will make a difference in long run I guess.

docker-compose.yml
```diff
@@ -1,6 +1,15 @@
 version: "3.1"
 
 services:
+  # Bitcoin node for development, if you want to start your own node, 
+  # uncomment below lines:
+  # bitcoin:
```
ajuszczak, line 6: Let's name it `bitcoin-node` and also I'm not sure if that should be commented out. I think you did that mostly, because someone can accidentally run node, while typing `docker-compose up` correct?

ajuszczak:
I see -3% in tests coverage. Is there something we can do about that? I mean, it is quite significant. We should rather go up than down here :D

conseil-common/src/main/resources/reference.conf
olownia: I've generated it. Should I change it to "password"?

docker-compose.yml
olownia: exactly, I'm wondering if it should be a part of the docker-compose, maybe I should create a wiki page about local Bitcoin development and move this code there?

ajuszczak: or, move that to completely new docker-compose file. Like `docker-compose.local.yml` and put it there. I believe to execute that, you need to run `docker-compose -f <filename>`. @ivanopagano @piotrkosecki what do you think about that?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinIndexer.scala
olownia: check it now, I've refactored `start` method a bit

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinIndexer.scala
olownia: Please review the code again

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperations.scala
olownia: Done

docker-compose.yml
piotrkosecki: this is the `local` docker-compose file, as the one located in the`./docker` directory has full integration for the current conseil/lorre/postgres. I would vote to keep it here. Just change the wiki to mention current development containers.

conseil-common/src/main/resources/reference.conf
piotrkosecki: I would go for something simple if it's just for dev purposes

ivanopagano: I mean, as long as it's nothing referring to a real working infrastructure it's fine. It looked like a real password, so I was worried about security holes, nothing more...

docker-compose.yml
ivanopagano: I agree with @piotrkosecki , every dev can comment out whatever is needed.

Personally I'd comment out the less needed components, like graphql and bitcoin, and commit it like that.
For me it's the most ergonomic solution for out-of-the-box development. YMMV

build.sbt
```diff
@@ -87,7 +87,7 @@ lazy val lorre = (project in file("conseil-lorre"))
     javaExtras = Seq("-Xmx512M", "-Xss1M", "-XX:+CMSClassUnloadingEnabled")
   )
   .enableAssembly()
-  .dependsOn(common, commonTestKit % Test)
+  .dependsOn(common % "compile->compile;test->test", commonTestKit % Test)
```
ivanopagano, line 90: Can you shortly explain what didn't work? Thanks

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinIndexer.scala
```diff
@@ -0,0 +1,146 @@
...
+  /**
+    * [[cats.ContextShift]] is the equivalent to [[ExecutionContext]],
+    * it's used by the Cats Effect related methods.
+    */
+  implicit private val contextShift = IO.contextShift(ExecutionContext.fromExecutor(indexerExecutor))
+
+  /**
+    * [[ExecutionContext]] to handle stop method.
+    */
+  implicit private val indexerEC = ExecutionContext.fromExecutor(indexerExecutor)
```
ivanopagano, line 57: A couple of questions

Is this implicitly required by `stop`? maybe we could pass it explicitly, or use the global executor there. I'm thinking if we could remove from the scope and the code an extra value which is needed only once in a specific place...

What happens when the thread running on an executor invokes his own executor's pool shutdown?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinIndexer.scala
```diff
@@ -0,0 +1,146 @@
...
+  //       to stop the mainLoop.
+  override def start(): Unit = {
+
+    /**
+      * Repeat [[cats.IO]] after the specified interval.
+      *
+      * @param interval finite duration interval
+      * @param f [[cats.IO]] to repeat
+      */
+    def repeatAfter[A](interval: FiniteDuration)(f: IO[A]): IO[Unit] =
```
ivanopagano, line 81: Sorry, my bad, but maybe `repeatEvery` is more correct?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinIndexer.scala
```diff
@@ -0,0 +1,146 @@
...
+    lorreConf: LorreConfiguration,
+    bitcoinConf: BitcoinConfiguration
+) extends LazyLogging
+    with LorreIndexer
+    with LorreProgressLogging {
+
+  /**
+    * Executor for the rpc client, timer and to handle stop method.
+    */
+  private val indexerExecutor = Executors.newFixedThreadPool(bitcoinConf.batchingConf.indexerThreadsCount)
```
ivanopagano, line 41: would it improve the overall design if these contexts were built within a Resource and composed with the indexer resource to automatically handle release and remove any need to put code in the `stop` method?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperations.scala
ivanopagano: Can it be renamed to `getIndexedBlockHeights`

To differentiate from "remote-fetching" and to make clear what is the return type?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperations.scala
```diff
@@ -0,0 +1,107 @@
...
+    tx.transact(
+      Tables.Blocks.filter(_.height between(range.start, range.end)).map(_.height).result
+    )
+
+  /**
+    * Get the last block from the database.
+    */
+  def getLastBlock: F[Option[Tables.BlocksRow]] =
```
ivanopagano, line 84: Can it be renamed to `getHighestIndexedBlock` or `getLatestIndexedBlock` 
for similar reasons as the previous comment.

Also it would be name-wise more similar to existing method names in the project AFAIR

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperationsTest.scala
```diff
@@ -0,0 +1,87 @@
...
+    with InMemoryDatabase
+    with BitcoinInMemoryDatabaseSetup
+    with BitcoinFixtures
+    with BitcoinStubs {
+
+  implicit val contextShift: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
+
+  "Bitcoin operations" should {
+      "return the latest block" in new BitcoinPersistenceStubs(dbHandler) with BitcoinOperationsStubs {
+        (txMock.transact[Option[Tables.BlocksRow]] _).expects(*).onCall { dbio: DBIO[Option[Tables.BlocksRow]] =>
```
ivanopagano, line 36: Sorry but I think I've lost track of what's the mock doing here.

Where is the mocking transactor defined and used, and isn't it pointless here if what it does is to actually call the real transactor?
Thanks for any help here.

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperationsTest.scala
```diff
@@ -0,0 +1,87 @@
...
+          tx.transact(dbio)
+        }
+
+        (for {
+          _ <- tx.transact(
+            Tables.Blocks += RpcFixtures.blockResult.copy(hash = "hash1", height = 1).convertTo[Tables.BlocksRow]
```
ivanopagano, line 42: Might turn that into a single call to store

```scala
Tables.Blocks ++= List(RpcFixtures.blockResult.copy(hash = "hash1", height = 1), RpcFixtures.blockResult.copy(hash = "hash2", height = 2)).map(_.convertTo...)
```

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinIndexer.scala
olownia: I've changed code to use implicitly `ExecutionContext.global` in `stop`

build.sbt
olownia: without it I can't use `DbFixtures` from common in Lorre tests

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinIndexer.scala
olownia: There is some inception, `ContextShift` has to be defined at the beginning, So I'm not able to wrap it into `Resource`. I'm afraid it must stay as it is :/

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperationsTest.scala
olownia: Sorry for that strange part :) I've moved this code to the `BitcoinPersistence`

build.sbt
ajuszczak, line 90: This `test -> test` is what we wanted to avoid by introducing `commonTestKit` and that's the reason of these duplications. It complicates things actually, and I believe I had a talk with @ivanopagano about that. Mostly because (I don't remember exactly) but compile may trigger testing? We don't want to have it.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperations.scala
```diff
@@ -0,0 +1,89 @@
...
+  /**
+    * Start Lorre with mode defined with [[Depth]].
+    *
+    * @param depth Can be: Newest, Everything or Custom
+    */
+  def loadBlocks(depth: Depth): Stream[F, Unit] =
+    Stream
+      .eval(tx.transact(persistence.getLatestIndexedBlock))
+      .zip(bitcoinClient.getBlockChainInfo)
+      .evalTap(_ => Concurrent[F].delay(logger.info(s"Start Lorre for Bitcoin")))
```
ajuszczak, line 37: Is this log done in the right place? I mean there is a place https://github.com/Cryptonomic/Conseil/pull/867/files#diff-03717b0004757d8738510e597ecaafeaR91 where I guess we should print it. Here, we are already indexing blocks (accounts will be probably in a different method), so we are not starting Lorre here. We are starting fetching blocks.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperations.scala
```diff
@@ -0,0 +1,89 @@
...
+      .eval(tx.transact(persistence.getExistingBlocks(range)))
+      .flatMap(
+        existingBlocks =>
+          Stream
+            .range(range.start, range.end)
+            .filter(height => !existingBlocks.contains(height))
+            .through(bitcoinClient.getBlockHash(batchConf.hashBatchSize))
+            .through(bitcoinClient.getBlockByHash(batchConf.blocksBatchSize))
+            .through(bitcoinClient.getBlockWithTransactions(batchConf.transactionsBatchSize))
+            .evalTap { case (block, _) => Concurrent[F].delay(logger.debug(s"Save block with height: ${block.height}")) }
```
ajuszczak, line 64: What about logging every X block? I believe we talked about that. Current solution will create to much noice and we can miss something.

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperationsTest.scala
```diff
@@ -0,0 +1,74 @@
...
+    with Matchers
+    with InMemoryDatabase
+    with BitcoinInMemoryDatabaseSetup
+    with BitcoinFixtures
+    with BitcoinStubs {
+
+  implicit val contextShift: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
+
+  "Bitcoin operations" should {
+      // "..." in new BitcoinPersistenceStubs(dbHandler) with BitcoinOperationsStubs {
```
ajuszczak, line 35: Are we going to test something here?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperations.scala
olownia: This is a debug log message. It isn't the default behavior

build.sbt
olownia: So where should I put the `BitcoinFixtures` to be available in common and Lorre? (`commonTestKit` doesn't depend on common so I can't use the classes I need)

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperationsTest.scala
olownia: yeap

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperations.scala
ajuszczak: Ah, you are right! Question, do we have any feedback from Lorre, while blocks are indexed? I'd like to avoid empty console, while process is working properly, maybe we could still have some feedback, but not that frequent like with every saved block?

olownia: ok :) I will add something like that

build.sbt
ivanopagano: for the moment you might need to put them in `common`?

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperationsTest.scala
```diff
@@ -32,8 +31,70 @@ class BitcoinOperationsTest
...
+              Some(DbFixtures.blockRow.copy(height = 5))
+            ))
+
+        // mock Bitcoin client calls
+        (bitcoinClientMock.getBlockChainInfo _) expects () returning (Stream(BlockchainInfo("mainnet", 10)).covary[IO])
+
+        // test if method runs indexer with the correct range of the blocks
+        (dummyBitcoinOperationsMock.loadBlocksWithTransactions _) expects (6 to 10) returning (Stream.empty)
+
+        dummyBitcoinOperationsStub.loadBlocks(Newest).compile.drain.unsafeRunSync() shouldBe (())
```
ivanopagano, line 47: Is there any real need to compare the unit result to something?

Am I right in my understanding that the test is supposed to verify what internals are being called?

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperationsTest.scala
```diff
@@ -32,8 +31,70 @@ class BitcoinOperationsTest
...
+        (bitcoinPersistenceMock.getLatestIndexedBlock _) expects ()
+        (txMock.transact[Option[Tables.BlocksRow]] _) expects (*) returning (IO(None))
+
+        // mock Bitcoin client calls
+        (bitcoinClientMock.getBlockChainInfo _) expects () returning (Stream(BlockchainInfo("mainnet", 10)).covary[IO])
+
+        // test if method runs indexer with the correct range of the blocks
+        (dummyBitcoinOperationsMock.loadBlocksWithTransactions _) expects (1 to 10) returning (Stream.empty)
+
+        dummyBitcoinOperationsStub.loadBlocks(Everything).compile.drain.unsafeRunSync() shouldBe (())
```
ivanopagano, line 61: same as before

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperationsTest.scala
```diff
@@ -32,8 +31,70 @@ class BitcoinOperationsTest
...
+              Some(DbFixtures.blockRow.copy(height = 3))
+            ))
+
+        // mock Bitcoin client calls
+        (bitcoinClientMock.getBlockChainInfo _) expects () returning (Stream(BlockchainInfo("mainnet", 10)).covary[IO])
+
+        // test if method runs indexer with the correct range of the blocks
+        (dummyBitcoinOperationsMock.loadBlocksWithTransactions _) expects (7 to 10) returning (Stream.empty)
+
+        dummyBitcoinOperationsStub.loadBlocks(Custom(3)).compile.drain.unsafeRunSync() shouldBe (())
```
ivanopagano, line 77: same as before

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperationsTest.scala
```diff
@@ -32,8 +31,70 @@ class BitcoinOperationsTest
...
+        (bitcoinClientMock.getBlockByHash _) expects (1) returning (_ => Stream(RpcFixtures.blockResult).covary[IO])
+        (bitcoinClientMock.getBlockWithTransactions _) expects (1) returning (
+                _ => Stream((RpcFixtures.blockResult, List(RpcFixtures.transactionResult))).covary[IO]
+            )
+
+        // check if the method saves data to the database
+        (bitcoinPersistenceMock.createBlock _) expects (*, *)
+        (txMock.transact[Unit] _) expects (*) returning (IO.unit)
+
+        bitcoinOperationsStub.loadBlocksWithTransactions(1 to 10).compile.drain.unsafeRunSync() shouldBe (())
```
ivanopagano, line 96: same as before

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperationsTest.scala
```diff
@@ -49,26 +110,51 @@ class BitcoinOperationsTest
...
-          transactionsBatchSize = 1
-        )
+        batchConf = batchConfig
       )
+
+    /**
+      * BitcoinOperations with mocked `loadBlocksWithTransactions`.
+      * The method is used when testing the indexer startup with Depth.
+      */
+    def dummyBitcoinOperationsStub: BitcoinOperations[IO] =
```
ivanopagano, line 148: If I may say, here I see a lot of dummy objects very custom tailored and whose purpose I cannot easily make out, at least not from the definition site.

My approach would be to try and customize the details in the body of the test, so that it's easier to compare the testing fixtures and their specific behaviour against the specific test case logic.
You can leave the dummy definition abstract, and override the missing methods in the test body, where it's apparent what you expect from the fixture.

I have a hard time understanding what's the actual use and difference of mocking a dummy subclass, and mocking the original class... 
I'm referring to the fact that I can see a 
- DummyBitcoinOperationsMock
- dummyBitcoinOperationsMock = mock[DummyBitcoin...]

How do we make use of mocking a dummy?

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperationsTest.scala
olownia: Maybe you're right, I've removed it

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperationsTest.scala
```diff
@@ -32,8 +31,70 @@ class BitcoinOperationsTest
...
+        (bitcoinClientMock.getBlockChainInfo _) expects () returning (Stream(BlockchainInfo("mainnet", 10)).covary[IO])
+
+        // test if method runs indexer with the correct range of the blocks
+        (dummyBitcoinOperationsMock.loadBlocksWithTransactions _) expects (7 to 10) returning (Stream.empty)
+
+        dummyBitcoinOperationsStub.loadBlocks(Custom(3)).compile.drain.unsafeRunSync() shouldBe (())
+      }
+
+      "index blocks with transactions with the given range" in new BitcoinOperationsStubs {
+        // check if the method gets existing blocks at the beginning
```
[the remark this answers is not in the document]
olownia: The question is if I should test it at all. All used methods are tested elsewhere. I've done this only to raise the coverage.

build.sbt
olownia: Currently `BitcoinFixtures` are in the common but in `test/` directory. Did you mean moving this to `main/`?

ivanopagano: My mistake, didn't realise that. I wonder if having fixtures for one module into another should be something we can avoid.

Are those fixtures also used in the common tests? Can we move them in the lorre module tests? Maybe we should simply duplicate those. It's an inter-modular dependency which we might want to loosen up for the sake of being more flexible.

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperationsTest.scala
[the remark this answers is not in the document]
ivanopagano: From one side, It's ok for me if it check some untested part of the system somewhat.
On the other, if it's not immediately apparent what's the subject of the test or if it's not providing any real value, we can probably question it's very existence.

Can you kindly verify assumptions regarding what I said and decide if improving the logic flow or remove it?

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperationsTest.scala
olownia: please check it now

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperationsTest.scala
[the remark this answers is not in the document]
olownia: please check it now

build.sbt
olownia: I've removed the dependency on the tests by creating a needed block row in Lorre without fixtures.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperations.scala
olownia: done

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/bitcoin/BitcoinOperations.scala
olownia: done

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/bitcoin/BitcoinPersistence.scala
```diff
@@ -34,6 +35,20 @@ class BitcoinPersistence[F[_]: Concurrent] extends LazyLogging {
       Tables.Inputs ++= transactions.flatMap(t => t.vin.map(c => (t, c))).map(_.convertTo[Tables.InputsRow]),
       Tables.Outputs ++= transactions.flatMap(t => t.vout.map(c => (t, c))).map(_.convertTo[Tables.OutputsRow])
     )
+
+  /**
+    * Get sequence of existing blocks heights from the database.
+    *
+    * @param range Inclusive range of the block's height
+    */
+  def getIndexedBlockHeights(range: Range.Inclusive): FixedSqlStreamingAction[Seq[Int], Int, Effect.Read] =
```
ivanopagano, line 44: I think being so precise on the return value is not a good choice with a public API.

Would you agree on showing something generic as `DBIO[Seq[Int]]`?

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/bitcoin/BitcoinPersistence.scala
```diff
@@ -34,6 +35,20 @@ class BitcoinPersistence[F[_]: Concurrent] extends LazyLogging {
...
+    Tables.Blocks.filter(_.height between(range.start, range.end)).map(_.height).result
+
+  /**
+    * Get the latest block from the database.
+    */
+  def getLatestIndexedBlock: SqlAction[Option[Tables.BlocksRow], NoStream, Effect.Read] =
```
ivanopagano, line 50: See above

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/bitcoin/BitcoinPersistence.scala
olownia: done

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/bitcoin/BitcoinPersistence.scala
olownia: done

PR #874, opened 2020-07-02
873 - automation of separate slick models per chain

### LINKED ISSUE #873: The source generator for slick models of database tables is not working
Currently we have multiple table namespaces that keeps separate blockchain definitions.

This isn't supported by the current app to generate the tables model.

We want to automatically generate separate files for each namespace, and put them in different scala sub-packages

ivanopagano:
Fixes #873


conseil-schema/src/main/scala/tech/cryptonomic/conseil/schema/GenSchema.scala
```diff
@@ -1,51 +1,148 @@
...
+            fibers.traverse(_.join)
+          }.void
+
+        /* sequence the operations, you might notice we use a custom context from cats.effect
+         * to run the blocking filesystem operations on
+         */
+        for {
+          model <- getModel
+          schemas <- getSchemas(model)
+          // fiber <- blockingPool.blockOn(writeSources(schemas.toList).start)
```
olownia, line 139: is this commented line needed?

conseil-schema/src/main/resources/application.conf
```diff
@@ -0,0 +1,16 @@
...
+  }
+  db {
+    driver: "org.postgresql.Driver"
+    url: "jdbc:postgresql:conseil-local"
+    user: "conseiluser"
+    password: "p@ssw0rd"
+    numThreads: 10
+    maxConnections: 10
+  }
+}
```
olownia, line 16: no newline

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/Tables.scala
```diff
@@ -1736,8 +1736,8 @@ trait Tables {
     /** Index over (level) (database name endorsing_rights_level_idx) */
     val index2 = index("endorsing_rights_level_idx", level)
 
-    /** Index over (delegate) (database name ix_delegate_slot) */
-    val index3 = index("ix_delegate_slot", delegate)
+    /** Index over (delegate,slot) (database name ix_delegate_slot) */
+    val index3 = index("ix_delegate_slot", (delegate, slot))
```
olownia, line 1740: is this related to this PR?

conseil-schema/src/main/scala/tech/cryptonomic/conseil/schema/GenSchema.scala
ivanopagano: boom, thanks

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/Tables.scala
ivanopagano: I regenerated the files with the new generator and this misalignment between definition and sources came out, so yet and no.
But I don't think it's worth its own fix PR

conseil-schema/src/main/scala/tech/cryptonomic/conseil/schema/GenSchema.scala
```diff
@@ -1,51 +1,147 @@
...
+    val externallyProvided = args.headOption.map(
+      confFile => Paths.get(confFile).toUri.resolve(configSection)
+    )
+    val classpathProvided = new URI(configSection) //this looks up in the standard application.conf as a resource
+    println(s"""
+    | Loading schema definitions from db configurations located at
+    | - externally provided: $externallyProvided
+    | - classpath fallback: application.conf$classpathProvided
+    |
+    | The source files will be generated under the folder /tmp/conseil-slick
```
ajuszczak, line 37: I guess under the folder: `slickgen.codegen.outputDir`, meaning configuration file (that might create a bit confusion, that this path is hardcoded somewhere and I hope it is not :D )

conseil-schema/src/main/scala/tech/cryptonomic/conseil/schema/GenSchema.scala
```diff
@@ -1,51 +1,147 @@
...
+    */
+  def io(
+      uri: URI,
+      outputDir: Option[String]
+  ): IO[Unit] = {
+    implicit val shift = IO.contextShift(ExecutionContext.global)
+
+    //we read configuration values
+    val dc = DatabaseConfig.forURI[PostgresProfile](uri)
+    val basePackage = dc.config.getString("codegen.package")
```
ajuszczak, line 83: Do we want to check this value against nullability? I believe, later in the code we may look for schema from `null.$schema`. I think it's worth to add some `println` here at least, for better user-experience. I don't think that this configuration will be missing in our case, but yet, someone may forgot or remove it for some reason, so please consider.

conseil-schema/src/main/scala/tech/cryptonomic/conseil/schema/GenSchema.scala
```diff
@@ -1,51 +1,147 @@
...
+        val getModel = IO.fromFuture(
+          IO.delay(
+            db.run(dc.profile.createModel(None)(ExecutionContext.global).withPinnedSession)
+          )
+        )
+
+        val getSchemas = (model: Model) =>
+          IO {
+            val schemas = model.tables.groupBy(_.name.schema).collect {
+              case (Some(schema), tables) if schema != "public" => (schema, tables)
```
ajuszczak, line 111: I think that "public" is not the only schema visible there. We still have `information_schema`, `pg_catalog` as a default schemas, or that totally different think and I'm confusing something?

PR #875, opened 2020-07-03
Change the mechanism of reading configuration as well as the style of it

ajuszczak:
This pull-request updates the style of the configuration, which is needed to improve way of passing custom properties to configuration file from the docker. This change is needed to add support for multiple blockchains.

In this change I wanted to address couple of improvements that can be done to address this issue:

a) We should not use 'sed' from shell, to replace our own macros with environment variables. Typesafe-config already supports this feature.
b) We should not copy entire configuration and fill up all of the variables, if user wants to update only couple of them.
c) Default variables should be stored in classpath's configuration, not somewhere in shell script (due to above mechanism).
d) There should be an option to run single and/or multiple blockchains for the single Conseil API instance.

All of this can be solved by using specific macro from typesafe-config. Sadly, this macro does not work when section of configuration is dynamic. Meaning:
Works:
```
{
  name: ${?CONSEIL_BLOCKCHAIN_NAME}
}
```
Won't work:
```
${?CONSEIL_BLOCKCHAIN_NAME} {
}
```
Sadly, the latter is how we have it right now. I found that not that intuitive as originally thought. Most of the things can be done by implementing this configuration as a list, instead of map.

My idea is to change the configuration into something like that:
```
platforms: [
  {
    name: "tezos"

    network: "zeronet"
    network: ${?CONSEIL_API_TEZOS_NETWORK}

    enabled: false
    enabled: ${?CONSEIL_API_TEZOS_ENABLED}

    node: {
      protocol: "https"
      protocol: ${?CONSEIL_API_TEZOS_NODE_PROTOCOL}

      hostname: "nautilus.cryptonomic.tech"
      hostname: ${?CONSEIL_API_TEZOS_NODE_HOSTNAME}

      port: 8732
      port: ${?CONSEIL_API_TEZOS_NODE_PORT}

      path-prefix: "tezos/zeronet/"
      path-prefix: ${?CONSEIL_API_TEZOS_NODE_PATH_PREFIX}
    }
  }
]
```

- [x] That change would give us better flexibility when it comes to deployment (cause you can see already, that these ${} are macros that can be filled from environment variables).
- [x] That change simplify the code of reading the configuration (at least I want to change the case-classes a bit and use as much of pureconfig mechanisms as we can, to make is easier to maintain and understand).
- [x] That change simplify the docker scripts (probably we will be able to get rid of most of the things from /docker/ folder).
- [x] That change create single point, where configuration will be stored with it's defaults. Meaning that only classpath's configuration should contain default values.
- [x] That change make the docker-compose file to specify only properties we want to overwrite, not all of them.
- During this process I'll also increase version of pureconfig, cause current one is quite old (0.11 vs 0.13 where significant change has been made). [won't be done]
- [x] Additionally, default naming convention has been unified where possible from CamelCase to kebab-case.

Sadly, I tried to update pureconfig but I there is significant change in a way of reading configuration which overcomplicates things. I guess they forgot about one mechanism and it will be implemented in a future. I'll try to address that issue nonetheless but in upcoming days.

This pull-requests addresses all of the mentioned points.


conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
```diff
@@ -75,20 +68,30 @@ object Platforms {
...
+
+    /** Defines the name of the network for specific blockchain */
     def network: String
   }
 
   /** collects all config related to a tezos network */
   final case class TezosConfiguration(
       network: String,
-      nodeConfig: TezosNodeConfiguration,
+      enabled: Boolean,
```
ivanopagano, line 85: should we give a default of `true` to this parameter, to simplify some code?

conseil-common/src/test/scala/tech/cryptonomic/conseil/common/config/PlatformsTest.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  private val networkTezos = PlatformDiscoveryTypes.Network("mainnet", "Mainnet", "tezos", "mainnet")
+  private val networkBitcoin = PlatformDiscoveryTypes.Network("testnet", "Testnet", "bitcoin", "testnet")
+
+  "Platforms.PlatformsConfiguration" should {
+      "return enabled platforms, by default" in {
+        config.getPlatforms() should contain only platformTezos
+      }
+
+      "return disabled platforms, when enabled = false" in {
```
ivanopagano, line 30: I'd suggest removing the technical detail of how to search disabled. The description will get out-of-date as soon as we change the code, and probably will not be updated with a refactor.

I'd strive to focus on the what and not the how.

conseil-common/src/test/scala/tech/cryptonomic/conseil/common/config/PlatformsTest.scala
```diff
@@ -0,0 +1,43 @@
...
+        config.getPlatforms(enabled = false) should contain only platformBitcoin
+      }
+
+      "return networks for enabled platforms and specific name, by default" in {
+        config.getNetworks("tezos") should contain only networkTezos
+      }
+
+      "return networks for disabled platforms and specific name, when enabled = false" in {
```
ivanopagano, line 38: Same considerations as above applies

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/config/LorreAppConfig.scala
```diff
@@ -132,26 +134,37 @@ object LorreAppConfig {
...
-            node <- loadConfig[TezosNodeConfiguration](namespace = s"platforms.$platform.$network.node")
-            tns <- loadConfig[Option[TNSContractConfiguration]](namespace = s"tns.$network")
-          } yield TezosConfiguration(network, node, tns)
-        case UnknownPlatform(_) => Right(UnknownPlatformConfiguration(network))
-      }
-    }
+      for {
+        config <- loadConfig[PlatformsConfiguration].flatMap {
+          _.platforms.find(x => x.platform.name == platform && x.network == network) match {
+            case Some(value) if value.enabled => Right(value)
```
ivanopagano, line 149: As before, I might prefer (personal tastes) that a specific name here would be used, to convey the meaning to the reader, without having to figure out the `value` type.

I tend to go in that direction because my goal is to make the code easy to understand for the reader, which is already a hard task! 😀

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/config/LorreAppConfig.scala
```diff
@@ -132,26 +134,37 @@ object LorreAppConfig {
...
+            case Some(value) if !value.enabled =>
+              Left(ConfigReaderFailures(new ConfigReaderFailure {
+                override def description: String =
+                  s"Could not find platform: $platform, network: $network because network is disabled"
```
ivanopagano, line 153: Minor note: what the error says it's incorrect, we actually found the platform, but we can't run it because it's disabled.

project/Dependencies.scala
```diff
@@ -169,7 +169,9 @@ object Dependencies {
 
   val conseilCommonInclude: Seq[ModuleID] =
     concat(
+      config,
       logging,
+      pureConfig,
```
ivanopagano, line 174: I think these libs can be removed from downstream modules that included them, since they also depend on common

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/config/ConseilAppConfigTest.scala
```diff
@@ -59,107 +67,71 @@ class ConseilAppConfigTest extends WordSpec with Matchers with EitherValues {
...
-
-        val results: List[Either[FailureReason, String]] = reason1
-            .asLeft[String] :: success.asRight[FailureReason] :: reason2.asLeft[String] :: Nil
-        val folded = foldReadResults(results)(_.mkString(""))
-
-        folded shouldBe 'left
-        val leftValue = folded.left.value
-        leftValue shouldBe a[FailureReason]
-        leftValue.description shouldBe "Cannot convert 'this' to that: reasons. Empty string found when trying to convert to something."
+      "extract None, when configuration for Nautlos Cloud does not exist" in {
```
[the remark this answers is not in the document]
ajuszczak: God, I have so much troubles to remember that name correctly xD

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/config/ConseilAppConfigTest.scala
```diff
@@ -59,107 +67,71 @@ class ConseilAppConfigTest extends WordSpec with Matchers with EitherValues {
...
+            |  port: 1234
+            |  path: "apiKeys/dev" // here should be an environment name after '/'
+            |  key: "exampleApiKeyDev"
+            |  delay: 10 seconds
+            |  interval: 30 seconds
+            |}
+            |""".stripMargin)
+        val typedConfig = pureconfig.loadConfig[Option[NautilusCloudConfiguration]](conf = config, namespace = "nautilus-cloud")
+
+        typedConfig.right.value shouldBe Some(
```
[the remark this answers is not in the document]
ajuszczak: Yea, `value.value` looks weird, but that's probably the best idea to do. Changing...

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilMainOutput.scala
```diff
@@ -52,12 +52,13 @@ trait ConseilMainOutput {
...
-        s"""
-           |  Platform: ${platform.name}$networks
-      """.stripMargin
-    }.mkString("\n")
+    conf.platforms
+      .groupBy(_.platform)
+      .map {
+        case (key, values) =>
+          val networks = values.map(_.network).mkString("\n  - ", "\n  - ", "\n")
+          s"Platform: ${key.name}$networks"
```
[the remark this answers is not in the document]
ajuszczak: By formatting, you mean these two spaces at the beginning? I believe that rest of it is not relevant, but correct me if I'm wrong.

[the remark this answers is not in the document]
ajuszczak: - [ ] Update formatting

ajuszczak:
- [ ] Testing

ajuszczak:
Closing this in favor of https://github.com/Cryptonomic/Conseil/pull/879 with history fixed.

PR #876, opened 2020-07-03
868-5 - Bitcoin accounts

olownia:


conseil-common/src/test/scala/tech/cryptonomic/conseil/common/bitcoin/BitcoinDbViewsTest.scala
```diff
@@ -0,0 +1,61 @@
...
+                  AND bitcoin.outputs.n = bitcoin.inputs.v_out
+                GROUP BY
+                  bitcoin.outputs.script_pub_key_addresses
+          """)
+          // test results
+          _ <- tx.transact(Tables.Blocks.result)
+          _ <- tx.transact(Tables.Transactions.result)
+          _ <- tx.transact(Tables.Inputs.result)
+          _ <- tx.transact(Tables.Outputs.result)
+
```
ivanopagano, line 53: Is it needed to read and ignore all tables results? Will it check something that might go wrong?

olownia: my mistake

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/bitcoin/Views.scala
```diff
@@ -0,0 +1,51 @@
...
+    val address: Rep[String] = column[String]("address")
+
+    /** Database column value SqlType(numeric) */
+    val value: Rep[scala.math.BigDecimal] = column[scala.math.BigDecimal]("value")
+  }
+
+  /** Collection-like TableQuery object for view Accounts */
+  lazy val Accounts = new TableQuery(tag => new Accounts(tag))
+
+  lazy val createAccountsViewSql = sqlu"""
```
piotrkosecki, line 34: I'm not sure if this should be here, as from what i see it's used only in tests. @ivanopagano @ajuszczak WDYT?

ivanopagano: I think @olownia is still looking for a way to make slick generate a view from class schema definitions, as opposed to real tables.

We never used db views till now so there's no test hook to allow you to set it up properly with the test db

olownia: I've created https://github.com/Cryptonomic/Conseil/issues/880 regarding this problem

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/bitcoin/Views.scala
```diff
@@ -0,0 +1,51 @@
...
+        script_pub_key_addresses AS address,
+        SUM(
+          CASE WHEN bitcoin.inputs.output_txid IS NULL THEN
+            value
+          ELSE
+            0
+          END) AS value
+      FROM
+        bitcoin.outputs
+        LEFT JOIN bitcoin.inputs ON bitcoin.outputs.txid = bitcoin.inputs.output_txid
```
ivanopagano, line 46: couldn't it be done with inner join or a where/having clause to avoid handling the case of no inputs transaction-id?

conseil-common/src/test/scala/tech/cryptonomic/conseil/common/bitcoin/BitcoinDbViewsTest.scala
```diff
@@ -0,0 +1,67 @@
...
+    with BitcoinInMemoryDatabaseSetup
+    with BitcoinFixtures
+    with BitcoinStubs {
+
+  implicit val contextShift: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
+
+  "Bitcoin db views" should {
+
+      "return balance for the given address" in new BitcoinPersistenceStubs(dbHandler) {
+        val unspentOutput = RpcFixtures.outputResult.copy(
```
ivanopagano, line 24: It would  be great if we could have a description of the test logic, starting from how the data is prepared and how the expectations are met.

What are the script keys pointing at?
What is the relationship between different test data? Would it make more apparent if the values set when copying different objects would point to the other data fields?

e.g.
```scala
val block1 = Rpc.blockResult.copy(height = 1, hash = "hash1")
val tx1 = Rpc.transactionResult.copy(txid = "tx1", blockhash = block1.hash, vout = List(unspent))
val input = Rpc.inputResult.copy(txid = Some(tx1.txid), vout = Some(0))
...
```

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/bitcoin/Views.scala
olownia: The question is if we want to have accounts with balance 0 in the result, I think they should be available, that's why I used left join

conseil-common/src/test/scala/tech/cryptonomic/conseil/common/bitcoin/BitcoinDbViewsTest.scala
olownia: I've added an additional description

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/bitcoin/Views.scala
ivanopagano: That makes sense, good thing

PR #879, opened 2020-07-08
Change the mechanism of reading configuration as well as the style of it

ajuszczak:
Originally taken from: https://github.com/Cryptonomic/Conseil/pull/875 but I decided to create another branch due to my mistake while doing rebase

This pull-request updates the style of the configuration, which is needed to improve way of passing custom properties to configuration file from the docker. This change is needed to add support for multiple blockchains.

In this change I wanted to address couple of improvements that can be done to address this issue:

a) We should not use 'sed' from shell, to replace our own macros with environment variables. Typesafe-config already supports this feature.
b) We should not copy entire configuration and fill up all of the variables, if user wants to update only couple of them.
c) Default variables should be stored in classpath's configuration, not somewhere in shell script (due to above mechanism).
d) There should be an option to run single and/or multiple blockchains for the single Conseil API instance.

All of this can be solved by using specific macro from typesafe-config. Sadly, this macro does not work when section of configuration is dynamic. Meaning:
Works:
```
{
  name: ${?CONSEIL_BLOCKCHAIN_NAME}
}
```
Won't work:
```
${?CONSEIL_BLOCKCHAIN_NAME} {
}
```
Sadly, the latter is how we have it right now. I found that not that intuitive as originally thought. Most of the things can be done by implementing this configuration as a list, instead of map.

My idea is to change the configuration into something like that:
```
platforms: [
  {
    name: "tezos"

    network: "zeronet"
    network: ${?CONSEIL_API_TEZOS_NETWORK}

    enabled: false
    enabled: ${?CONSEIL_API_TEZOS_ENABLED}

    node: {
      protocol: "https"
      protocol: ${?CONSEIL_API_TEZOS_NODE_PROTOCOL}

      hostname: "nautilus.cryptonomic.tech"
      hostname: ${?CONSEIL_API_TEZOS_NODE_HOSTNAME}

      port: 8732
      port: ${?CONSEIL_API_TEZOS_NODE_PORT}

      path-prefix: "tezos/zeronet/"
      path-prefix: ${?CONSEIL_API_TEZOS_NODE_PATH_PREFIX}
    }
  }
]
```

- [x] That change would give us better flexibility when it comes to deployment (cause you can see already, that these ${} are macros that can be filled from environment variables).
- [x] That change simplify the code of reading the configuration (at least I want to change the case-classes a bit and use as much of pureconfig mechanisms as we can, to make is easier to maintain and understand).
- [x] That change simplify the docker scripts (probably we will be able to get rid of most of the things from /docker/ folder).
- [x] That change create single point, where configuration will be stored with it's defaults. Meaning that only classpath's configuration should contain default values.
- [x] That change make the docker-compose file to specify only properties we want to overwrite, not all of them.
- During this process I'll also increase version of pureconfig, cause current one is quite old (0.11 vs 0.13 where significant change has been made). [won't be done]
- [x] Additionally, default naming convention has been unified where possible from CamelCase to kebab-case.

Sadly, I tried to update pureconfig but I there is significant change in a way of reading configuration which overcomplicates things. I guess they forgot about one mechanism and it will be implemented in a future. I'll try to address that issue nonetheless but in upcoming days.

This pull-requests addresses all of the mentioned points.

Code review:

- [x] Update documentation for `PlatformsConfiguration`
- [x] Remove `Unknown` blockchain type and assign different one for ArgumentsConfig
- [x] Remove HOCON macros for the properties we don't want to easily expose (these, that won't work without coding anyway)
- [x] Change key in application.conf from `CONSEIL_API_SECURITY_KEY` to `CONSEIL_API_KEY`
- [x] Change key in application.conf from `CONSEIL_API_SECURITY_ALLOW_BLANK` to `CONSEIL_API_ALLOW_BLANK_KEYS`


docker/entrypoint.sh
```diff
@@ -1,54 +1,24 @@
...
-  java -Xmx$JVM_XMX_LORRE -Dconfig.file=$CONFIG -cp /root/conseil-lorre.jar tech.cryptonomic.conseil.indexer.Lorre $XTZPLATFORM $XTZNET
-fi
+  if [ -z "${CONFIG_PATH}" ]; then
+    echo "Bootstrapping Conseil Lorre"
+    java -Xmx$JVM_XMX -cp /root/conseil-lorre.jar tech.cryptonomic.conseil.indexer.Lorre $LORRE_RUNNER_PLATFORM $LORRE_RUNNER_NETWORK
+  else
+    echo "Bootstrapping Conseil Lorre with custom configuration from: $CONFIG_PATH..."
+    java -Xmx$JVM_XMX -Dconfig.file=$CONFIG_PATH -cp /root/conseil-lorre.jar tech.cryptonomic.conseil.indexer.Lorre $LORRE_RUNNER_PLATFORM $LORRE_RUNNER_NETWORK
+  fi
+fi
```
olownia, line 24: no newline

conseil-api/src/main/resources/application.conf
```diff
@@ -11,10 +15,18 @@ conseil {
   # Replace database name, user and password in an env-specific config file.
   db {
     dataSourceClass: "org.postgresql.ds.PGSimpleDataSource"
+    dataSourceClass: ${?CONSEIL_API_DB_SOURCE_CLASS}
```
ivanopagano, line 18: To be honest, I doubt it would work with something different, but let's leave it like this... maybe in the future?

conseil-api/src/main/resources/application.conf
```diff
@@ -28,7 +40,10 @@ conseil {
   security.apiKeys {
     # List of authorized API keys needed to query Conseil
     keys: []
+    keys+= ${?CONSEIL_API_SECURITY_KEY}
```
ivanopagano, line 43: Please refer to this as CONSEIL_API_KEYS

conseil-api/src/main/resources/application.conf
```diff
@@ -28,7 +40,10 @@ conseil {
...
+    
     allow-blank: false
+    allow-blank: ${?CONSEIL_API_SECURITY_ALLOW_BLANK}
```
ivanopagano, line 46: Please consider CONSEIL_API_ALLOW_BLANK_KEYS

conseil-api/src/main/resources/application.conf
ajuszczak: You are right, but either way, we are giving this option (to change that) no matter if I'll expose it like that or now. You can always attach your own configuration file and change that. I'm just following the convention (at least for most of the properties). If you think it should be "harder" to overwrite that, we can remove it, not a problem. Maybe even we should to:

a) expose only these fields, we want to be easily changed
b) remove macros for fields, we don't want so easy to be changed (potential user may ask us, why it's not exposed and we can explain that). 

Which option would you like to follow? I like b) a bit more tho - meaning, I'll remove it.

conseil-api/src/main/resources/application.conf
ajuszczak: I believe, you can only specify a "KEY" not "KEY**S**". At least I don't know how to do that (meaning pass a list via this macro). If you know that it's a matter of changing `keys+= ${?CONSEIL_API_KEYS}` into `keys: ${?CONSEIL_API_KEYS}` then I can change that. By the way, I've added `SECURITY` in a path, just to reflect new section in it.

conseil-api/src/main/resources/application.conf
ajuszczak: Sounds better. Wonder if we should put `SECURITY` in between.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
```diff
@@ -10,32 +10,31 @@ object Platforms {
...
+  /** Represents blockchain platform for Bitcoin */
   case object Bitcoin extends BlockchainPlatform {
     val name = "bitcoin"
-    type ConfigurationType = BitcoinConfiguration
   }
 
-  case class UnknownPlatform(name: String) extends BlockchainPlatform {
-    type ConfigurationType = UnknownPlatformConfiguration
+  /** Represents blockchain platform, which is use only during arguments parsing, for better error handling */
+  case object Unknown extends BlockchainPlatform {
```
ivanopagano, line 26: Didn't you remove this?

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
```diff
@@ -44,21 +43,22 @@ object Platforms {
     * Should associates each platform to a list of its internally defined configuration type, matching the
     * inner `platform.ConfigurationType`
     */
-  case class PlatformsConfiguration(platforms: Map[BlockchainPlatform, List[PlatformConfiguration]]) {
+  case class PlatformsConfiguration(platforms: List[PlatformConfiguration]) {
```
ivanopagano, line 47: The docs need updating

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
ajuszczak: Yes I did, but then I changed the way of reading command-line arguments and for this specific reason (in here: tech.cryptonomic.conseil.indexer.config.LorreAppConfig.ArgumentsConfig) default value is required. I could use option tho, but this is not optional. I could use `null` as default value, but I'm not sure if that is a better thing to do.

I did this change (String -> BlockchainPlatform), because previously we allowed unknown platforms to go really deep in the code, when, in my opinion this should be handled as soon as possible (and it handled at reading command-line arguments) for Lorre process only.

I can actually change default value to any platform we have (Tezos even), cause it does not really matter. It will be overwritten anyway during parsing process.

conseil-api/src/main/resources/application.conf
ivanopagano: as I mentioned, I'd lean for option `b`, to remove it

conseil-api/src/main/resources/application.conf
ivanopagano: oh, I didn't realise that... is that valid HOCON? I mean, the `+=` operation?

Ok, for the security section, yet when we talk about api keys from the user perspective it was never referred to as security keys... that would create some confusion?

conseil-api/src/main/resources/application.conf
ivanopagano: see my previous comment regarding my reservations

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
ivanopagano: We can leave the unknown, or at this point we can rename it as undefined?

In any case, I just pointed that out, because I remembered you removing this case and I thought you forgot it somehow when re-opening the PR.

conseil-api/src/main/resources/application.conf
ajuszczak: Yes it is valid option. I was surprised too. 
Definition: `+= appends an element to a previous array. If the previous value was not an array, an error will result just as it would in the long form a = ${?a} [b]. Note that the previous value is optional (${?a} not ${a}), which allows a += b to be the first mention of a in the file (it is not necessary to have a = [] first).`

About second one, alright, without `SECURITY` then :)

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
ajuszczak: Yea. I mean, I just realized that there is no really point in having it. At case class initialization we can put whatever we want, it will be changed either way, so I just put `Tezos` instead of `Unknown` and that was the only place, where this case object was used. I feel better without having it :)

PR #886, opened 2020-07-10
Bitcoin accounts endpoint

ajuszczak:
This pull requests exposes accounts in the conseil-api and also adds support for creating views in postgres database during integration tests.

### DIFF
--- a/conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/bitcoin/BitcoinDataQueriesTest.scala
+++ b/conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/bitcoin/BitcoinDataQueriesTest.scala
@@ -124,9 +124,9 @@ class BitcoinDataQueriesTest
         dbHandler.run(Tables.Inputs ++= inputs).isReadyWithin(5.seconds) shouldBe true
         dbHandler.run(Tables.Outputs ++= outputs).isReadyWithin(5.seconds) shouldBe true
 
-        whenReady(sut.fetchAccountByAddress("script_pub_key_address_1")) { result =>
-          result.value should (contain key "address" and contain value Some("script_pub_key_address_1"))
-          result.value should (contain key "value" and contain value Some("1"))
+        whenReady(sut.fetchAccountByAddress("script_pub_key_address_2")) { result =>
+          result.value should (contain key "address" and contain value output2.scriptPubKeyAddresses)
+          //TODO Once we will replace BigDecimal into something better, we should compare 'value' field here as well
         }
       }
     }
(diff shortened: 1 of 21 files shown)


conseil-common-testkit/src/main/scala/tech/cryptonomic/conseil/common/testkit/InMemoryDatabaseSetup.scala
```diff
@@ -14,16 +16,42 @@ trait InMemoryDatabaseSetup {
...
+  }
+
+  /*** Represents fixture for common slick tables */
+  case class TFixture[U](query: TableQuery[_ <: Table[U]]) extends Fixture {
     def create: DBIOAction[Unit, NoStream, Effect.Schema] = query.schema.create
     def delete: DBIOAction[Int, NoStream, Effect.Write] = query.delete
   }
 
+  /*** Represents fixture for custom slick views */
+  case class VFixture(createSql: Iterable[String]) extends Fixture {
```
olownia, line 48: actually this is a fixture for any SQL statement, not only views, maybe we should change the name to more generic?

ivanopagano: as long as the commodity `Fixture.table/view` methods are there, I agree that this can be made general. though I wonder if it makes sense to run `.create` on something that, I don't know, reset some value or configures a db feature...

ivanopagano: Since internally it uses a ddl, what do you have in mind? Sequences, functions?

olownia: I would see it more like: Create tables and run some queries (that can contain view, function or whatever) but it's up to Adrian

```scala
Fixture.table(Tables.Blocks)
Fixture.sql(Views.AccountsViewSql)
```

ajuszczak: Since there is no special need for creating functions or anything like that, I'm going to stick to what we have now. As you can see, it would be (probably) quite easy to do, but since creating anything different than view is not tested and we do not have time for that, I'm not going to extend that only because it is possible to do.

I'm not saying it is not possible or bad, only that it is not needed yet :)

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/bitcoin/BitcoinDataQueriesTest.scala
```diff
@@ -104,6 +104,31 @@ class BitcoinDataQueriesTest
...
+      "return proper account by address" in {
+        // given
+        dbHandler.run(Tables.Blocks ++= blocks).isReadyWithin(5.seconds) shouldBe true
+        dbHandler.run(Tables.Transactions ++= transactions).isReadyWithin(5.seconds) shouldBe true
+        dbHandler.run(Tables.Inputs ++= inputs).isReadyWithin(5.seconds) shouldBe true
+        dbHandler.run(Tables.Outputs ++= outputs).isReadyWithin(5.seconds) shouldBe true
+
+        whenReady(sut.fetchAccountByAddress("script_pub_key_address_2")) { result =>
+          result.value should (contain key "address" and contain value output2.scriptPubKeyAddresses)
+          //TODO Once we will replace BigDecimal into something better, we should compare 'value' field here as well
```
ivanopagano, line 129: Please expand on this note: give a reference to what is the BigDecimal in this case, what is the impact and limitations that this entails on the current tests, and how it's supposed to be fixed in the future.

Please write comments like someone other than you would need to take charge of this task, so he/she would need all the details to manage by him/herself

PR #888, opened 2020-07-16
887 - db schema for Ethereum blocks, transactions and logs

molowny:


sql/conseil.sql
```diff
@@ -869,3 +869,49 @@ FROM
     AND bitcoin.outputs.n = bitcoin.inputs.v_out
   GROUP BY
     bitcoin.outputs.script_pub_key_addresses;
+
+CREATE SCHEMA ethereum;
```
ajuszczak, line 873: Since we want to support multiple networks in the same schema, shouldn't we add "network" column as we discussed with Vishakh?

molowny: I've added `blockchain` column below.

ajuszczak: Note, we decided to name it "network" with limited options.

ajuszczak:
Can you add this class, to be ignored by code coverage? Besides, looks good 👍

sql/conseil.sql
```diff
@@ -899,7 +903,7 @@ CREATE TABLE ethereum.blocks (
 CREATE TABLE ethereum.transactions (
   hash text NOT NULL PRIMARY KEY,
   block_hash text NOT NULL,
-  block_number text NOT NULL,
+  block_number integer NOT NULL,
```
ivanopagano, line 906: will it be enough or it might grow out-of-bounds?

sql/conseil.sql
```diff
@@ -869,3 +869,69 @@ FROM
...
+
+CREATE SCHEMA ethereum; 
+
+-- List of the Ethereum based blockchains implemented in Conseil
+CREATE TYPE ethereum.networks AS ENUM ('ethereum', 'quorum');
+
+-- Table is based on eth_getBlockByHash from https://eth.wiki/json-rpc/API
+CREATE TABLE ethereum.blocks (
+  hash text NOT NULL PRIMARY KEY,
+  number integer NOT NULL,
```
ivanopagano, line 881: will it be enough or it might grow out-of-bounds?

sql/conseil.sql
```diff
@@ -869,3 +869,69 @@ FROM
     AND bitcoin.outputs.n = bitcoin.inputs.v_out
   GROUP BY
     bitcoin.outputs.script_pub_key_addresses;
+
+CREATE SCHEMA ethereum; 
+
+-- List of the Ethereum based blockchains implemented in Conseil
+CREATE TYPE ethereum.networks AS ENUM ('ethereum', 'quorum');
```
ivanopagano, line 876: Food for thought: Will it work fine like this, or will it hinder code flexibility to "encode" this choice at the DB level with no explicit mapping at the code level?

ajuszczak:
I'm in the middle of building API queries on top of it and I find out one issue we need to address, which is having `network` in all entities. This is needed due to the fact that queries for listing all transactions and logs with specific `Query` (configurable by user) are too dynamic to be supported by slick type safe queries. I'm using `queryWithPredicate` method and I'm adding additional `predicate` which is `network eq <network>`. Since I'm doing that and there is no other way now, to support type safe and dynamic approach of these queries we need to duplicate this information.

By the cost of space, we will increase performance, because (let's skip Slick issue for a second) that we will need to join transactions/logs with blocks to see which network it belongs to. Could we add `network` for transactions and logs and see how it goes? Maybe we will be able to optimize that in the future.

sql/conseil.sql
piotrkosecki: I also have a problem with this line, I recently used `rinkeby` network for ETH and I believe it should be included in metadata conf, not hardcoded in the DB. Other reason is that, I don't know how often new test network appears, so we would need to add new network to the schema.

ajuszczak: By the way, it's not going to be `ethereum` and `quorum` rather `mainnet` (for Ethereum) and `quorum` (for JP Morgan Quorum). If they have different networks, then it should be named `quorum-xyz` I believe. Note that the url for this endpoint will be: `/data/ethereum/mainnet/blocks` or `/data/ethereum/quorum/blocks`. If we want to have `/data/quorum/mainnet/blocks` for the latter example, then we are not talking about networks, but about platforms and it makes entire configuration a bit more complicated, because this part is not "dynamic" in current design. We agreed to treat `Quorum` as `Ethereum`. As you can see, this part is quite dynamic and actually API is going to handle the case, when given network does not exists in the database, so I'm not sure either if we should limit this here. After some consideration maybe we should remove that in initial commits?

molowny: I've removed network column

ivanopagano:
@molowny maybe it's appropriate to put a comment on the sql schema definition to remark how we plan to do some dynamic version of the ethereum/quorum tablespace to remove the duplication in the future?

PR #889, opened 2020-07-21
Improve logging and do not pre-cache while all networks are disabled

ajuszczak:
In a situation, when all of the networks are disabled (and by default they are), application did not initialized cache properly, which caused following stack trace to be displayed:
```
10:43:05.607 [conseil-system-akka.http.dispatcher-7] ERROR         t.cryptonomic.conseil.api.ConseilApi - Pre-caching metadata failed
java.lang.UnsupportedOperationException: empty.reduceLeft
	at scala.collection.LinearSeqOptimized.reduceLeft(LinearSeqOptimized.scala:139)
	at scala.collection.LinearSeqOptimized.reduceLeft$(LinearSeqOptimized.scala:138)
  | => aat scala.collection.immutable.List.reduceLeft(List.scala:89)
	at scala.collection.TraversableOnce.reduce(TraversableOnce.scala:213)
	at scala.collection.TraversableOnce.reduce$(TraversableOnce.scala:213)
	at scala.collection.AbstractTraversable.reduce(Traversable.scala:108)
	at tech.cryptonomic.conseil.api.routes.platform.discovery.GenericPlatformDiscoveryOperations.$anonfun$preCacheEntities$2(GenericPlatformDiscoveryOperations.scala:211)
```

Additionally, in some cases it could fall into infinite loop, causing computer to be not responding.

This pull-requests adds additional logging information as well as prevents caching from being done, when there is no network enabled in configuration.

### DIFF
--- a/conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
+++ b/conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
@@ -10,6 +10,7 @@ import akka.stream.ActorMaterializer
 import cats.effect.{ContextShift, IO}
 import ch.megard.akka.http.cors.scaladsl.CorsDirectives.cors
 import com.typesafe.scalalogging.LazyLogging
+import tech.cryptonomic.conseil.api.ConseilApi.NoNetworkEnabledError
 import tech.cryptonomic.conseil.api.config.ConseilAppConfig.CombinedConfiguration
 import tech.cryptonomic.conseil.api.directives.{EnableCORSDirectives, RecordingDirectives, ValidatingDirectives}
 import tech.cryptonomic.conseil.api.metadata.{AttributeValuesCacheConfiguration, MetadataService, UnitTransformation}
@@ -30,6 +31,11 @@ import scala.concurrent.ExecutionContext
 import scala.util.{Failure, Success}
 
 object ConseilApi {
+
+  /** Exception, which is thrown when no network is enabled in configuration */
+  case class NoNetworkEnabledError(message: String) extends Exception(message)
+
+  /** Creates Conseil API based on a given configuration */
   def create(config: CombinedConfiguration)(implicit system: ActorSystem): ConseilApi = new ConseilApi(config)
 }
 
@@ -195,13 +201,11 @@ class ConseilApi(config: CombinedConfiguration)(implicit system: ActorSystem)
         case Success(_) => logger.info("Pre-caching attributes successful!")
       }
     } else {
-      logger.error(
-        """
-          |Pre-catching could not be done, because all of the block-chains are disabled! 
-          | Application is probably misconfigured. Please double check configuration files and start application again.
-          |""".stripMargin
-      )
-      sys.exit(1)
+      throw NoNetworkEnabledError("""
+        |Pre-catching could not be done, because all of the block-chains are disabled!
+        | It means that application is NOT configured properly.
+        | Please double check configuration file and start API again.
+        |""".stripMargin)
     }
   }
 
--- a/conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
+++ b/conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
@@ -201,11 +201,13 @@ class ConseilApi(config: CombinedConfiguration)(implicit system: ActorSystem)
         case Success(_) => logger.info("Pre-caching attributes successful!")
       }
     } else {
-      throw NoNetworkEnabledError("""
-        |Pre-catching could not be done, because all of the block-chains are disabled!
-        | It means that application is NOT configured properly.
-        | Please double check configuration file and start API again.
-        |""".stripMargin)
+      throw NoNetworkEnabledError(
+        """|Pre-caching can't be done, because there is no enabled block-chain defined.
+           | This probably means the application is NOT properly configured.
+           | The API needs a platform section to be defined, which lists at least one enabled chain and network.
+           | Please double-check the configuration file and start the API service again.
+           |""".stripMargin
+      )
     }
   }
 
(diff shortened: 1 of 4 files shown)


conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
```diff
@@ -183,15 +184,17 @@ class ConseilApi(config: CombinedConfiguration)(implicit system: ActorSystem)
       network <- transformation.overrideNetworks(platform.path, config.platforms.getNetworks(platform.name))
     } yield platform -> network
 
-    cachedDiscoveryOperations.init(visibleNetworks).onComplete {
-      case Failure(exception) => logger.error("Pre-caching metadata failed", exception)
-      case Success(_) => logger.info("Pre-caching successful!")
-    }
+    if (visibleNetworks.nonEmpty) { // At least one blockchain is enabled
```
molowny, line 187: should application run to this point if there isn't enabled network? Maybe we should stop it earlier?

ajuszczak: That's good questions. I was thinking about that and I decided to let I go, because still you can see the version you are running. I believe you can also take a look at the OpenDoc (shouldn't OpenDoc work in the same way that endpoints are working, meaning show only what is enabled/disabled?), so you can get some value from it, but still no data, no metadata can be served there. Maybe, at some point we would like to host the documentation under separate instance, just to make it a bit more responsive while in the same time, heavy load is on another instances?

I guess it all comes down to whether or not, all of the networks being disabled is just an another state of the application or something what should never happen?

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilMainOutput.scala
```diff
@@ -56,7 +56,8 @@ trait ConseilMainOutput {
       .groupBy(_.platform)
       .map {
         case (platform, configuration) =>
-          val networks = configuration.map(_.network).mkString("\n  - ", "\n  - ", "\n")
+          val networks =
+            configuration.map(c => s"${c.network} (enabled: ${c.enabled})").mkString("\n  - ", "\n  - ", "\n")
```
ivanopagano, line 60: Will it be less "noisy" if we simply print a flag when a network is disabled?

would look like

```
 - mainnet
 - netflix (disabled)
 - fakenet
```?

Comment (ajuszczak) [reply conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilMainOutput.scala:60]: Well, in a sea of networks then probably this proposition would work better, but considering there are mostly 2 by blockchain (if even), then this change from readability perspective is rather cosmetic. Nonetheless, I've change that to what you proposed, hope it will looks nice now :)

File: conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala (hunk 1 above)
Comment (ajuszczak) [reply conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala:193]: Should be done now, please take a look

File: conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala [hunk 3]
```diff
@@ -194,7 +200,13 @@ class ConseilApi(config: CombinedConfiguration)(implicit system: ActorSystem)
         case Failure(exception) => logger.error("Pre-caching attributes failed", exception)
         case Success(_) => logger.info("Pre-caching attributes successful!")
       }
-    } else logger.warn("Pre-catching could not be done, because all of the block-chains are disabled!")
+    } else {
+      throw NoNetworkEnabledError("""
+        |Pre-catching could not be done, because all of the block-chains are disabled!
+        | It means that application is NOT configured properly.
+        | Please double check configuration file and start API again.
+        |""".stripMargin)
```
ivanopagano, line 208: Sorry to be pedantic on the wording, but I might suggest some tweaks, and correct a few typos

>Pre-caching can't be done, because there is no enabled block-chain defined.
This probably means the application is NOT properly configured.
The API needs a platform section to be defined, which lists at least one enabled chain and network.
Please double-check the configuration file and start the API service again.

ivanopagano: I appreciate the intent, nonetheless, the spirit was correct

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilMainOutput.scala
```diff
@@ -56,10 +56,16 @@ trait ConseilMainOutput {
...
           s"  Platform: ${platform.name}$networks"
       }
       .mkString("\n")
 
+  /* prepare output to display existing networks */
+  private def showAvailableNetworks(configuration: List[PlatformConfiguration]): String =
+    configuration.map { c =>
+      val disabled = if (!c.enabled) " (disabled)" else ""
+      s"${c.network}$disabled"
+    }.mkString("\n  - ", "\n  - ", "\n")
```
ivanopagano, line 69: 🙇

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/util/RetryStrategy.scala
```diff
@@ -0,0 +1,17 @@
...
+
+object RetryStrategy {
+
+  /** Method which defines custom strategy, when retry mechanism should not continue
+    *
+    * Currently retry stops when:
+    * - there is no network enabled in the configuration
+    * */
+  val retryGiveUpStrategy: Throwable => Boolean = {
+    case _: NoNetworkEnabledError => true
```
ivanopagano, line 13: Is the error message properly shown on the console when this happens? Is there a better way to log it, so that we make sure it will also show on logstash?

ajuszczak: Zrzut ekranu 2020-07-22 11 47 04

This is how it looks. Like regular exception I guess, unless you want me to put it in single line instead of indentation?

ivanopagano: It's good

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/ConseilApi.scala
ajuszczak: Don't be sorry :) I've updated that.

PR #891, opened 2020-07-22
887-1 - Ethereum JSON-RPC client and persistence

molowny:
This is an implementation of JSON-RPC communication for the Ethereum node. The solution is based on previously created integration for the Bitcoin. There aren't any new logic.

### DIFF
--- a/conseil-common/src/main/scala/tech/cryptonomic/conseil/common/ethereum/EthereumPersistence.scala
+++ b/conseil-common/src/main/scala/tech/cryptonomic/conseil/common/ethereum/EthereumPersistence.scala
@@ -10,7 +10,7 @@ import slick.jdbc.PostgresProfile.api._
 
 import tech.cryptonomic.conseil.common.util.Conversion
 import tech.cryptonomic.conseil.common.util.Conversion.Syntax._
-import tech.cryptonomic.conseil.common.ethereum.rpc.json.{Block, Transaction}
+import tech.cryptonomic.conseil.common.ethereum.rpc.json.{Block, Log, Transaction}
 import tech.cryptonomic.conseil.common.ethereum.EthereumPersistence._
 
 /**
@@ -109,6 +109,27 @@ object EthereumPersistence {
           v = from.v,
           r = from.r,
           s = from.s
+        )
+    }
+
+  /**
+    * Convert form [[Log]] to [[Tables.LogsRow]]
+    * TODO: This conversion should be done with the Chimney,
+    *       but it's blocked due to the https://github.com/scala/bug/issues/11157
+    */
+  implicit val logToLogsRow: Conversion[Id, Log, Tables.LogsRow] =
+    new Conversion[Id, Log, Tables.LogsRow] {
+      override def convert(from: Log) =
+        Tables.LogsRow(
+          address = from.address,
+          blockHash = from.blockHash,
+          blockNumber = Integer.decode(from.blockNumber),
+          data = from.data,
+          logIndex = from.logIndex,
+          removed = from.removed,
+          topics = from.topics.mkString(","),
+          transactionHash = from.transactionHash,
+          transactionIndex = from.transactionIndex
         )
     }
 
--- a/conseil-common/src/main/scala/tech/cryptonomic/conseil/common/ethereum/EthereumPersistence.scala
+++ b/conseil-common/src/main/scala/tech/cryptonomic/conseil/common/ethereum/EthereumPersistence.scala
@@ -1,0 +1,115 @@
+package tech.cryptonomic.conseil.common.ethereum
+
+import java.sql.Timestamp
+import java.time.Instant
+
+import cats.Id
+import cats.effect.{Concurrent, Resource}
+import com.typesafe.scalalogging.LazyLogging
+import slick.jdbc.PostgresProfile.api._
+
+import tech.cryptonomic.conseil.common.util.Conversion
+import tech.cryptonomic.conseil.common.util.Conversion.Syntax._
+import tech.cryptonomic.conseil.common.ethereum.rpc.json.{Block, Transaction}
+import tech.cryptonomic.conseil.common.ethereum.EthereumPersistence._
+
+/**
+  * Ethereum persistence into the database using Slick.
+  */
+class EthereumPersistence[F[_]: Concurrent] extends LazyLogging {
+
+  /**
+    * Create [[DBIO]] seq with blocks and transactions that can be wrap into one transaction.
+    *
+    * @param block JSON_RPC block
+    * @param transactions JSON_RPC block's transactions
+    */
+  def createBlock(
+      block: Block,
+      transactions: List[Transaction]
+  ): DBIOAction[Unit, NoStream, Effect.Write] =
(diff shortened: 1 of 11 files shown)


ivanopagano:
As a best-practice, can we consider to always provide (unless patently useless) a description to each PR?
It helps to frame the problem, identify scope-unrelated changes, in general keep track of what was done, even as post-facto information/

Thanks

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/ethereum/rpc/EthereumRpcClient.scala
```diff
@@ -0,0 +1,79 @@
...
+    */
+  def getBlockWithTransactions(batchSize: Int): Pipe[F, Block, (Block, List[Transaction])] =
+    stream =>
+      for {
+        block <- stream // get each Block from the stream
+        transactions <- Stream
+          .emits(block.transactions)
+          .map(EthGetTransactionByHash.request)
+          .through(client.stream[EthGetTransactionByHash.Params, Transaction](batchSize))
+          .chunkN(block.transactions.size)
```
ivanopagano, line 65: Out of curiosity: why choosing the whole number of transactions to chunk the stream?

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/ethereum/rpc/EthereumRpcCommands.scala
```diff
@@ -0,0 +1,86 @@
...
+    * `eth_getBlockByNumber` Ethereum JSON-RPC api method.
+    * If verbosity is true it returns the full transaction objects,
+    * if false only the hashes of the transactions.
+    */
+  object EthGetBlockByNumber extends EthereumRpcMethod {
+    val rpcMethod = "eth_getBlockByNumber"
+    case class Params(number: String, verbosity: Boolean)
+    def request(number: String) = RpcRequest("2.0", rpcMethod, Params(number, false), s"egbbn_$number")
+
+    implicit val encodeParams: Encoder[Params] = new Encoder[Params] {
```
ivanopagano, line 38: pro-tip: since java8+ you can make use of functional interfaces to reduce the boilerplate for Single-method-instances as a lambda, i.e.

```scala
implicit val encodeParams: Encoder[Params] = (params: Params) =>
  Json.arr(
    Json.fromString(params.number),
    Json.fromBoolean(params.verbosity)
  )
```

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/ethereum/rpc/EthereumRpcClient.scala
molowny: this stream is able to fetch block's transaction in batches and save them in one transaction at the end, so the algorithm looks like: get block -> split into transactions -> get transactions in batches -> merge blocks with transactions again (`chunkN`) -> save block with the transactions in one db transaction

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/ethereum/rpc/EthereumRpcCommands.scala
molowny: I did it according to Circe doc, your code actually looks better :)

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/ethereum/rpc/EthereumRpcCommands.scala
```diff
@@ -0,0 +1,86 @@
...
+    * Returns the number of most recent block.
+    */
+  object EthBlockNumber extends EthereumRpcMethod {
+    val rpcMethod = "eth_blockNumber"
+    def request = RpcRequest("2.0", rpcMethod, Nil, "bn")
+  }
+
+  /**
+    * `eth_getBlockByNumber` Ethereum JSON-RPC api method.
+    * If verbosity is true it returns the full transaction objects,
```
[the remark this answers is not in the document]
molowny: I've added info that we only use a `false` option.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/ethereum/rpc/EthereumRpcClient.scala
```diff
@@ -0,0 +1,79 @@
...
+  *
+  *   val ethereumClient = new EthereumClient[IO](rpcClient)
+  *
+  *   // To call [[fs2.Pipe]] methods use:
+  *   Stream(1, 2).through(ethereumClient.getBlockByNumber(batchSize = 10)).compile.toList
+  *   // The result will be:
+  *   val res0: List[Block] = List(block1, block2)
+  * }}}
+  */
+class EthereumClient[F[_]: Concurrent](
```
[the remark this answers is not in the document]
molowny: This is only client implementation. It'll be used in the next PR with the Lorre.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/ethereum/EthereumPersistence.scala
```diff
@@ -0,0 +1,136 @@
...
+        transactionsRoot = from.transactionsRoot,
+        uncles = Option(from.uncles).filter(_.nonEmpty).map(_.mkString(",")),
+        timestamp = Timestamp.from(Instant.ofEpochSecond(Integer.decode(from.timestamp).toLong))
+      )
+  }
+
+  /**
+    * Convert form [[Transaction]] to [[Tables.TransactionsRow]]
+    * TODO: This conversion should be done with the Chimney,
+    *       but it's blocked due to the https://github.com/scala/bug/issues/11157
```
piotrkosecki, line 92: I've tested on this case and chimney transformation compiles correctly:
```
      import io.scalaland.chimney.dsl._

      override def convert(from: Transaction) =
        from
          .into[Tables.TransactionsRow]
          .withFieldComputed(_.blockNumber, x => Integer.decode(x.blockNumber).toInt)
          .transform
```
Can it be replaced in this case, or am I missing something?

molowny: shouldn't it be done for all places with the `tech.cryptonomic.conseil.common.util.Conversion`? I think that we should get rid of `Conversion` first, then refactor all the places with it to use only `chimney`. Of course, I can only change that here too. @ivanopagano WDYT?

ivanopagano: I'll give you some feedback on this, but won't be able right now.

I need to test the effect on updating to latest chimney (as suggested to me) to see if it solves anything.

I've sent a bugfix for that issue mentioned, but it doesn't seem to solve the issue on my side when I tried locally... so it might not be related.

Eventually, the latest chimney api would allow us to get rid of the Conversions, I think. And be able to operate directly with the chimney-defined type-class which does the same job...

PR #893, opened 2020-07-27
Feature/887 ethereum endpoints

ajuszczak:
This pull-request introduces API Endpoints for Ethereum and Quorum which is based on the Ethereum (and compatible with it).

During this change, `endpoints` library has been bumped but only couple of versions due to significant changes made there and potential of this pull-request to explode with number of changes. I'd like to advise to do it in completely separate pull-request.

Creating draft and waiting for Lorre process to validate `/data/` endpoints


conseil-api/src/main/resources/metadata/quorum.mainnet.conf
```diff
@@ -0,0 +1,7 @@
+// Since currently both networks (JP Morgan Quorum and Ethereum) have the same data entites,
+// we can use the same metadata configuration for both
+{
+  include "ethereum.mainnet.conf"
+
+  visible: true
+}
```
piotrkosecki, line 7: please add new line

conseil-api/src/main/resources/metadata/ethereum.mainnet.conf
```diff
@@ -0,0 +1,147 @@
...
+        to {
+          visible: true
+        }
+        transaction_index {
+          visible: true
+        }
+        value {
+          visible: true
+        }
+        v {
```
piotrkosecki, line 102: just a question, does those letters mean anything specific? Maybe it would be a good idea to override their names with `name:`?

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataEndpointsCreator.scala
```diff
@@ -0,0 +1,63 @@
...
+trait EthereumDataEndpointsCreator extends ApiDataEndpoints with ApiDataJsonSchemas with EthereumFilterFromQueryString {
+
+  /** V2 Blocks endpoint definition */
+  private[ethereum] def blocksEndpoint(root: Path[String]) : Endpoint[((String, EthereumFilter), Option[String]), Option[
+    List[QueryResponse]
+  ]] =
+    endpoint(
+      request = get(url = root / "blocks" /? ethereumQsFilter, headers = optHeader("apiKey")),
+      response = compatibilityQuery[List[QueryResponse]]("blocks"),
+      tags = List("Blocks")
```
piotrkosecki, line 16: as we add new chains I think we should include the name of the chain in the tags

ajuszczak: Would you like me to add that to all of the endpoints?

conseil-api/src/main/resources/metadata/ethereum.mainnet.conf
ajuszczak: Let me check documentation. Sounds like a good idea. Probably Scala-doc could be also extended to contain description for such fields.

ajuszczak: I managed to find the description (at least from the official wiki). Not sure if it's going to help but still... No entity for it, so no Scala-doc as well.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataEndpointsCreator.scala
ajuszczak: Well, did that for all of them :)

conseil-api/src/main/resources/metadata/ethereum.mainnet.conf
piotrkosecki: sorry, my mistake, what I meant was `display-name` not `name`.

ajuszczak: That is correct. I should also verify that. Changed, thanks for letting know!

conseil-api/src/main/resources/metadata/quorum.mainnet.conf
```diff
@@ -0,0 +1,7 @@
+// Since currently both networks (JP Morgan Quorum and Ethereum) have the same data entites,
+// we can use the same metadata configuration for both
+{
+  include "ethereum.mainnet.conf"
+
+  visible: true
```
ivanopagano, line 6: I'm just wondering, what if we moved the visibility flag outside the metadata file into the main "including file"?

That would centralize the switches, and allow to easily check the situation at a glance.

What do you think? @ajuszczak @molowny ?

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/openapi/OpenApiDoc.scala
```diff
@@ -40,6 +43,18 @@ object OpenApiDoc
...
+    ethereumBlockByHashEndpoint,
+    ethereumTransactionsEndpoint,
+    ethereumTransactionByHashEndpoint,
+    ethereumLogsEndpoint,
+    quorumBlocksEndpoint,
+    quorumBlocksHeadEndpoint,
+    quorumBlockByHashEndpoint,
+    quorumTransactionsEndpoint,
+    quorumTransactionByHashEndpoint,
+    quorumLogsEndpoint,
```
ivanopagano, line 57: What would you think about each platform providing a method that collects all endpoints for each?
The end result should look something like

```scala
openApi(Info...) {
  queryEndpoint ::
  platformsEndpoint ::
  entitiesEndpoint ::
  tezosEndpoints ++
  bitcoinEndpoints ++
  ethereumEndpoints ++
  quorumEndoints : _*
}
```

ajuszczak: I'm pretty sure this is technically not possible due to very weird types returned by endpoints4j library. I tried that already, when I introduced second blockchain. I don't want to put compiler errors here, but if you want you can quickly validate your solution. Just try to add: `Seq(bitcoinOutputsEndpoint, bitcoinAccountsEndpoint): _*` at the end. That's pretty sad. Maybe it is solved in the latest endpoints4j library, but I stuck couple days ago on parsing arguments for Tezos Endpoint. There are many of them, where endpoints4j supports up to 5 out of the box, so we need to solve it first. Either way, it's kind of mess there (in this lib, with these types).

conseil-api/src/main/resources/metadata/quorum.mainnet.conf
ajuszczak: Do you mean, something like that: 
```
ethereum {
    visible: true
    networks {
      mainnet {
        include "metadata/ethereum.mainnet.conf"
        visible: true
      }
    }
  }
```
in the `metadata.conf`? If so, then I like it a bit more. I mean, at least you see that this specific network is visible. Which is more important to see quickly than which fields we are exposing from specific entities in the network. I like the idea.

ivanopagano: Yes, exactly what I meant

ajuszczak: Let me change that and we will see :)

ivanopagano: I'm not sure if this would mean that we have to read the metadata differently than now...

When we include, does a sub-object gets created within the network object, or are the included-object fields "merged" with the including?

To make clear: does the `visible` attribute in the included metadata object already appear as an attribute of the `mainnet` network? Or is it nested in another object?
Personally I can't see no other way to think about it, since the included object is not assigned to any field of the including object, so a form of merging of attributes must happen

ajuszczak: They will be merged (with latest appearance of the same field being the one set at the end). Final file will looks exactly the same. I'm not planning to change reading implementation. Just configuration file.

ajuszczak: So:
```
include "ethereum.mainnet.conf"
```
and 
```
{
   include "ethereum.mainnet.conf"
}
```
technically is the same. I had to add `{ }`, because I wanted to śet visibility field. Since we do not need it anymore, I can simply have single include file there.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ApiFilterQueryString.scala
```diff
@@ -0,0 +1,21 @@
+package tech.cryptonomic.conseil.api.routes.platform.data
+
+import cats.Functor
+import endpoints.algebra
+import tech.cryptonomic.conseil.api.routes.platform.data.ApiFilter.Sorting
+import tech.cryptonomic.conseil.api.routes.platform.data.ApiFilter.Sorting._
+
+trait ApiFilterQueryString { self: algebra.JsonEntities =>
```
ivanopagano, line 8: Maybe docs would be helpful to make sense of all our bandwagon of traits that are mixed and meshed?

Personally I have a hard time following the logic of all the pieces when I read the code without having to put hands on it....

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ApiValidation.scala
```diff
@@ -0,0 +1,32 @@
...
+import endpoints.algebra.Documentation
+import tech.cryptonomic.conseil.api.routes.platform.data.CsvConversions._
+import tech.cryptonomic.conseil.common.generic.chain.DataTypes.{
+  OutputType,
+  QueryResponseWithOutput,
+  QueryValidationError
+}
+import tech.cryptonomic.conseil.common.util.Conversion.Syntax._
+
+object ApiValidation {
```
ivanopagano, line 16: Once more, maybe docs would greatly improve the reading experience and understanding

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ApiCirceJsonSchema.scala
```diff
@@ -1,36 +1,14 @@
...
-    case Left(errors) =>
-      complete(StatusCodes.BadRequest -> s"Errors: \n${errors.mkString("\n")}")
-    case Right(QueryResponseWithOutput(queryResponse, OutputType.csv)) =>
-      complete(HttpEntity.Strict(ContentTypes.`text/csv(UTF-8)`, ByteString(queryResponse.convertTo[String])))
-    case Right(QueryResponseWithOutput(queryResponse, OutputType.sql)) =>
-      complete(HttpEntity.Strict(ContentTypes.`text/plain(UTF-8)`, ByteString(queryResponse.head("sql").get.toString)))
-    case Right(success) =>
-      response(success)
-  }
+trait ApiCirceJsonSchema extends AkkaCirceJsonSchema with ApiDataJsonSchemas {
```
ivanopagano, line 11: Where has the validation actually gone? Which module does it now?

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataHelpers.scala
```diff
@@ -0,0 +1,35 @@
...
+import tech.cryptonomic.conseil.api.routes.platform.data.ApiValidation.defaultValidated
+import tech.cryptonomic.conseil.api.routes.platform.data.{
+  ApiCirceJsonSchema,
+  ApiDataEndpoints,
+  ApiDataJsonSchemas,
+  ApiFilterQueryString
+}
+import tech.cryptonomic.conseil.common.generic.chain.DataTypes.QueryValidationError
+
+/** Represents helper for Data Endpoints that can be used to implement custom encoder for Bitcoin specific types */
```
ivanopagano, line 16: Wrong reference to Bitcoin in the docs, copy-paste error

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataRoutesCreator.scala
```diff
@@ -0,0 +1,126 @@
...
+import tech.cryptonomic.conseil.common.ethereum.EthereumTypes.EthereumBlockHash
+import tech.cryptonomic.conseil.common.generic.chain.DataTypes.QueryResponseWithOutput
+import tech.cryptonomic.conseil.common.metadata
+import tech.cryptonomic.conseil.common.metadata.{EntityPath, NetworkPath, PlatformPath}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/** Trait, which contains routes for Ethereum-related block-chains */
+trait EthereumDataRoutesCreator
+  extends EthereumDataHelpers
```
ivanopagano, line 21: I'm not sure: is this a trait extending a concrete class?

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataRoutesCreator.scala
```diff
@@ -0,0 +1,126 @@
...
+    concat(
+      blocksHeadRoute,
+      blockByHashRoute,
+      blocksRoute,
+      transactionsRoute,
+      transactionByIdRoute
+    )
+
+  /** Function for validation of the platform and network with flatten */
+  private def platformNetworkValidation[A](network: String)(
```
ivanopagano, line 113: Is this really needed as a separate method from the `pathValidation`?
It only adds a `NetworkPath` creation that can be directly done inside the latter... I think merging the two would simplify things eventually.

```scala
private def platformNetworkValidation[A](network: String)(operation: ...): Future[Option[A]] = {
  if (metadataService.exists(NetworkPath(network, platform)) operation
  else Future.successful(None)
}
```

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/tezos/TezosFilterFromQueryString.scala
```diff
@@ -4,92 +4,57 @@ import cats.Functor
...
-          operationParticipants.toSet,
-          operationKinds.toSet,
-          accountIDs.toSet,
-          accountManagers.toSet,
-          accountDelegates.toSet,
-          sortBy,
-          order.flatMap(Sorting.fromString)
-        )
-    }
+  val tezosQsFilter: QueryString[TezosFilter] = filterQs.map(TezosFilter.tupled)
```
ivanopagano, line 58: 😃

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/bitcoin/BitcoinDataOperationsTest.scala
```diff
@@ -69,7 +68,7 @@ class BitcoinDataQueriesTest
         dbHandler.run(Tables.Blocks ++= blocks).isReadyWithin(5.seconds) shouldBe true
         dbHandler.run(Tables.Transactions ++= transactions).isReadyWithin(5.seconds) shouldBe true
 
-        whenReady(sut.fetchBlocks(Query.empty)) { result =>
+        whenReady(sut.fetchTransactions(Query.empty)) { result =>
```
ivanopagano, line 71: Was this actually a test bug?

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/config/ConseilAppConfigTest.scala
```diff
@@ -81,14 +81,14 @@ class ConseilAppConfigTest extends AnyWordSpec with Matchers with EitherValues w
                                               |    }
                                               |  },
                                               |  {
-                                              |    name: "ethereum"
+                                              |    name: "OpenChain"
```
ivanopagano, line 84: what does this actually change? I can't see what's impacted...

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataOperationsTest.scala
```diff
@@ -0,0 +1,253 @@
...
+      val sut = new EthereumDataOperations("ethereum") {
+        override lazy val dbReadHandle = dbHandler
+      }
+
+      "return proper number of blocks, while fetching all blocks" in {
+        // given
+        dbHandler.run(Tables.Blocks ++= blocks).isReadyWithin(5.seconds) shouldBe true
+
+        whenReady(sut.fetchBlocks(Query.empty)) { result =>
+          result.value.size shouldBe 3
```
ivanopagano, line 42: I'm usually quite picky about this kind of things, so this is also a request for opinions:

Is it better to hardcode the number of expected values, or to reuse the fixture data to compute it, so it remains in sync with the fixture definition, whatever might happen? I.e. I'd choose

```scala
result.value.size shouldBe (blocks.size)
```

And the same holds for all the similar cases following this one

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataRoutesTest.scala
```diff
@@ -0,0 +1,211 @@
...
+
+      val postRequest = HttpRequest(
+        HttpMethods.POST,
+        uri = "/v2/data/ethereum/mainnet/blocks",
+        entity = HttpEntity(MediaTypes.`application/json`, jsonStringRequest)
+      )
+
+      postRequest ~> addHeader("apiKey", "hooman") ~> routes.postRoute ~> check {
+        val resp = entityAs[String]
+        resp.filterNot(_.isWhitespace) shouldBe jsonStringResponse.filterNot(_.isWhitespace)
```
ivanopagano, line 75: I think there is a scalatest plugin lib to compare json strings for match, that we already use.
It's more robust to how the json is formatted, and should check the content instead of the string.

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataRoutesTest.scala
```diff
@@ -0,0 +1,211 @@
...
+        status shouldBe StatusCodes.OK
+      }
+    }
+
+    "return 404 NotFound status code for request for the not supported platform with POST" in {
```
ivanopagano, line 80: Could you kindly replace all the 
"not supported X" with
"unsupported X" in the test descriptions?

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ApiFilterQueryString.scala
ajuszczak: Sure. I'll provide some documentation where possible and sensible :)

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ApiCirceJsonSchema.scala
ajuszczak: `ApiValidation` does that.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataRoutesCreator.scala
ajuszczak: I'm not sure what you are asking for? This trait is combined with Circe and Akka HTTP and indeed `EthereumDataHelpers` is a class. It is a class to cut off this long traits tail at some point but it could be a trait. I probably had different composition previously and did not revert it back.

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/bitcoin/BitcoinDataOperationsTest.scala
ajuszczak: Yes, it was a bug indeed. We have been testing wrong method, but the returned result was pretty much the same.

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/config/ConseilAppConfigTest.scala
ajuszczak: Yea, so once we added integration for `ethereum` this test started to pass, but we expect failure here, so I changed to the blockchain we are not supporting yet ;)

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataOperationsTest.scala
ajuszczak: Yea, I thought that it was better to have what you proposed for couple of years, but then I realized that changing fixtures had to big of an impact on rest of the tests. Meaning that it's wasn't my conscious decision that assertion changed. Since tests are really important and I like when I see that some conditions are changing explicitly in pull-requests (instead of implicit via fixture changes) I think it's better to have them explicitly mentioned. Even, when you are comparing the same stuff (like asking for specific `id` and checking if the same `id` has been returned, I would code them twice).

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataRoutesTest.scala
ajuszczak: I can :)

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataRoutesTest.scala
ajuszczak: Right, you are taking probably about `matchJson` from custom library for scalatest. Find that, so it's added already.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataRoutesCreator.scala
ivanopagano: Sorry but I've never seen a concrete class being subclassed with a trait... I'm scared of possible consequences down the line 😅

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataRoutesCreator.scala
ivanopagano: @ajuszczak what about this? do you plan to do it?

ajuszczak: Forgot about that, sure I can merge that.

ajuszczak: I did checked that and actually this method is used (I mean the second one). Please take a look at how `postRoute` has been created. It is so dynamic, that it requires this specific method, but I guess we can modify that a bit. 

Question is, if we want to do that, just to remove single method - not sure if I won't brake something down the road.

ivanopagano: Probably not worth it, I must've missed the other usage site.

PR #897, opened 2020-08-07
887 -  Lorre for Ethereum

molowny:
Most of the code is based on existing Bitcoin implementation.


conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
```diff
@@ -137,20 +137,44 @@ object Platforms {
     override val platform: BlockchainPlatform = Bitcoin
   }
 
+  /** configurations to describe a ethereum node */
+  final case class EthereumNodeConfiguration(
```
ivanopagano, line 141: Should we define a common data for this config?

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
```diff
@@ -137,20 +137,44 @@ object Platforms {
...
+      hostname: String,
+      port: Int,
+      protocol: String,
+      path: String
+  ) {
+    val url = s"$protocol://$hostname:$port$path"
+  }
+
+  /** configurations to describe a ethereum batch fetch */
+  final case class EthereumBatchFetchConfiguration(
```
ivanopagano, line 151: Could this entries be documented, please?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/ethereum/EthereumIndexer.scala
```diff
@@ -0,0 +1,144 @@
...
+  //       to stop the mainLoop.
+  override def start(): Unit = {
+
+    /**
+      * Repeat [[cats.IO]] after the specified interval.
+      *
+      * @param interval finite duration interval
+      * @param f [[cats.IO]] to repeat
+      */
+    def repeatEvery[A](interval: FiniteDuration)(f: IO[A]): IO[Unit] =
```
ivanopagano, line 79: What  about picking a name for `f` like `code`, `operations`, `process`, ...?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/ethereum/EthereumIndexer.scala
```diff
@@ -0,0 +1,144 @@
...
+              */
+            IO.delay(logger.info(s"Start Lorre for Ethereum")) *>
+              ethereumOperations.loadBlocksAndLogs(lorreConf.depth).compile.drain
+          }
+      )
+      .unsafeRunSync()
+  }
+
+  override def stop(): Future[LorreIndexer.ShutdownComplete] =
+    Future {
```
ivanopagano, line 104: Why don't we go here for an `IO` operation and then `unsafeToFuture`?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/ethereum/EthereumOperations.scala
```diff
@@ -0,0 +1,115 @@
...
+import tech.cryptonomic.conseil.common.rpc.RpcClient
+import tech.cryptonomic.conseil.common.ethereum.EthereumPersistence
+import tech.cryptonomic.conseil.common.ethereum.rpc.EthereumClient
+import tech.cryptonomic.conseil.indexer.config.{Custom, Depth, Everything, Newest}
+
+/**
+  * Ethereum operations for Lorre.
+  *
+  * @param ethereumClient Ethereum client instance
+  * @param persistence Ethereum persistence instance
```
ivanopagano, line 18: Can I suggest to refrain from describing params like this? We might try to explain what each component does.
Otherwise I'd recommend leaving the desc empty

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/ethereum/EthereumOperations.scala
```diff
@@ -0,0 +1,115 @@
...
+    * Start Lorre with mode defined with [[Depth]].
+    *
+    * @param depth Can be: Newest, Everything or Custom
+    */
+  def loadBlocksAndLogs(depth: Depth): Stream[F, Unit] =
+    Stream
+      .eval(tx.transact(persistence.getLatestIndexedBlock))
+      .zip(ethereumClient.getMostRecentBlockNumber.map(Integer.decode))
+      .flatMap {
+        case (block, mostRecentBlockNumber) =>
```
ivanopagano, line 39: Possibly `block` is too generic. Could we use something like `recentIndexedBlock` or similar descriptive name?
Is that useful?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/ethereum/EthereumOperations.scala
```diff
@@ -0,0 +1,115 @@
...
+          val range = depth match {
+            case Newest => block.map(_.number + 1).getOrElse(1) to mostRecentBlockNumber
+            case Everything => 1 to mostRecentBlockNumber
+            case Custom(depth) => (mostRecentBlockNumber - depth) to mostRecentBlockNumber
```
ivanopagano, line 43: Might be easy to check if we're asking too deep a level, here.

- use 1 if depth > mostRecent
- use block.number if higher than mostRecent - depth

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/ethereum/EthereumOperations.scala
```diff
@@ -0,0 +1,115 @@
...
+        existingBlocks =>
+          Stream
+            .range(range.start, range.end)
+            .filter(height => !existingBlocks.contains(height))
+            .map(n => s"0x${n.toHexString}")
+            .through(ethereumClient.getBlockByNumber(batchConf.blocksBatchSize))
+            .through(ethereumClient.getBlockWithTransactions(batchConf.transactionsBatchSize))
+            .evalTap { // log every 10 block
+              case (block, _) if Integer.decode(block.number) % 10 == 0 =>
+                Concurrent[F].delay(logger.info(s"Save block with height: ${block.number}"))
```
ivanopagano, line 71: Not sure if it might be overboard but there's a `ifM` or `ifA` combinator to only execute IO on certain conditions.

sql/conseil.sql
```diff
@@ -918,7 +918,7 @@ ALTER TABLE ONLY ethereum.transactions
 
 -- Table is based on eth_getLogs from https://eth.wiki/json-rpc/API
 CREATE TABLE ethereum.logs (
-  address text NOT NULL PRIMARY KEY,
+  address text NOT NULL,
```
ivanopagano, line 921: Is this a shortcoming or does it have specific reason for not being unique?

molowny: There can be multiple log entries corresponding to one address, this is my mistake.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
molowny: do you mean some trait?

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
molowny: Please check it now

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/ethereum/EthereumOperations.scala
molowny: Please check it now

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/ethereum/EthereumOperations.scala
molowny: I leave it as it is, I don't think the code is more readable with `if*`

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
ivanopagano: No, just that it's not chain dependent... it just describes any internet resource path on a host with port

Why having different classes if they all essentially describe an rpc endpoint?

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/ethereum/EthereumOperations.scala
ivanopagano: I'm not sure what you did cover the cases I had in mind, that is:
 1. passed depth is higher than the mostRecent height on the chain: start from 1 to mostRecent
 2. same as before, but instead of starting from the first block, we might want to behave as if Newest was passed, so we start from latestIndexedBlock number + 1, if available

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/ethereum/EthereumOperations.scala
ivanopagano: awesome

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
molowny: good point, I've changed it to the `java.net.URL` I was influenced by the Tezos code :P

ivanopagano: Not exactly what I meant, but... 😄 
Hell, this is so simpler, why not?

I'm not sure how frequent you might want to only change a single element of the endpoint, so for now I'll keep it like that.

P.s. to be clear, I meant that the whole case class could be put in common between all the chains, since it didn't specify anything chain-specific

molowny: ok :) ok, now I don't know if I did it right :) we'll see, I'll change it to case class if there are any problems in the future

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/ethereum/EthereumOperationsTest.scala
```diff
@@ -0,0 +1,143 @@
...
+import tech.cryptonomic.conseil.indexer.config.{Custom, Everything, Newest}
+
+class EthereumOperationsTest extends AnyWordSpec with MockFactory with Matchers {
+
+  implicit val contextShift: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
+
+  "Ethereum operations" should {
+      "run indexer for the newest blocks" in new EthereumOperationsStubs {
+        // Ethereum operations with mocked loadBlocksWithTransactions to test only loadBlocks.
+        val ethereumOperationsStub = new EthereumOperationsMock {
```
ivanopagano, line 25: This naming is counter-intuitive to me.
What's a mock and what's a stub?
Why a stub is implemented via a mock?
And why is the mock not using the library to be a full mock and not some hybrid implementation.

I think we should get feedback from more devs than me on this, since I'm openly biased against this type of testing.
For more context, I see no value in checking that internally hidden components of the class under test are called with the proper inputs and provide back outputs that needs to lock correctly to produce the expected overall outcome.

I think a system should be checked at the outer layer, and not by exploiting internal knowledge of its sub-parts.
This violates hiding principles and the inversion of control on dependencies because now you program against implementations and not interfaces.

I find the overall approach extremely brittle, the mock frameworks quite limited and hard to satisfy when I need to do more complicated scenarios.
In the past I've spent quite a lot of time to correct a failing mock test only because I needed to add internal details to a system, not directly related to the outcome that were being verified. And it took me a lot to figure out how to bypass the limitations of the mock dsl.
This is also why I feel quite uncomfortable having these tests in my projects.

I'm open to debate, but we had this conversation in the past, and I'm willing to accept the choice as long as I'm the only one complaining.

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/ethereum/EthereumOperationsTest.scala
```diff
@@ -0,0 +1,143 @@
...
+      httpFetchThreadsCount = 1,
+      blocksBatchSize = 1,
+      transactionsBatchSize = 1,
+      logsBatchSize = 1
+    )
+
+    class EthereumClientMock(rpcClient: RpcClient[IO]) extends EthereumClient[IO](rpcClient)
+    class EthereumPersistenceMock extends EthereumPersistence[IO]
+
+    val ethereumClientMock = mock[EthereumClientMock]
```
ivanopagano, line 111: I fail to see the value of having a mock over a mock instead of a mock of the real thing? 🤔

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
ivanopagano: yes, I think we can always revert back to individual pieces if ever needed. I personally doubt that.

conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/ethereum/EthereumOperationsTest.scala
molowny: After consideration I understand your point of view, actually, I've created this spec only to fulfill the test coverage policy. So the simplest solution is to remove it and handle integration/acceptance tests in a smoke test.

ivanopagano: Thank you for taking that into account

ivanopagano (approved):

ajuszczak (approved):

piotrkosecki (approved):
Looks good!

PR #906, opened 2020-09-04
887 - Updated Ethereum endpoints

molowny:
Conseil API with new Ethereum's entities.


ajuszczak (approved):

conseil-api/src/main/resources/metadata/ethereum.mainnet.conf
```diff
@@ -145,5 +145,120 @@
         }
       }
     }
+    recipts {
```
ivanopagano, line 148: is it correctly spelled? shouldn't be receipts?

molowny: I've fixed all occurrences in both PRs. What a shame :)

piotrkosecki (approved):

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataRoutesCreator.scala
```diff
@@ -99,14 +99,77 @@ trait EthereumDataRoutesCreator
...
+  }
+
+  /** V2 Route implementation for accounts by address endpoint */
+  private val accountsByAddressRoute: Route = ethereumAccountByAddressEndpoint.implementedByAsync {
+    case ((network, address), _) =>
+      platformNetworkValidation(network) {
+        operations.fetchAccountByAddress(address)
+      }
+  }
+
```
ivanopagano, line 157: More of a philosophical question than anything else... I look at this code and think...
what actually changes between the calls...? Why do we copy paste it? Is it clearer? Is it better?

Or is a single abstraction called `delegateToOperations(operationCall: in => out)` clearer and cleaner?

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataOperationsTest.scala
```diff
@@ -94,6 +94,79 @@ class EthereumDataOperationsTest
...
+      "return proper account by address" in {
+        // given
+        dbHandler.run(Tables.Blocks ++= blocks).isReadyWithin(5.seconds) shouldBe true
+        dbHandler.run(Tables.Transactions ++= transactions).isReadyWithin(5.seconds) shouldBe true
+
+        whenReady(sut.fetchAccountByAddress("to")) { result =>
+          result.value should (contain key "address" and contain value transaction1.to)
+          //TODO Fix test for 'value' field in AccountRow
+          //Issue with the current approach is that type inside case class is from 'scala' while slick is mapping data inside 'QueryResponse' to 'java'.
+          //Additionally BigDecimals are not easy comparable with 'scalatest'.
```
ivanopagano, line 162: Why exactly is not easy to compare BigDecimals? Maybe we can find a solution together?
Or is it just that we didn't yet decide what to use to describe the result, and this will be fixed then?

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataOperationsTest.scala
```diff
@@ -94,6 +94,79 @@ class EthereumDataOperationsTest
...
+
+      "return proper account by address" in {
+        // given
+        dbHandler.run(Tables.Blocks ++= blocks).isReadyWithin(5.seconds) shouldBe true
+        dbHandler.run(Tables.Transactions ++= transactions).isReadyWithin(5.seconds) shouldBe true
+
+        whenReady(sut.fetchAccountByAddress("to")) { result =>
+          result.value should (contain key "address" and contain value transaction1.to)
+          //TODO Fix test for 'value' field in AccountRow
+          //Issue with the current approach is that type inside case class is from 'scala' while slick is mapping data inside 'QueryResponse' to 'java'.
```
ivanopagano, line 161: Hmmm, what kind of object is returned by the slick response? And exactly why is it a java one?

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataOperationsTest.scala
```diff
@@ -249,5 +322,121 @@ object EthereumDataOperationsTest {
       transactionIndex = transaction3.transactionIndex
     )
     val logs: Seq[LogsRow] = List(log1, log2, log3)
+
+    val receipt1: ReceiptsRow = ReceiptsRow(
```
ivanopagano, line 326: Might be there is an easy way to factor out this repetition in a easy-to-understand smart-constructor method?
Or even use a single default object and use `.copy` to customize it in the test depending on what we need each time?

ivanopagano (commented):
Left comments for improvements and clarification.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataRoutesCreator.scala
ajuszczak: Well... It is not so easy, mostly because different endpoints takes (or produces) different arguments. Sometimes you have an address, sometimes you have a filter, but maybe that can be extracted (I'm not sure if that is possible with this weird types in Endpoints4s - to find generic enough). 

Additionally, I see only one think that could be extracted, which is `platformNetworkValidation`. If you look at the documentation from Endpoints4s ([Authentication part](https://endpoints4s.github.io/guides/custom-authentication.html) to be specific), you will see there is a generic way of checking API-Key. I believe there is possibility to do the same thing but with `platformNetworkValidation` - to extract this from this place and implement it only once, cause `network` is all you need and you will have it always.

This being done properly, will reduce number of parameters here (apiKey that we have here which is mapped to `_` could be also removed). We could also get rid of this `platformNetworkValidation` and the only think left would be
`ethereumAccountByAddressEndpoint.implementedByAsync(address => operations.fetchAccountByAddress(address))` which is an essence of this route. You cannot reduce it more :) 

So in short we could try do, what you want, but that would be middle step. The change I'm talking about is a bit more complex tho, and can take couple of days to implement.

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataOperationsTest.scala
ajuszczak: I believe mostly because scaliest does not provide any support for comparing this types, but I remember that the fact that it was sometimes BigDecimal from scala package, sometimes from java package was painful in this test case, but since I have some time I can take a look at this with fresh mind and investigate that more.

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataRoutesCreator.scala
ivanopagano: I think it might be worth it, but the end call's yours

ajuszczak: It might, be we are talking about re-designing 40% of the current API's implementation :D

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataOperationsTest.scala
ivanopagano: ow, the fact that it's changing between java/scala version is a nuisance, I have to admit...

What about extracting a comparable value from the big-decimal, like a asLong, considering that tests might be controlled to not overflow it?
Extraction could be done with pattern matching to handle both java and scala BD types...

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataRoutesCreator.scala
ivanopagano: 😮 ... I guess we have a clear choice then... 😬

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataOperationsTest.scala
ajuszczak: So I did investigated why we need to deal with `java.math.BigDecimal` instead of `scala.math.BigDecimal`.

Everything starts here: `tech.cryptonomic.conseil.common.util.DatabaseUtil.QueryBuilder#getMap`.
To access the data, while we are making generic query, we are using above `GetResult` implicit created only for this purpose. 

We are calling `positionedResult.nextObjectOption` in there. Looking into its implementation, you will see that this method is calling:
```
final def nextObjectOption()     = { val npos = pos + 1; val r = rs getObject     npos; val rr = (if(rs.wasNull) None else Some(r)); pos = npos; rr }
```

It works differently for `nextBigDecimalOption` tho:
```
final def nextBigDecimalOption() = { val npos = pos + 1; val r = rs getBigDecimal npos; val rr = (if(rs.wasNull) None else Some(BigDecimal(r))); pos = npos; rr }
```
So here implementation is wrapping `java.math.BigDecimal` into `scala.math.BigDecimal`, but we are not using this mechanism. Let's skip it for a while.

Gladly, Scala's implementation is just a wrapper, so we can call `.bigDecimal` and it should work like that:
```
result.value should (contain key "value" and contain value Some((transaction2.value + transaction3.value).bigDecimal))
```
... right? No... it's not going to work, because the scale for `BigDecimal` from SQL Query is `2`, while in our case it is `0`. You may ask, how the hell BigDecimals are comparable then? They are but not with `==`, which is used by default by `ScalaTest` while dealing with Collections/Maps. We should use `compareTo == 0` for that purpose.

Since we are not able to do that, my suggestion is to simply adjust the scale for this purpose to `2` and wrap that with comment, why we are doing that.

conseil-api/src/test/scala/tech/cryptonomic/conseil/api/routes/platform/data/ethereum/EthereumDataOperationsTest.scala
ajuszczak: I decided to use default object + copy. Mostly because in previous company I was working with object builders and there was 2x more code that we have now and we also had to test it properly. I believe, I managed to reduce the code and complexity.

ivanopagano (approved):
Looks great

Issue #870, opened 2020-06-30
The indexer is terribly slow due to fees processing


ivanopagano:
As shown by the query engine the way fees are collected takes a long time. 
The process is run once every N cycles of Lorre and loads past fees.

Analysis:
As the number of fees grows the indexing to collect `distinct` values becomes useless and the engine is forced to do a linear scan.
The query is: `select distinct "fee", "timestamp", "cycle", "level" from "tezos"."operations" where "kind" = 'transaction' order by "timestamp" desc limit 1000`
>the longest amount of time is being taken up by the sort and the subsequent sequential scan, please keep in mind that  kind is already an index
>plan is limit(.086ms) -> Unique(2.406 ms) -> sort(12,105ms) -> seq_scan(11,221ms)

Proposed solutions:
Put a lower bound on the query timestamp to reduce the number of rows returned for the computation when doing the `distinct` operation.
Forego the `distinct` requirement, if it's not necessary (why should it happen that we collect more than one unique fee?)

PR #907, opened 2020-09-07
870 - Improve fees computation query performance

ivanopagano:
Solves #870 

Now use shared assumptions on number of transactions per block to limit the number of results to average upon, before sorting the timestamps. This is done by a lower bound on the timestamps themselves.

To make it clear:
- we now use a time-window to decide how many operations are used to do the statistics computation
- this allows the db to avoid full-scanning the table before filtering out unneeded values, because the filter is done upfront

Overall this should improve the optimized query-plan and provide the same level of accuracy as before

## Configuration changes

The internal indexer `application.conf` now expresses fees computation windows in term of days, instead of number of fees

from 
```
  number-of-fees-averaged: 1000	
```
to
```
  fees-average-time-window: 1 day
```

This _should not_ impact deployed environments, unless this configuration entry is overridden via ENV-VARS or docker env


piotrkosecki (approved):

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
```diff
@@ -761,11 +761,41 @@ object TezosDatabaseOperations extends LazyLogging {
...
+     *
+     * This is what we know now:
+     * - each cycle bakes 4096 blocks
+     * - a cycle takes around 2-3 days to run
+     * - each block stores a variable number of transactions, down-to the min of 1.
+     *
+     * We can make conservative computations to figure out how far in the past we need to go,
+     * to guarantee a [[numberOfFeesAveraged]] values.
+     *
+     * We can assume a single transaction per block (1 trans/block), hence we need numberOfFeesAveraged blocks.
```
vishakh, line 786: Assuming a single transaction per block will not be sound in the long term. Is it possible to either:

1. Move the calculation completely to the database?
2. Simply change the function so it takes a time range rather than number of transactions?

vishakh (changes requested):

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
ivanopagano: Let me first put down a couple of details that may help us:

1. for our current fees averaging count configured to 1000, we don't need to get back past 1 day in the worst case
2. Assuming 1 transaction/block means handling the worst possible case, that is, the longest possible time-window to get all the required transaction needed for averaging.

The latter is to say that blocks with a lot more transactions could impact performance negatively, for sure, but never as much as when doing a table scan (required to filter out based on the outcome of the sort operation).
Once we've cut out most results due to the time-window, I'd assume that the db will be more than able to handle a few million records?
This is only theoretical, so we might want to add measurements to this code and have timing results based on the number of records, and decide on those.

That said, we can move the averaging to the db nonetheless.
Also doing the average on time-range is also reasonable, but I don't know if it's still sensible from the "business" perspective, I lack the perspective on that.

Give me some feedback, please based on the additional information @vishakh

vishakh: Your reasoning makes sense to me. However, we effectively end up with a time window. Won't it be simpler to just explicitly make it a time window and use your heuristic to set the value in a config?

ivanopagano: I see the point, then this is what I plan to do:

1. switch to time-windowing in config (say how many days past should the averaging go)
2. set the config to 1 day for now
3. see how hard it is to compute the averaging directly as part of the db-query

vishakh: Sounds good to me!

ivanopagano: Part 1 and 2 are done

Looking at 3

ivanopagano: Completed all the points

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosFeeOperations.scala
```diff
@@ -26,17 +26,21 @@ private[tezos] object TezosFeeOperations extends LazyLogging {
     "endorsement"
   )
 
-  /**
-    * Calculates average fees for each operation kind and stores them into a fees table.
+  /** Calculates average fees for each operation kind and stores them into a fees table.
+    * The computation will use a limited number of fees, as the result of a selection window in days.
+    * Only blocks belonging within such window in the past, relative to the calculation moment, will be considered.
+    *
     * @param numberOfFeesAveraged a limit on how many of the latest fee values will be used for averaging
```
vishakh, line 33: This param is now dead.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosFeeOperations.scala
```diff
@@ -26,17 +26,21 @@ private[tezos] object TezosFeeOperations extends LazyLogging {
...
-    * @param ex the needed ExecutionContext to combine multiple database operations
+    * @param selectionWindow the max number of days back from the current block timestamp to use when averaging
```
vishakh, line 34: The parameter type seems to take an arbitrary duration but you seem to indicate here that it's always in days.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosFeeOperations.scala
```diff
@@ -26,17 +26,21 @@ private[tezos] object TezosFeeOperations extends LazyLogging {
...
     * @return a future result of the number of rows stored to db, if supported by the driver
     */
-  def processTezosAverageFees(numberOfFeesAveraged: Int)(implicit ex: ExecutionContext): Future[Option[Int]] = {
+  def processTezosAverageFees(selectionWindow: FiniteDuration)(implicit ec: ExecutionContext): Future[Option[Int]] = {
     logger.info("Processing latest Tezos fee data...")
+
     val computeAndStore = for {
-      fees <- DBIOAction.sequence(operationKinds.map(TezosDb.calculateAverageFees(_, numberOfFeesAveraged)))
-      dbWrites <- TezosDb.writeFees(fees.collect { case Some(fee) => fee })
+      feeStats <- TezosDb.FeesStatistics.calculateAverage(selectionWindow.toDays)
```
vishakh, line 42: Why not keep it as an arbitrary duration? It will be more general that way.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
```diff
@@ -21,12 +21,11 @@ import tech.cryptonomic.conseil.common.tezos.Tables
 import tech.cryptonomic.conseil.common.util.ConfigUtil
 import tech.cryptonomic.conseil.common.util.CollectionOps._
 import tech.cryptonomic.conseil.common.util.Conversion.Syntax._
-import tech.cryptonomic.conseil.common.util.MathUtil.{mean, stdev}
 import tech.cryptonomic.conseil.indexer.sql.DefaultDatabaseOperations._
 
 import scala.collection.immutable.Queue
 import scala.concurrent.{ExecutionContext, Future}
-import scala.math.{ceil, max}
+import scala.math
```
vishakh, line 28: Unused import

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosFeeOperations.scala
ivanopagano: Makes things more complex, but not too much.
Yet I wonder if that's worth it right now. It could be tuned later quite easily to afford more complex definitions.

For the moment the level of granularity in the definition amounts to a whole day.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosFeeOperations.scala
ivanopagano: will remove

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosFeeOperations.scala
ivanopagano: As explained below, the input is then approximated up to the day

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
ivanopagano: will remove

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosFeeOperations.scala
vishakh: Makes sense.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosDatabaseOperations.scala
ivanopagano: actually used at lines `778/779`

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosFeeOperations.scala
ivanopagano: done

vishakh (approved):

piotrkosecki (approved):

PR #925, opened 2020-09-30
922 - missing rolls and bakers during test period

### LINKED ISSUE #922: Fix incorrect baker data in accounts
Problem: The Tezos RPC call `../votes/listings` is used to fetch the number of baker rolls and, ultimately, set baker status in the accounts tables. This RPC call returns no results for certain phases of governance like testing. This might have always been a bug or triggered by some relatively recent change in node behavior. Due to this, the accounts table shows incorrect data for `is_baker` in the accounts table.

Fixes:
- [x] Change the logic for setting `is_baker` in the accounts tables to check the `bakers` table for a valid record.
- [x] To set the 'rolls' value in the bakers tables, diving the 'staking_balance' field by a constant configured to 8,000.

Validation:
- [ ] For any well-known baker, there should be valid data in the accounts tables through all four cycles of governance.
- [ ] The new `rolls` values should match the value returned by the `../votes/listings` RPC route for the proposal governance period.

ivanopagano:
This interim solution closes #922 

It doesn't use rolls to identify bakers, but reads existing bakers from the corresponding database table.
It also simplifies rolls computation with a simple rule based on staking balances.

There are also a few refactors useful to better separate dependencies between the processor component for accounts and the database instance.
This is done by re-use of the indexed-data operations component, which collects read operations on data already stored in the indexer. Eventually I might try to re-use this separation for other entity processors too.
Finally the metadata operations related traits where renamed to better reflect the actual api exposed, which is related to running slick-db actions through a database instance. Hence the rename and reuse of such interface.

### DIFF
--- a/conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/TezosTypes.scala
+++ b/conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/TezosTypes.scala
@@ -522,12 +522,8 @@ object TezosTypes {
   ) {
 
     /** Compute rolls based on staked balance */
-    def rolls: Option[Int] = staking_balance match {
-      case PositiveDecimal(value) =>
-        Some((value quot Voting.BakerRollsSize).toIntExact)
-      case InvalidPositiveDecimal(jsonString) =>
-        None
-    }
+    def rolls: Option[Int] = Baking.computeRollsFromStakes(staking_balance)
+
   }
 
   final case class CycleBalance(
@@ -603,14 +599,28 @@ object TezosTypes {
 
   object Voting {
 
-    /** how big is a baker roll */
-    final val BakerRollsSize = BigDecimal.decimal(8000)
-
     final case class Vote(value: String) extends AnyVal
     final case class Proposal(protocols: List[(ProtocolId, ProposalSupporters)], block: Block)
     final case class BakerRolls(pkh: PublicKeyHash, rolls: Int)
     final case class Ballot(pkh: PublicKeyHash, ballot: Vote)
     final case class BallotCounts(yay: Int, nay: Int, pass: Int)
+  }
+
+  /** Utilities related to the baking process */
+  object Baking {
+
+    //we'll move this to a proper chain configuration value later on
+    /** how big is a baker roll */
+    final val BakerRollsSize = BigDecimal.decimal(8000)
+
+    def computeRollsFromStakes(stakingBalance: PositiveBigNumber): Option[Int] =
+      stakingBalance match {
+        case PositiveDecimal(value) =>
+          Some((value quot BakerRollsSize).toIntExact)
+        case InvalidPositiveDecimal(jsonString) =>
+          None
+      }
+
   }
 
   object Syntax {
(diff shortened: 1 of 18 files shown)


conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BakersProcessor.scala
```diff
@@ -231,6 +205,14 @@ class BakersProcessor(
       case InvalidPositiveDecimal(_) => None
     }
 
+    /** Computes updated rolls from the stake balance being updated */
+    def rollsFromStakes(balance: PositiveBigNumber): Int = balance match {
```
vishakh, line 209: How come the same logic is replicated in two places, i.e. here as well as `TezosTypes`? Might caught confusion and bugs later.

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/TezosNodeOperator.scala
```diff
@@ -21,6 +21,29 @@ private[tezos] object TezosNodeOperator {
   type LazyPages[T] = Iterator[Future[T]]
   type Paginated[T, N] = (LazyPages[T], N)
 
+  /** Process every single page of results by converting the content with a custom function.
+    * The returned value will be the content of the paginated results.
+    */
+  def mapByPage[T](pages: LazyPages[T])(pageMapper: T => T)(implicit ec: ExecutionContext): LazyPages[T] =
```
vishakh, line 27: Doesn't look like this function is being used anywhere. Is this left here for future use?

ivanopagano: Indeed, to me it looked like a missing piece and without it the other felt very ad-hoc... but in the end we could remove it.

Actually I'd need some time to check if we could put it to use in some place already, but possibly in a separate task work. I would make a separate PR to use it or remove it, based on the outcome of a quick survey

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/TezosTypes.scala
```diff
@@ -597,6 +603,9 @@ object TezosTypes {
 
   object Voting {
 
+    /** how big is a baker roll */
+    final val BakerRollsSize = BigDecimal.decimal(8000)
```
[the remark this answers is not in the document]
ivanopagano: Totally agree, the goal here was to quickly patch

conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/tezos/processing/BakersProcessor.scala
ivanopagano: I've spotted this additional change late evening, and wanted something quick up for you all to review.

I'll try to clean up now

ivanopagano: Easier than I thought

piotrkosecki (approved):

vishakh (approved):

PR #929, opened 2020-10-02
Bugfix - NC configuration change

piotrkosecki:
after refactor/adding support for other blockchains the configuration for NC was omitted. This should fix the issue.


ivanopagano (approved):
The changes look fine.
Only two things to check:

* Why does this solve the issue with multiple chains?
* Please document the needed config changes in the issue description, as a heads-up for our infra deployments, if any

vishakh (approved):

PR #931, opened 2020-10-06
922-2 mark bakers after delegates are fetched

### LINKED ISSUE #922: Fix incorrect baker data in accounts

ivanopagano:
Should fix #922 

This version only match accounts for the bakers table after the whole blocks page is being processed, so that bakers from the same batch will figure when checking accounts.

In any case we look for any new delegate in each blocks page fetched. This allows to mark late delegate accounts that were previously originated but never used as delegates.

This PR introduce a correction for Rolls computation, due to previously incorrect scaling of rolls units of measures.

As a minor addition, we introduce a feature flag driven by the Env-var `CONSEIL_LORRE_BLOCK_RIGHTS_FETCHING_ENABLED`.
It allows to run the tezos indexer without computing future baking/endorsing rights, which comes useful for testing purposes only, to reduce noise when the feature under test is unrelated to future rights.

### DIFF
--- a/conseil-lorre/src/main/resources/application.conf
+++ b/conseil-lorre/src/main/resources/application.conf
@@ -93,6 +93,8 @@ lorre {
     cycle-size: 4096 # size of the cycle, by default 4096
     fetch-size: 200 # amount of rights we fetch at once
     update-size: 16 # amount of rights we update after Lorre syncs
+    enabled: true # set to false to turn this feature off
+    enabled: ${?CONSEIL_LORRE_BLOCK_RIGHTS_FETCHING_ENABLED}
   }
 
 }
--- a/conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/TezosTypes.scala
+++ b/conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/TezosTypes.scala
@@ -610,13 +610,14 @@ object TezosTypes {
   object Baking {
 
     //we'll move this to a proper chain configuration value later on
-    /** how big is a baker roll */
+    /** how big is a baker roll in tez */
     final val BakerRollsSize = BigDecimal.decimal(8000)
 
+    /** Starting from the staking balance, infers number of rolls for a given baker. */
     def computeRollsFromStakes(stakingBalance: PositiveBigNumber): Option[Int] =
       stakingBalance match {
         case PositiveDecimal(value) =>
-          Some((value quot BakerRollsSize).toIntExact)
+          Some((value quot (BakerRollsSize * BigDecimal(1000))).toIntExact)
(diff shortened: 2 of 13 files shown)


ivanopagano:
Drafted until I can add a unit-test to verify correct rolls computation from staking balances

conseil-lorre/src/main/resources/application.conf
```diff
@@ -93,6 +93,8 @@ lorre {
     cycle-size: 4096 # size of the cycle, by default 4096
     fetch-size: 200 # amount of rights we fetch at once
     update-size: 16 # amount of rights we update after Lorre syncs
+    enabled: true # set to false to turn this feature off
+    enabled: ${?CONSEIL_LORRE_BLOCK_RIGHTS_FETCHING_ENABLED}
```
piotrkosecki, line 97: can you put it also in the `./docker/docker-compose.yml`?

piotrkosecki (approved):
One comment, but overall looks good.

conseil-lorre/src/main/resources/application.conf
ivanopagano: Makes sense

vishakh (approved):

conseil-lorre/src/main/resources/application.conf
ivanopagano: @vishakh I'm not sure how to consider this: is that an internal debugging flag, or a fully public feature flag?

It makes sense to make it public, if we add even more feature-flags to be disabled. Yet I wonder if they make sense for any production, since the data integrity would be possibly compromised by removing arbitrary features like this.

My feeling would tell me to add to the wiki, but under a "debug" section? Making it clear that disabling this and any other feature-flag we add in the near future could compromise any production data.

vishakh:
Yep, makes sense!

conseil-lorre/src/main/resources/application.conf
ivanopagano: Did that

PR #939, opened 2020-10-16
Added new baking/endorsing rights indexes

Taaseen-Ali:
Baking rights
- `cycle` 
- `cycle, delegate, priority`

Endorsing rights
- `delegate, level`


vishakh (approved):

ivanopagano (approved):
I see nothing wrong, do you know what would be the defaults for sort and null handling if not specified?

PR #944, opened 2020-11-12
Add integration tests to Conseil

PrathamRawat:
Develop integration tests for Conseil

This runs a script to test different (Tezos for now) endpoints on Conseil to ensure that everything is working as is known. It does this in Travis, building Conseil from source, and running the API and Lorre for a fixed length of time, before sending requests to the node. This should ensure that in any fix/update, no existing functionality in broken.

If some core behavior has changed, it is expected that the script's Constants should be updated to ensure that the script is up to date for future versions.


conseil-api-tests/src/main/scala/tech/cyptonomic/conseil/api/tests/ApiTestRun.scala
```diff
@@ -0,0 +1,289 @@
...
+
+        val queryString: String =
+          """
+            |{
+            |     "field": ["%FIELD%"],
+            |     "predicates": [
+            |         {
+            |             "field": "%FIELD%",
+            |             "operation": "isnull",
+            |             "set": [""],
```
piotrkosecki, line 225: cosmetics: set can be an empty list `[]` for `isnull` operation.

conseil-api-tests/src/main/scala/tech/cyptonomic/conseil/api/tests/ApiTestRun.scala
```diff
@@ -0,0 +1,289 @@
...
+
+  /**
+   * Validate a query output by searching for JSON keys
+   * @param key The json key to search for
+   * @param result The result of a query to search in
+   * @param notNull Whether the key's value should be allowed to be null or not
+   */
+  def validateJsonByKey(key: String, result: String, notNull: Boolean = false): Unit = {
+
+    if(result.compareTo("") == 0) throw new NullPointerException("Query Returned an Empty String")
```
piotrkosecki, line 274: why not just `result == ""`?

conseil-api-tests/src/main/scala/tech/cyptonomic/conseil/api/tests/ApiTestRun.scala
```diff
@@ -0,0 +1,289 @@
...
+      `Content-Type`(MediaType.application.json),
+      Accept(MediaType.application.json),
+      Header("apiKey", CONSEIl_API_KEY)
+    )
+
+    val result: String = client.resource.use { client =>
+      client.expect[String](conseilRequest)
+    }.unsafeRunSync()
+
+    if(validationItems.nonEmpty) validationItems.foreach(item => {
```
piotrkosecki, line 117: you can write `validationItems.foreach(item => { ` without using `(` - similarly in other cases below

conseil-api-tests/src/main/scala/tech/cyptonomic/conseil/api/tests/ApiTestRun.scala
```diff
@@ -0,0 +1,289 @@
...
+   * @param key The json key to search for
+   * @param result The result of a query to search in
+   * @param notNull Whether the key's value should be allowed to be null or not
+   */
+  def validateJsonByKey(key: String, result: String, notNull: Boolean = false): Unit = {
+
+    if(result.compareTo("") == 0) throw new NullPointerException("Query Returned an Empty String")
+
+    parse(result) match {
+      case Left(failure) => throw new IllegalArgumentException("Query did not return valid JSON")
```
piotrkosecki, line 277: `failure` is not used an can be replaced with `_`

conseil-api-tests/src/main/scala/tech/cyptonomic/conseil/api/tests/ApiTestRun.scala
```diff
@@ -0,0 +1,289 @@
...
+            |     "limit": 1
+            |}
+            |""".stripMargin.replace("%FIELD%", attribute)
+
+        println("\n\n\nValidating Tezos " + entity._1 + " " + attribute + " Data\n")
+        var result: String = ""
+        try {
+          result = sendConseilQuery(httpClient, CONSEIL.withPath(Requests.getTezosQueryPath(entity._1)), queryString)
+        } catch {
+          case x: org.http4s.client.UnexpectedStatus => {
```
piotrkosecki, line 240: `x` is not used an can be replaced with `_`

conseil-api-tests/src/main/scala/tech/cyptonomic/conseil/api/tests/ApiTestRun.scala
PrathamRawat: Oh I thought like Java, string comparison might only be properly done through comparison methods

conseil-api-tests/src/main/scala/tech/cyptonomic/conseil/api/tests/ApiTestRun.scala
```diff
@@ -0,0 +1,289 @@
...
+            |     ],
+            |     "orderBy": [],
+            |     "aggregation": [],
+            |     "limit": 1
+            |}
+            |""".stripMargin.replace("%FIELD%", attribute)
+
+        println("\n\n\nValidating Tezos " + entity._1 + " " + attribute + " Data\n")
+        var result: String = ""
+        try {
```
piotrkosecki, line 237: I don't like try/catch here and in other cases. I would rather see handling errors with `.handleErrorWith(...)` method on `IO` and making calls asynchronous.

PrathamRawat: Why should calls be asynchronous?

piotrkosecki: Basically when you are using any Future/IO in Scala it is preferred to compose those effects and use `unsafeRunSync`/`unsafeRunAsync` just once in the main object.

vishakh:
Stale but we should revisit in the future.

PR #948, opened 2020-11-25
upgrade to latest 0.x endpoints version

ivanopagano:
This Pr upgrades several dependencies to remove binary-incompatibilities which results from including, e.g. different logging library to the project.

Essentially we need to bump to more recent versions to avoid getting stuck and unable to add newer libraries.

This impacts mostly the endpoints library which is updated to the latest 0.x branch version available.
The project has migrated as endpoints4s in the 1.x incarnation, but we didn't go that far.

This cascades to including more recent akka-http versions and to adapt the conseil-api models to the latest library framing.


piotrkosecki (approved):
I want to say, that this PR was extremely helpful with my migration of the NC, and basically reviewed it in the meantime. Great job 👍

vishakh (approved):

PR #949, opened 2020-11-29
942 - switch logging to scribe library

### LINKED ISSUE #942: Conseil logging needs improvements
Latest testing on a local node lead to an app failure which stopped with little to no information to figure out the cause.

Here is the output
```
15:13:35.809 [run-main-0]                        INFO  t.c.c.indexer.tezos.TezosIndexer     [?:?] - Successfully made initial connection to Tezos
15:13:36.065 [run-main-0]                        INFO  t.cryptonomic.conseil.indexer.Lorre$ [?:?] - Doing clean-up.
15:13:36.066 [run-main-0]                        INFO  com.zaxxer.hikari.HikariDataSource   [?:?] - lorre.db - Shutdown initiated...
15:13:36.075 [run-main-0]                        INFO  com.zaxxer.hikari.HikariDataSource   [?:?] - lorre.db - Shutdown completed.
15:13:36.131 [-akka.actor.default-dispatcher-4]  INFO  RightsFetcher                        [?:?] - Fetching future baking and endorsing rights
15:13:36.177 [run-main-0]                        INFO  t.cryptonomic.conseil.indexer.Lorre$ [?:?] - All things closed.
[error] (run-main-0) io.circe.ParsingFailure: expected json value got 'Failed...' (line 1, column 1)
[error] io.circe.ParsingFailure: expected json value got 'Failed...' (line 1, column 1)
[error] Nonzero exit code: 1
[error] (lorre / runTask) Nonzero exit code: 1
[error] Total time: 4 s, completed Oct 22, 2020, 3:13:36 PM
[warn] Thread[Logback shutdown hook [default],5,run-main-group-0] loading ch.qos.logback.core.LifeCycleManager after test or run has completed. This is a likely resource leak
```

Here we can see a few improvements possible:
 1. the Parsing Failure had no stack trace to use for identifying the actual reason, which is not obvious from the output
 2. the placeholders to show source file and line of the output are empty
 3. (not shown here) there is no immediate way to verify which calls are actually made from conseil to the chain node

I'm gonna elaborate on the above list and possibly open separate sub-issues to find solutions for each

 1. we need to verify if Circe parsing errors have a lazy/missing stack-trace as commonly happen with scala libraries, to improve performance when handling errors with typed approach (e.g. Either[Err, T]). Or if the stack's there but the runtime won't show that upon an uncaught error reaching the main app thread loop
 2. there are a couple of alternatives in scala which provide source-level information to the logs, and try to improve the stack-traces. We might look at a couple of alternatives, and evaluate the impact to the current design (which is based on typesafe logging, and akka logging). Possible alternatives to evaluate
    * [sourcecode library](https://github.com/lihaoyi/sourcecode#logging) which manually allows to extract the info at compile time and add as part of the the output
    * [airframe logging](https://wvlet.org/airframe/docs/airframe-log) a lightweight alternative which includes source code locations and it's based on java.util.logging and compatible with sl4j
 3. There is a way to "probe" every element passing through the akka-stream which makes the batched queries. We can use a flag to turn on this extra log and "trace" the http communication enough to get additional visibility into the possible issues

ivanopagano:
Closes #942 
We change logging library from slf4j+logback to scribe, essentially to gain insight of the source-code level point of logging.

This change also provides a certain flexibility, due to the secondary effect that now loggers can be configured programmatically.
We add a custom module to define the base-ground  logging configuration shared in any conseil module and object.
We also add custom functionality to read configuration from a HOCON file.

Here follows a breakdown of required existing features/capabilities of logging in conseil we want to keep or improve

1. add a logger property to any class by mixin a trait
2. allow compatibility with indirect use (via dependencies) of slf4j loggers
3. configure a standard format across all loggers
4. custom-configure levels for individual loggers, possibly checking in the code when a level is actually enabled (for debugging)
5. enable level configuration through optionally available environment variables
6. change level configuration without need to recompile the code (e.g. restart with a changed config file or env)
7. easily switch off any logging across the whole application
8. support MDC key-values when available
9. support json format, in particular for logstash usage

The scribe library supports all previous points as follows
1. mixin the `scribe.LoggerSupport` trait or import a scribe package to have implicit logging for any class
2. configure loggers with names corresponding to slf4j loggers
3. use a custom made object that, at initialization resets the default root logger with a commonly shared format, which is inherited by default. The format is defined to preserve more or less the legacy output
4. we can get any logger by name/id and reconfigure it (setting several properties) via the .replace() method. We can check if a logging level is enabled with a specific method
5. when reading a custom config we look for a specific property pointing to an env var name. The property is optional
6. our custom object for logging is initialized to read a HOCON configuration that will customize any logging level required, through a pre-defined bespoke schema we made
7. the config file accepts a global "muted" property to shut off any logging. A custom test trait can be mixed-in to switch off any logging during test execution (can be overriden by the individual test at will)
8. scribe reads the MDC and exposes the values in the output format
9. scribe has an extension to send logs to logstash, which we customized to use the same json format either on the console output or calling an external service endpoint

The configuration file format is defined in the docs for the `tech.cryptonomic.conseil.common.io.Logging.Config` class.
Such doc will be integrated into the conseil wiki upon PR being merged.

## Additional logging usage
The Pr introduces an optional `trace-calls` flag on the Platform node configuration. If set to `true` a debug entry will be logged for each http requests/response pair being done via the TezosRpcInterface.
This can be a "temporary" solution to throughly scan how Lorre interacts with the tezos node in case of unexpected failures.

We also explicitly "unroll" the stack-trace upon circe decoding errors, which by default were suppressed.

## shortcomings
Currently I can't seem to override the output identifying the source class of the log entry. This means that the log will always refer to the source code file, even when we actually include it as part of a trait being mixed-in in another specific class.

The log will show as being done from the generic trait, instead of referring to the including class, which is misleading or uninformative at the least.

## Future improvements enabled by this change
The programmatic nature of scribe configuration opens up the possibility to write a bespoke interactive process that could enable dynamic reconfiguration of the log levels, with no need to restart the application process.
This would require some design and should probably be disabled by default.

Scribe also enables handling async stack traces meaningfully, by using a custom execution context to run Futures, which tracks the call stack between the async threads and show it on the log trace of any exception being thrown.
Look [here for more details](https://github.com/outr/scribe/wiki/asynchronous-tracing)


conseil-api/src/main/scala/tech/cryptonomic/conseil/api/Conseil.scala
```diff
@@ -16,11 +16,14 @@ import scala.concurrent.{Await, ExecutionContext, ExecutionContextExecutor}
...
-object Conseil extends App with LazyLogging with ConseilAppConfig with FailFastCirceSupport with ConseilMainOutput {
+object Conseil extends App with ConseilAppConfig with FailFastCirceSupport with ConseilMainOutput {
+
+  Logging.init()
 
   loadApplicationConfiguration(args) match {
     case Left(errors) =>
-    //nothing to do
+      //nothing to do
+      logger.error("There is an error in the provided configuration")
```
piotrkosecki, line 26: can we get errors from `Left` logged here?

piotrkosecki (approved):
left one comment, but overall looks good

conseil-api/src/main/scala/tech/cryptonomic/conseil/api/Conseil.scala
ivanopagano: IIRC, no need to do it here because they've been already logged by the `loadApplicationConfiguration` call...

vishakh (approved):

PR #952, opened 2020-12-01
951 - Remove duplicates in operation results balance updates if part of the operation already

### LINKED ISSUE #951: Duplicate value of balance updates from both operations and their results
image

The image above shows how operations and their results can contain the same rows of balance updates.

We want to eliminate such repetitions

ivanopagano:
Fixes #951 

CAVEAT: We might want to reconsider if keeping duplicates has some specific meaning and we don't want to remove 'em.


piotrkosecki (approved):

vishakh:
@piotrkosecki will make a fresh PR using an updated algorithm for balance updates.

PR #978, opened 2021-03-30
ethereum - field name and types improvements

redroy44:
for transactions.input, when the value is 0x we truncate it

in blocks `number` renamed to `level` field in database
in transactions `from` renamed to `source`, `to` renamed to `destination`, `value` renamed to `amount` field in database

in transactions, fields `index`, `gasprice`, `gas` should be numeric
in blocks, fields `difficulty`, `gaslimit`, `gasused`, `size`, `totaldifficulty`, should be numeric
in logs `transaction_index`, `log_index` are numeric
in transaction_receipt `transaction_index`, `cumulativeGasUsed`, `gasUsed` are numeric


conseil-common/src/main/scala/tech/cryptonomic/conseil/common/ethereum/Views.scala
```diff
@@ -36,11 +36,11 @@ trait Views {
   lazy val AccountsViewSql: SqlAction[Int, NoStream, Effect] = sqlu"""
     CREATE OR REPLACE VIEW ethereum.accounts AS
       SELECT
-        "to" AS address,
-        SUM(value) AS value
+        destination AS address,
+        SUM(amount) AS value
```
piotrkosecki, line 40: maybe `balance` would be better name instead of value - WDYT?

piotrkosecki (approved):
left one comment - beside that, LGTM

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/ethereum/Views.scala
redroy44: I'll change this to `balance` when I merge accounts view, contracts and tokens tables into `accounts` table

anonymoussprocket (approved):

PR #981, opened 2021-03-31
ethereum - fix duplicated contracts and skipping of some data

redroy44:
Fixes duplicates in contracts table when for given block there were multiple contracts created, before it created n^2 rows.

Also fixes dropping rows when some elements were empty ex. with empty logs, contracts creation was ignored.

Saving block is now the final step, which means that if there are some block data that was not fully parsed, that block will be picked up after lorre restart

### DIFF
--- a/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/ethereum/EthereumOperations.scala
+++ b/conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/ethereum/EthereumOperations.scala
@@ -3,7 +3,6 @@ package tech.cryptonomic.conseil.indexer.ethereum
 import cats.effect.{Concurrent, Resource}
 import fs2.Stream
 import slickeffect.Transactor
-
 import tech.cryptonomic.conseil.common.io.Logging.ConseilLogSupport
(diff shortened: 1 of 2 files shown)


conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/ethereum/EthereumOperations.scala
```diff
@@ -87,53 +86,56 @@ class EthereumOperations[F[_]: Concurrent](
...
                   .evalTap(contracts => tx.transact(persistence.createContracts(contracts.toList)))
+                  .map(_ => (block, txs, receipts))
+              case (block, txs, receipts) => Stream.emit((block, txs, receipts))
+
+            }
+            .evalTap { // log every 10 block
+              case (block, txs, receipts) if Integer.decode(block.number) % 10 == 0 =>
+                Concurrent[F].delay(
+                  logger.info(
+                    s"Save block with height: ${block.number} txs: ${txs.size} logs: ${receipts.map(_.logs.size).sum}"
```
piotrkosecki, line 131: small thing: maybe something like `Saving block with level: (...)`?

piotrkosecki (approved):
left one comment, but overall LGTM

PR #990, opened 2021-04-15
fixed micheline -> michelson conversion 

piotrkosecki:
Problem was that in edonet protocol update pairs could represent lists. Our parser handled it, but it did not nest them after conversion.
Simplified example: `{"prim": "pair", "args":[1,2,3]}` we represented as `pair (1 2 3)`, which was incorrect. 
Correct result should look like `pair (1  pair (2 3) )`.

Fixed that and added test to prove it.


anonymoussprocket:
how does this jive with the new "combs" in Edo?

piotrkosecki:
@anonymoussprocket what do you mean by new combs? do you have any example?

PR #994, opened 2021-04-26
Bugfix/more ethereum fixes

redroy44:
This PR contains a number of minor fixes to ethereum indexer:
* expose `tokens` and `contract` views through the api
* expose `bytecode_hash` field in `contracts` view
* properly parse `decimals` and `total_supply` fields of tokens
* fix occasional mixups of token data fields
* properly parse unicode like "🦄"


conseil-common/src/main/scala/tech/cryptonomic/conseil/common/ethereum/EthereumPersistence.scala
```diff
@@ -64,16 +64,20 @@ class EthereumPersistence[F[_]: Concurrent] extends ConseilLogSupport {
...
+  }
 
   /**
     * Create [[DBIO]] seq with account balances.
     *
     * @param accounts account eth_getBalance data
     */
-  def createAccountBalances(accounts: List[Account]): DBIO[Option[Int]] =
-    Tables.AccountsHistory ++= accounts.map(_.convertTo[Tables.AccountsHistoryRow])
+  def createAccountBalances(accounts: List[Account]): DBIO[Option[Int]] = {
```
piotrkosecki, line 77: I'm not sure if its correct doing insertOrUpdateAll on AccountsHistory. This table should contain all of the account states in time, and I don't think we should be doing any updates there or am I missing something?

redroy44: You're right, the problem appeared when within one block there was a contract deployed and there were some transactions on it. Now I'm filtering out contract addresses that were deployed in a block when saving accounts_history data

``
Stream
    .emits(txs)
    .filter(t => List(Some(t.from), t.to).forall(addr => !receipts.map(_.contractAddress).contains(addr)))
```

piotrkosecki (approved):

PR #1006, opened 2021-06-01
Conseil api routes fix

redroy44:
This PR fixes a bug in conseil api where for multiple blockchains enabled, queryEndpoint was always by TezosDataRoutes.postRoute causing failing SQL query to be built for different blockchains.

Fix was to split common queryEndpoint into separate query endpoints


piotrkosecki (approved):
LGTM!

PR #1010, opened 2021-06-08
Bitcoin indexing fixes

redroy44:
This PR introduces a couple of fixes related to indexing bitcoin:
- txid can be duplicated https://github.com/bitcoin/bitcoin/issues/612
- `version` field is sometimes larger that Int

- added missing tables to hasura metadata


piotrkosecki (approved):
LGTM

PR #1013, opened 2021-06-30
Feature/support multiple dbs

redroy44:
This PR introduces support for per-blockchain databases in conseil-api. This will allow to store indexed blockchain data in separate dbs

Note: This PR requires changes in config `platforms` key

## Changes to the .conf files

`reference.conf` introduces a new sub-entry `db` for each platform defined.

This entry allows user to define separate database for each platform defined. The data is still expected to be in the appropriate db namespace.

If the data for all defined platforms is stored in the same database, the sub-entry `db` should have the same value.

```
platforms: [
  {
    name: "tezos"
    network: "alphanet"
    enabled: true,
    node: {
      protocol: "http",
      hostname: "localhost",
      port: 8732
       path-prefix: "tezos/alphanet/"
    }

    db {
      dataSourceClass: "org.postgresql.ds.PGSimpleDataSource"
      properties {
        user: "foo"
        password: "bar"
        url: "jdbc:postgresql://localhost:5432/postgres"
      }
      numThreads: 10
      maxConnections: 10
    }

  }
```


piotrkosecki (approved):
Good job!

PR #1015, opened 2021-07-07
expose json logging via env var

redroy44:
This PR enables json logging for conseil docker container through `CONSEIL_ENABLE_JSON_OUTPUT` env variable

Ex. `CONSEIL_ENABLE_JSON_OUTPUT="true"`

Also http4s log level for lorre was set to INFO to lower log volume

### DIFF
--- a/conseil-api/src/main/resources/logging.conf
+++ b/conseil-api/src/main/resources/logging.conf
@@ -1,6 +1,10 @@
 logging: {
     muted: false
-    output-json-with-service-name: ${?CONSEIL_LOG_JSON_SERVICE_NAME}
+
+    enable-json-output: false
+    enable-json-output: ${?CONSEIL_ENABLE_JSON_OUTPUT}
+
+    output-json-with-service-name: "conseil-api"
     loggers: [
         {
             name: "com.base22",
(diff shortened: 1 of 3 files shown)


conseil-api/src/main/resources/logging.conf
```diff
@@ -1,5 +1,6 @@
 logging: {
     muted: false
+    output-json-with-service-name: ${?CONSEIL_LOG_JSON_SERVICE_NAME}
```
piotrkosecki, line 3: can you put here the default value of this variable? from this I can't tell if it's boolean or something else

conseil-lorre/src/main/resources/logging.conf
```diff
@@ -1,10 +1,15 @@
 logging: {
     muted: false
+    output-json-with-service-name: ${?CONSEIL_LOG_JSON_SERVICE_NAME}
```
piotrkosecki, line 3: the same as above

conseil-api/src/main/resources/logging.conf
redroy44: Hmm, right now setting this to any value will trigger json logging which is not great

I'll add another field `enable-json-output` that will work like a on/off switch, that will make it clearer

piotrkosecki (approved):

PR #1019, opened 2021-07-21
1018 - make registered tokens dynamic

### LINKED ISSUE #1018: Make registered_tokens dynamic
The `registered_tokens` table is used not only to serve data directly but also for indexing metadata. It is currently read statically from a CSV file which is part of the codebase. As the number of available tokens is exploding, it is no longer enough to wait for Conseil releases to update the contents of the table. We should therefore load this information dynamically. 

Lorre should read the URL of a CSV file containing registered tokens data from the conf file and use it to the populate the database periodically. Based on a configuration item in the conf file, Lorre should periodically refresh the contents of the table based on the given URL. If no URL is given, Lorre should read it from the same CSV file it uses now.

piotrkosecki:
resolves #1018


vishakh:
Not merging in favor of #1024

PR #1021, opened 2021-07-23
Improvement/eeth indexer improvement

redroy44:
Save block data in one transaction.

This will prevent block data from being corrupted on indexer interrupt/kill.


piotrkosecki (approved):
LGTM

PR #1030, opened 2021-08-12
Granada related fixes and updates.

### LINKED ISSUE #966: Lorre does not log the initial connection attempt to Tezos
Initial Tezos queries should be logged just like regular Tezos queries.

piotrkosecki:
This PR contains bunch of fixes related to the Granada upgrade:
1. Baking/endorsing rights update after main loop reaches the head(previously only `estimated_time` was updated)
2. Explicitly using config setting for tracing calls and `parsing.max-to-strict-bytes`(responsible for recent errors with request too large) and adding reading  those values from env variables
3. Adding renamed field `deposits` to `CycleBalance` case class so we will avoid some warnings(and keeping the original one so we will be backward compatible)
4. Adding env variables for logging - it seems that we already support logging initial connection to Tezos node, but it was hidden - I've added it explicitly. We could say that this PR also resolves #966

### DIFF
--- a/conseil-lorre/src/main/resources/logging.conf
+++ b/conseil-lorre/src/main/resources/logging.conf
@@ -52,6 +52,7 @@ logging: {
             #  get extra info from the akka-streamed async calls to the node, if trace-calls is enabled on the node conf.
             # change the level of the root logger handler accordingly to see the output (i.e. add outputLevel: "DEBUG")
             name: "tech.cryptonomic.conseil.indexer.tezos.node-rpc.batch",
+            from-env: "LORRE_NODE_RPC_BATCH_LOG_LEVEL",
             level: "DEBUG"
         }
     ]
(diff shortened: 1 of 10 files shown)


SiddharthV1 (approved):

SiddharthV1 (approved):

conseil-lorre/src/main/resources/logging.conf
```diff
@@ -52,6 +52,7 @@ logging: {
             #  get extra info from the akka-streamed async calls to the node, if trace-calls is enabled on the node conf.
             # change the level of the root logger handler accordingly to see the output (i.e. add outputLevel: "DEBUG")
             name: "tech.cryptonomic.conseil.indexer.tezos.node-rpc.batch",
+            from-env: "LORRE_NODE_RPC_BATCH_LOG_LEVEL",
```
vishakh, line 55: Let's document this.

vishakh (changes requested):

vishakh (approved):

PR #1038, opened 2021-09-22
Fixed loading `Depth` config

amillert:
In the current version, there are two ways to configure `Lorre` via:
1. argument parser from sbt's cli,
2. corresponding config files.

The former worked correctly, allowing one to parse arguments in either textual form, e.g. `new`, or a respective numerical representation, e.g. `0`. However, the latter option couldn't handle numerical arguments passed correctly. This PR fixes this behavior.

Additionally, it establishes a priority between the two parsers. By principle, the former (argument) parser overwrites whatever comes from the config files. But it provides default values - `Newest` for `depth` parameter, and `None` as a `headHash`. Previously, if the user wanted to take advantage of the config files, hence didn't provide any arguments when running an sbt command, the app would overwrite the specified configuration with the default values. Such behavior is undesired and has been fixed in this PR as well.


conseil-lorre/src/main/scala/tech/cryptonomic/conseil/indexer/config/util/package.scala
```diff
@@ -0,0 +1,21 @@
+package tech.cryptonomic.conseil.indexer.config
+
+import scala.util.Try
+
+package object util {
```
piotrkosecki, line 5: I think you can move contents ot this package object to the ConfigUtil object

amillert: Done, named it `ConfigDepthUtil` since it solely relates to `Depth`.

piotrkosecki: I meant that you can move it to the different file which is named `ConfigUtil`, it has other objects inside and you can name it just `Depth`

amillert: done

piotrkosecki (approved):
LGTM!

amillert:
closing in favor of #1049

PR #1039, opened 2021-09-23
927 - Bakers rolls size in config

amillert:
This PR adds bakers rolls size to the `reference.conf` config file in `conseil-common`. It additionally updates tests respectively.

### DIFF
--- a/conseil-common/src/main/resources/reference.conf
+++ b/conseil-common/src/main/resources/reference.conf
@@ -24,6 +24,10 @@ platforms: [
   {
     name: "tezos"
 
+    baker-rolls-size: 8000
+    baker-rolls-size: ${?CONSEIL_XTZ_BACKER_ROLLS_SIZE}
+
+    network: "mainnet"
     network: "granadanet"
     network: ${?CONSEIL_XTZ_NETWORK}
 
--- a/conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
+++ b/conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
@@ -125,6 +125,7 @@ object Platforms {
       network: String,
       enabled: Boolean,
       node: TezosNodeConfiguration,
+      bakerRollsSize: BigDecimal,
       db: Config,
       tns: Option[TNSContractConfiguration]
   ) extends PlatformConfiguration {
--- a/conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/TezosTypes.scala
+++ b/conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/TezosTypes.scala
@@ -1,8 +1,14 @@
 package tech.cryptonomic.conseil.common.tezos
 
-import java.time.{Instant, ZonedDateTime}
-import scala.util.Try
+import tech.cryptonomic.conseil.common.config.Platforms._
+
 import cats.Functor
+import pureconfig._
+import pureconfig.generic.auto._
+
+import java.time.Instant
+import java.time.ZonedDateTime
+import scala.util.Try
 
 /**
   * Classes used for deserializing Tezos node RPC results.
@@ -14,7 +20,7 @@ object TezosTypes {
 
   //TODO use in a custom decoder for json strings that needs to have a proper encoding
   lazy val isBase58Check: String => Boolean = (s: String) => {
-    val pattern = "^[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]*$".r.pattern
+    val pattern = "^[^\\W0OlI_]*$".r.pattern
     pattern.matcher(s).matches
   }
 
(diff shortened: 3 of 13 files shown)


conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
```diff
@@ -125,6 +125,7 @@ object Platforms {
       network: String,
       enabled: Boolean,
       node: TezosNodeConfiguration,
+      bakerRollsSize: BigDecimal,
```
piotrkosecki, line 128: why is it big decimal? shouldn't `Int` be enough?

conseil-common/src/main/resources/reference.conf
```diff
@@ -24,6 +24,8 @@ platforms: [
   {
     name: "tezos"
 
+    baker-rolls-size: 8000
```
piotrkosecki, line 27: I would put it in the `conseil-lorre/src/main/resources/application.conf` - I think it's more suitable place for it as we have some more constants defined there.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/TezosTypes.scala
```diff
@@ -668,9 +674,9 @@ object TezosTypes {
   /** Utilities related to the baking process */
   object Baking {
 
-    //we'll move this to a proper chain configuration value later on
     /** how big is a baker roll in tez */
-    final val BakerRollsSize = BigDecimal.decimal(8000)
+    final val BakerRollsSize =
+      loadConfig[TezosConfiguration](namespace = "platforms").fold(_ => BigDecimal.decimal(8000), _.bakerRollsSize)
```
piotrkosecki, line 679: you don't have to put default value here, it's enough when it is in the config(but check my previous comment where it should be)

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/TezosTypes.scala
```diff
@@ -14,7 +20,7 @@ object TezosTypes {
 
   //TODO use in a custom decoder for json strings that needs to have a proper encoding
   lazy val isBase58Check: String => Boolean = (s: String) => {
-    val pattern = "^[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]*$".r.pattern
+    val pattern = "^[^\\W0OlI_]*$".r.pattern
```
piotrkosecki, line 23: are you sure this pattern negates every character that shouldn't be there? Shouldn't you use [a-zA-Z1-9], so you wouldn't have any special characters and then remove specific letters/numbers?

amillert: Yea, it should work as desired; all tests are passing as well. We negate all non-word symbols as well as underscore (which is captured by `\w`), plus the symbols that can be mismatched, namely - `I`, `l`, `O`, and `0`. In case it's ambiguous, maybe I should simply leave it as it was originally?

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/config/Platforms.scala
amillert: originally it was `BigDecimal`

amillert: The entire `TezosType` domain is built around `PositiveBigNumber` type being a wrapper for BigDecimal. Therefore, the remaining functionality expects `BigDecimal` as well.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/TezosTypes.scala
piotrkosecki: I would leave it as it was, as it was not the purpose of this ticket - in the future refactor we can think of changing it and additional tests.

conseil-common/src/main/resources/reference.conf
amillert: The only place in the code, where it's used are `MainOutputs`, and `TezosTypes` which are both located in `common`. All other usages are for testing.

We decided to leave the location as is.

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/TezosTypes.scala
amillert: sure, removed it

conseil-common/src/main/scala/tech/cryptonomic/conseil/common/tezos/TezosTypes.scala
amillert: I don't see how else can I extract `right` value from `Either` without mapping `left` to a value of the `right`'s type

piotrkosecki (approved):
LGTM!

vishakh (approved):

SiddharthV1 (approved):

PR #1041, opened 2021-09-27
Fixed test error + proper use of Try if import found in scope

amillert:
This PR fixes the error in `JsonDecodersTestFixtures` caused by the absence of the expected constructor's argument and the wrong type of another provided argument. It's enough to provide dummy `None` values for an optional parameter. Additionally, there were some occurrences of `scala.util.Try` even though there was already an import `import scala.util.Try` in the scope.


piotrkosecki (approved):
good job!

vishakh (approved):

Issue #1020, opened 2021-07-23
Question: Client library?


r4ravi2008:
Conseiljs seems to exist for javascript ecosystem. 

If I were to use this in a jvm based backend, is there a client library available?

vishakh:
@r4ravi2008 If you need to submit operations in Scala, then we can revive the operation signing code. Otherwise, we can just package the project as is as a JAR and provide it to you through Maven. What do you think?

r4ravi2008:
I dont need operations for now. If you could point me mvn repository coordinates that will be great! Thanks for considering.

PR #1042, opened 2021-09-29
Bumped up the version of `endpoints4s` library

amillert:
As part of #1020 we decided to upgrade the version of the `endpoints4s` library which has meanwhile moved from the author's repo to the official one. The newest library version, namely `1.5.0` seems to not be entirely compatible with our current codebase, hence, for the time being, we've used `1.3.0`.

This PR also fixes all the related imports regarding `endpoints4s`.


piotrkosecki (approved):
Looks good!

vishakh (approved):

PR #1051, opened 2021-11-10
Registered tokens files for hangzhounet

SiddharthV1:
Added empty csv(with column names) and empty json []


piotrkosecki (approved):
CSV file is not needed, but overall good catch

vishakh (approved):

PR #1061, opened 2021-12-17
Adding offset support to Lorre

piotrkosecki:
as described above


conseil-lorre/src/test/scala/tech/cryptonomic/conseil/indexer/tezos/TezosNodeOperatorTest.scala
```diff
@@ -47,89 +47,16 @@ class TezosNodeOperatorTest
       block.futureValue.votes shouldBe TezosTypes.CurrentVotes.empty
     }
 
-  "getLatestBlocks" should "correctly fetch all the blocks if no depth is passed-in" in {
```
vishakh, line 50: How come we removed these tests?

vishakh (approved):

